PR #4, opened 2023-01-05
Cluster watching

philbrookes:


.gitignore
```diff
@@ -16,3 +16,6 @@
 
 # Ignore bin directory contents
```
maleck13, line 17: Prob can just copy this over?
https://github.com/kcp-dev/kcp-glbc/blob/main/.gitignore

Dockerfile
```diff
@@ -12,16 +12,16 @@ COPY go.sum go.sum
 RUN go mod download
```
maleck13, line 13: same here? https://github.com/kcp-dev/kcp-glbc/blob/main/Dockerfile

Makefile
```diff
@@ -66,11 +66,11 @@ local-setup: kind kustomize helm ## Setup multi cluster traffic controller local
 
 .PHONY: build
 build: manifests generate fmt vet ## Build manager binary.
```
maleck13, line 68: do we want to add the basics like make lint make test etc? not essential

config/rbac/secret_editor_role.yaml
```diff
@@ -0,0 +1,31 @@
...
+    app.kubernetes.io/name: clusterrole
+    app.kubernetes.io/instance: secret-editor-role
+    app.kubernetes.io/component: rbac
+    app.kubernetes.io/created-by: multi-cluster-traffic-controller
+    app.kubernetes.io/part-of: multi-cluster-traffic-controller
+    app.kubernetes.io/managed-by: kustomize
+  name: secret-editor-role
+rules:
+- apiGroups:
+  - kuadrant.io
```
maleck13, line 15: this seems wrong?

config/rbac/secret_viewer_role.yaml
```diff
@@ -0,0 +1,27 @@
...
+    app.kubernetes.io/name: clusterrole
+    app.kubernetes.io/instance: secret-viewer-role
+    app.kubernetes.io/component: rbac
+    app.kubernetes.io/created-by: multi-cluster-traffic-controller
+    app.kubernetes.io/part-of: multi-cluster-traffic-controller
+    app.kubernetes.io/managed-by: kustomize
+  name: secret-viewer-role
+rules:
+- apiGroups:
+  - kuadrant.io
```
maleck13, line 15: same here some kind of weird generation perhaps?

config/samples/secret.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  labels:
+    argocd.argoproj.io/secret-type: cluster
```
maleck13, line 5: just a thought for the future but prob should add our own label to secrets to indicate which clusters users want us to manage ingress for?

pkg/_internal/metadata/labels_test.go
```diff
@@ -0,0 +1,52 @@
+package metadata
```
maleck13, line 1: assuming most of this was copied over?

pkg/controllers/secret/secret_controller.go
```diff
@@ -0,0 +1,136 @@
+/*
+Copyright 2022 pbrookes@redhat.com.
```
maleck13, line 2: we should prob remove this header and set it similar to the one in KCP
``` Copyright 2021 The MultiCluster Traffic Controller Authors.``` and leave the apache license part

File: pkg/controllers/secret/secret_controller.go [hunk 9]
```diff
@@ -0,0 +1,136 @@
...
+type ArgoClusterConfig struct {
+	BearerToken        string          `json:"bearerToken,omitempty"`
+	Username           string          `json:"username,omitempty"`
+	Password           string          `json:"password,omitempty"`
+	TlsClientConfig    TLSClientConfig `json:"tlsClientConfig,omitempty"`
+	ExecProviderConfig ProviderConfig  `json:"execProviderConfig,omitempty"`
+}
+
+const (
+	ARGO_CLUSTER_LABEL       = "argocd.argoproj.io/secret-type"
```
maleck13, line 65: Yeah I think, while we can support the format we should prob have our own label, WDYT? it can be a follow on issue though

pkg/controllers/secret/secret_controller.go
```diff
@@ -0,0 +1,136 @@
...
+	_ = log.FromContext(ctx)
+	previous := &corev1.Secret{}
+	err := r.Client.Get(ctx, client.ObjectKey{Namespace: req.Namespace, Name: req.Name}, previous)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	secret := previous.DeepCopy()
+
+	//only interested in ArgoCD cluster secrets
+	if !v1.HasLabel(secret.ObjectMeta, ARGO_CLUSTER_LABEL) || secret.ObjectMeta.Labels[ARGO_CLUSTER_LABEL] != ARGO_CLUSTER_LABEL_VALUE {
```
maleck13, line 92: is it possible to add a filter, so we only see secrets with these labels?

pkg/controllers/secret/secret_controller.go
```diff
@@ -0,0 +1,136 @@
...
+	KeyData  []byte `json:"keyData,omitempty"`
+}
+
+type ProviderConfig struct {
+	Command    string   `json:"command,omitempty"`
+	Args       []string `json:"args,omitempty"`
+	APIVersion string   `json:"apiVersion,omitempty"`
+}
+
+type ArgoClusterConfig struct {
```
maleck13, line 56: I think we could change this also to be ClusterConfig. But again not needed right now

pkg/controllers/secret/secret_controller.go
```diff
@@ -0,0 +1,136 @@
...
+
+	//only interested in ArgoCD cluster secrets
+	if !v1.HasLabel(secret.ObjectMeta, ARGO_CLUSTER_LABEL) || secret.ObjectMeta.Labels[ARGO_CLUSTER_LABEL] != ARGO_CLUSTER_LABEL_VALUE {
+		return ctrl.Result{}, nil
+	}
+
+	log.Log.Info("got Argo cluster secret", "secret", secret.Namespace+"/"+secret.Name, "server", string(secret.Data["server"]))
+
+	clusterConfig := &ArgoClusterConfig{}
+	err = json.Unmarshal(secret.Data["config"], clusterConfig)
```
maleck13, line 99: any reason this couldn't be unmarshalled into a rest.Config. All the fields seem to match and then just need to set the host?

pkg/controllers/secret/secret_controller.go
```diff
@@ -0,0 +1,136 @@
...
+	hostUrl, err := url.Parse(string(secret.Data["server"]))
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	clusterClientConfig := &rest.Config{
+		Host:        string(secret.Data["server"]),
+		BearerToken: clusterConfig.BearerToken,
+		Username:    clusterConfig.Username,
+		Password:    clusterConfig.Password,
+		TLSClientConfig: rest.TLSClientConfig{
```
maleck13, line 113: see above about rest.Config

config/rbac/secret_editor_role.yaml
philbrookes: yeah, this is left over from using kubebuilder to make the controller boilerplate, I'll remove it.

config/samples/secret.yaml
philbrookes: I did it this way, as this is how Argo labels it's secrets for clusters. Would we expect a user of Argo to go and add our label to the secrets created by Argo?

pkg/_internal/metadata/labels_test.go
philbrookes: yes

pkg/controllers/secret/secret_controller.go
philbrookes: 👍

pkg/multiClusterWatch/main.go
```diff
@@ -0,0 +1,165 @@
...
+			}
+		},
+	})
+
+	informerFactory.Start(ctx.Done())
+	informerFactory.WaitForCacheSync(ctx.Done())
+
+	log.Log.Info("started watcher events", "cluster watcher", w.ClusterName)
+
+	<-ctx.Done()
```
maleck13, line 145: I am guessing start is called in it own go routine

maleck13: have we tried with multiple secrets yet?

maleck13 (approved):
mostly looks fine. Just a few comments that might be worth addressing before merging

config/samples/secret.yaml
maleck13: I think this is not a bit problem from my understanding you can create the secrets declaratively but not an issue right now. I think though we would also want them to add annotations such as geo and cloud provider at somepoint

PROJECT
```diff
@@ -3,4 +3,14 @@ layout:
...
+resources:
+- api:
+    crdVersion: v1
+    namespaced: true
+  controller: true
+  domain: kuadrant.io
+  group: networking.k8s.io
+  kind: Ingress
+  path: github.com/Kuadrant/multi-cluster-traffic-controller/api/v1
+  version: v1
```
mikenairn, line 15: I'm not sure any of this should be added since we don't provide that api?

cmd/multi-cluster-traffic-controller/main.go
```diff
@@ -68,23 +72,20 @@ func main() {
...
+		Client:  mgr.GetClient(),
+		Scheme:  mgr.GetScheme(),
+		MCWatch: &multiClusterWatch.WatchController{Manager: mgr, HandlerFactory: multiClusterWatch.NewIngressHandlerFactory()},
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "Secret")
+		os.Exit(1)
+	}
 	//+kubebuilder:scaffold:builder
 
 	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
```
mikenairn, line 91: Is there a reason we are moving main.go here? kubebuilder scaffold commands seem to not like it:
```
kubebuilder create api --group kuadrant.io --version v1 --kind DNSRecord
...
2023/01/06 10:32:36 failed to create API: unable to run pre-scaffold tasks of "base.go.kubebuilder.io/v3": main.go file should present in the root directory
```

config/samples/secret.yaml
david-martin: It sounds reasonable for the controller to only look for secrets with a specific label for mctc.
It would avoid a dependency on the labelling that ArgoCD expects and allow some clusters to be omitted from mctc but still used by ArgoCD

Dockerfile
philbrookes: I'm not seeing the value of that

PROJECT
philbrookes: nice spot. Removed.

pkg/multiClusterWatch/main.go
philbrookes: Not yet. This does need testing, but yes, the kubebuilder manager runs this in a separate goroute.

cmd/multi-cluster-traffic-controller/main.go
philbrookes: I've moved it back now, it was just for personal preference

pkg/controllers/secret/secret_controller.go
philbrookes: I'm struggling with this in kubebuilder. So I'll merge this PR now, to unblock other work, and keep looking into this for a separate PR.

.gitignore
```diff
@@ -12,7 +12,67 @@
 *.out
 
 # Dependency directories (remove the comment below to include it)
-# vendor/
+vendor/
+
+#Ignore .kcp directory
+.kcp/
```
mikenairn, line 18: Not sure why we need this, but also not a big deal to add it here i guess

.gitignore
```diff
@@ -12,7 +12,67 @@
...
+
+#Ignore temporal directoryo
+tmp/
```
mikenairn, line 21: Sounds fancy "temporal directoryo"

mikenairn (approved):

PR #7, opened 2023-01-13
WIP host assign and tls creation

maleck13:


maleck13:
found a bug in this with multiple clusters. So adding a fix for that now

hack/local-setup.sh
```diff
@@ -50,6 +55,11 @@ deployCertManager() {
   ${KUSTOMIZE_BIN} build ${CERT_MANAGER_KUSTOMIZATION_DIR} --enable-helm --helm-command ${HELM_BIN} | kubectl apply -f -
   echo "Waiting for Cert Manager deployments to be ready..."
   kubectl -n cert-manager wait --timeout=300s --for=condition=Available deployments --all
+
+  k delete validatingWebhookConfiguration mctc-cert-manager-webhook
```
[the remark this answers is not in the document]
philbrookes: this should be kubectl (or define the alias in the script)

philbrookes:
I'm working through the validation steps, and seeing this error when the controller tries to  create the dnsrecord:
```
2023-01-24T10:54:23Z    ERROR   Dropping key after max failed retries   {"key": "default/echo", "retries": 5, "error": "DNSRecord.kuadrant.io \"s6yxqyfx3awf53iasbk7a6.\" is invalid: metadata.name: Invalid value: \"s6yxqyfx3awf53iasbk7a6.\": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')"}
github.com/Kuadrant/multi-cluster-traffic-controller/pkg/multiClusterWatch.(*ClusterWatcher).processNextWorkItem
```

Also, the controller is trying to reconcile the argocd ingress.

maleck13:
did you set the ZONE_ROOT_DOMAIN?

maleck13:
yeah I haven't added any label selector yet so it is reconciling any ingress it sees

philbrookes:
> did you set the ZONE_ROOT_DOMAIN?

ah, no I missed the new envvar, cheers!

main.go
```diff
@@ -116,3 +137,22 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+func getDNSUtilities(hostResolverType string) dns.HostResolver {
+	switch hostResolverType {
+	case "default":
+		fmt.Println("using defualt host resolver")
+		return dns.NewDefaultHostResolver()
+	case "e2e-mock":
```
philbrookes, line 146: I'd prefer if we found a better way to mock dns for e2e than this. It was quite a pain to use in GLBC.

pkg/controllers/traffic/traffic_controller.go
```diff
@@ -19,46 +19,116 @@ package traffic
...
+	// create empty DNSRecord with assigned host
+	managedHosts, records, err := r.Hosts.EnsureManagedHost(ctx, trafficAccessor)
+	if err != nil && err != dns.AlreadyAssignedErr {
 		return ctrl.Result{}, err
 	}
-
-	value := cm.Data[cmField]
-
-	metadata.AddLabel(trafficAccessor, "dummy-configmap-value", value)
+	for i, managedHost := range managedHosts {
```
philbrookes, line 82: I think this logic should be moved out to it's own function, it's quite verbose and would be better if the handle function were kept cleaner by just calling out to private functions and checking for errors.

My concern is that as the logic grows, this will become a very difficult function to scan through for the part you're interested in.

pkg/dns/hostresolv.go
```diff
@@ -0,0 +1,171 @@
...
+
+type HostAddress struct {
+	Host string
+	IP   gonet.IP
+	TTL  time.Duration
+	TXT  string
+}
+
+// ConfigMapHostResolver is a HostResolver that looks up the IP address of
+// a host from a ConfigMap. Used for testing purposes
```
philbrookes, line 37: as mentioned above. I don't think we need the configmap resolver yet? So I'd prefer to remove it, and implement something neater when we require it.

pkg/dns/service.go
```diff
@@ -0,0 +1,272 @@
...
+// Divides the number of IPs by a known weight allowance for a cluster/ingress, note that this means:
+// * Will always return 1 after a certain number of ips is reached, 60 in the current case (maxWeight / 2)
+// * Will return values that don't add up to the total maxWeight when the number of ingresses is not divisible by numIPs
+//
+// The aws weight value must be an integer between 0 and 255.
+// https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-values-weighted.html#rrsets-values-weighted-weight
+func awsEndpointWeight(numIPs int) string {
+	maxWeight := 120
+	if numIPs > maxWeight {
+		numIPs = maxWeight
```
philbrookes, line 269: why not just return 1 here?

pkg/multiClusterWatch/watch.go
```diff
@@ -84,70 +94,147 @@ func (w *WatchController) WatchCluster(config *rest.Config) (Watcher, error) {
...
+	defer w.Queue.ShutDown()
+	informerFactory := informers.NewSharedInformerFactory(w.client, RESYNC_PERIOD)
+
+	w.WatchIngress(ctx, informerFactory)
 	informerFactory.Start(ctx.Done())
 	informerFactory.WaitForCacheSync(ctx.Done())
 
 	log.Log.Info("started watcher events", "cluster watcher", w.ClusterName)
-
+	for i := 0; i < 1; i++ {
```
philbrookes, line 127: not following the point of this for statement... it reads like it will only ever iterate the once, so why not remove it?

pkg/traffic/ingress.go
```diff
@@ -36,6 +43,57 @@ func (a *Ingress) GetHosts() []string {
...
+		return false
+	}
+	// we now know what rules we have already covered so now calculate any new ones
+	for _, existing := range a.Spec.Rules {
+		if existing.Host == h || isCovered(existing.HTTP) {
+			continue
+		}
+
+		additional := existing.DeepCopy()
+		additional.Host = h
```
philbrookes, line 72: This would appear to overwrite the host, even if it were not empty?

pkg/multiClusterWatch/watch.go
```diff
@@ -84,70 +94,147 @@ func (w *WatchController) WatchCluster(config *rest.Config) (Watcher, error) {
 	return watcher, nil
 }
 
-func (w *ClusterWatcher) Start(ctx context.Context) error {
-	log.Log.Info("Starting cluster watcher", "name", w.ClusterName)
-
-	informerFactory := informers.NewSharedInformerFactory(w.client, RESYNC_PERIOD)
+func (w *ClusterWatcher) WatchIngress(ctx context.Context, sharedInformer informers.SharedInformerFactory) {
```
philbrookes, line 97: it looks like ctx is no longer used here, let's remove it?

main.go
maleck13: sure but I think that can be a follow up?

pkg/dns/service.go
maleck13: again this was just copied over from GLBC

pkg/multiClusterWatch/watch.go
maleck13: sure. again  i think it is just from the original

pkg/traffic/ingress.go
maleck13: it adds a new rule with the new host, but the old one  is left in place

pkg/dns/hostresolv.go
maleck13: do we know what something is? I worry we will just end up adding this back when we want to add more e2e tests etc?

philbrookes: I don't want to put the task of implementing a new e2e dns resolver on this unrelated PR - so I'm avoiding discussing how we could make a better one here. I'm just concerned that we're adding this when we don't need it. Which makes it more likely we'll go the way of reusing it, when we get to writing e2e tests. Why add something that isn't working very well for us, when we don't even need it yet?

In summary, I'm not looking for you to write or create a new mock dns resolver, just delete this unrequired and suboptimal copy/pasted code.

maleck13: sure ok will update

pkg/controllers/traffic/traffic_controller.go
maleck13: could you take a look now? I think it is relatively simple. Also this logic is going to be spit soon between sync and webhook

philbrookes (approved):

hack/local-setup.sh
```diff
@@ -50,6 +49,11 @@ deployCertManager() {
   ${KUSTOMIZE_BIN} build ${CERT_MANAGER_KUSTOMIZATION_DIR} --enable-helm --helm-command ${HELM_BIN} | kubectl apply -f -
   echo "Waiting for Cert Manager deployments to be ready..."
   kubectl -n cert-manager wait --timeout=300s --for=condition=Available deployments --all
+
+  kubectl delete validatingWebhookConfiguration mctc-cert-manager-webhook
+  kubectl delete mutatingWebhookConfiguration mctc-cert-manager-webhook
```
mikenairn, line 54: Why are we doing this? If they aren't required, can they be prevented from being created in the first place via some helm chart config option?

maleck13: They don't work with self signed certs so locally they kept giving errors

david-martin: I've seen this used before to filter resources from the output of helm & kustomize
```
kustomize build . | docker run --rm -i ryane/kfilt -i kind=validatingWebhookConfiguration | kubectl apply -f -
```

You can put it in a var for easier use as well
```
KFILT = docker run --rm -i ryane/kfilt
kustomize build . | $(KFILT) -i kind=validatingWebhookConfiguration | kubectl apply -f -
```

pkg/apis/v1/dnsrecord_types.go
```diff
@@ -221,6 +221,78 @@ type DNSZoneCondition struct {
...
+			deleted = true
+		} else {
+			providerSpecific = append(providerSpecific, pair)
+		}
+	}
+
+	endpoint.ProviderSpecific = providerSpecific
+	return deleted
+}
+
```
mikenairn, line 295: I assume these were copied from kcp-glbc. These functions had duplicates and were a bit all over the place as i think two people added similar things. This was the reason i didn't just add them all as is form kcp-glbc in the first place. Just noting we should come back to these and make it consistent in how they work are named. e.g. There is already a GetProviderSpecificProperty function. Doesn't need to block a merge.

pkg/dns/service.go
```diff
@@ -0,0 +1,272 @@
...
+			return activeDNSTargetIPs, fmt.Errorf("DNSLookup failed for host %s : %s", target.Value, err)
+		}
+		for _, add := range addr {
+			activeDNSTargetIPs = append(activeDNSTargetIPs, add.IP.String())
+		}
+	}
+	return activeDNSTargetIPs, nil
+}
+
+func (s *Service) GetManagedRecords(ctx context.Context, traffic traffic.Interface) ([]*v1.DNSRecord, error) {
```
mikenairn, line 61: Why `GetManagedRecords` and not `GetDNSRecords` since that's what it's returning?

maleck13: Fair. Will change

mikenairn (approved):

PR #9, opened 2023-01-18
Webhooks setup and server

sergioifg94:
## Description

Add the resources and logic to set up a webhook server running alongside the controller, as well as the reconciliation of the webhook configuration

## Verification steps

In order to verify these changes, the controller must run in-cluster and be exposed via an Ingress.

1. Run the local-setup
    ```sh
    make local-setup
    ```
1. Build the controller image
    ```sh
    # Build the image
    make docker-build
    # Load it into the cluster
    kind load docker-image controller:latest --name mctc-control-plane  --nodes mctc-control-plane-control-plane
    ```
1. Deploy the controller. Prior to that, ensure that the manager includes the required environment variables to run
    ```sh
    make deploy
    ```
1. Deploy the secret with the kubeconfig for the control cluster
1. Verify the controller Ingress is created
    ```sh
    kubectl get ingress mctc-ingress-mctc -n multi-cluster-traffic-controller-system
    ```
    ```yaml
    kind: Ingress
    apiVersion: networking.k8s.io/v1
    metadata:
      name: mctc-ingress-mctc
      namespace: multi-cluster-traffic-controller-system
      annotations:   
        mctc-component: webhook 
    spec:
      ingressClassName: nginx
      tls:
        - hosts:
            - cwxkxzmah79fqyecvplwzf.sf.hcpapps.net
          secretName: cwxkxzmah79fqyecvplwzf.sf.hcpapps.net
      rules:
        - host: host.whatever.com
          http:
            paths:
              - path: /
                pathType: Prefix
                backend:
                  service:
                    name: mctc-mctc-webhooks
                    port:
                      number: 8082
        - host: cwxkxzmah79fqyecvplwzf.sf.hcpapps.net
          http:
            paths:
              - path: /
                pathType: Prefix
                backend:
                  service:
                    name: mctc-mctc-webhooks
                    port:
                      number: 8082
    status:
      loadBalancer:
        ingress:
          - ip: 172.21.0.2
    ```
1. Verify that the webhook configuration is reconciled
    ```sh
     kubectl get mutatingwebhookconfiguration mctc -o yaml
    ```
    Ensure it references the webhook server and the `caClient` field is set
    ```yaml
    // ...
    clientConfig:
      caBundle: ...
      url: https://cwxkxzmah79fqyecvplwzf.sf.hcpapps.net/ingress
    // ...    
    ```
2. Verify that the webhook handler is reached successfully by making a change in the webhook ingress. For example, adding an annotation
    ```sh
    kubectl edit ingress mctc-ingress-mctc -n multi-cluster-traffic-controller-system
    ```
    The Pod logs should append `level=info msg="WEBHOOK HANDLER"`, and the update should be performed correctly
3. In order to verify independently that the webhook creates successfully the new DNSRecord and Certificate, disable the traffic controller
    1.  Comment out the code that starts up the controller at [main.go](https://github.com/Kuadrant/multi-cluster-traffic-controller/blob/c7293b4d6bc1d792c872a3f1e2694f5c5b0adbba/main.go#L119-L127)
    1. Follow step 2 to build the image
    2. Delete the controller Pod so that a new Pod with the new image is created
4. Create the sample service so that the creation of the Ingress triggers the webhook
    ```sh
    kubectl apply -f samples/echo-service/echo.yaml -n default
    ```
5. Verify that the DNSRecord and the Certificate are created
    ```sh
    kubectl get dnsrecords s6yxqyfx3awf53iasbk7a6.sf.hcpapps.net -n argocd
    NAME                                    AGE
    s6yxqyfx3awf53iasbk7a6.sf.hcpapps.net   117m

    kubectl get certificate s6yxqyfx3awf53iasbk7a6.sf.hcpapps.net -n argocd
    NAME                                    READY   SECRET                                  AGE
    s6yxqyfx3awf53iasbk7a6.sf.hcpapps.net   True    s6yxqyfx3awf53iasbk7a6.sf.hcpapps.net   118m
    ```
6. Verify that the Ingress is mutated
    ```sh
    kubectl get ingress echo -n default -o yaml
    ```
    ```yaml
    apiVersion: networking.k8s.io/v1
    kind: Ingress
    metadata:
      annotations:
        kuadrant.io/managed-hosts: s6yxqyfx3awf53iasbk7a6.sf.hcpapps.net
      creationTimestamp: "2023-01-25T10:07:15Z"
      generation: 1
      name: echo
      namespace: default
      resourceVersion: "8576"
      uid: caee7c05-cefe-43d6-bd67-6cd56a3033cd
    spec:
      ingressClassName: nginx
      rules:
      - http:
          paths:
          - backend:
              service:
                name: echo
                port:
                  number: 80
            path: /
            pathType: Prefix
      - host: s6yxqyfx3awf53iasbk7a6.sf.hcpapps.net
        http:
          paths:
          - backend:
              service:
                name: echo
                port:
                  number: 80
            path: /
            pathType: Prefix
      tls:
      - hosts:
        - s6yxqyfx3awf53iasbk7a6.sf.hcpapps.net
        secretName: s6yxqyfx3awf53iasbk7a6.sf.hcpapps.net
    status:
      loadBalancer:
        ingress:
        - ip: 172.21.0.2
    ```
7. 
    image


pkg/admission/ingress/webhook_handler.go
```diff
@@ -0,0 +1,147 @@
...
+			originalSerialised.Bytes(),
+			currentSerialised.Bytes(),
+		)
+	}
+
+	return admission.Allowed("")
+
+}
+
+func (h *Handler) handle(ctx context.Context, ingress *networkingv1.Ingress) (bool, error) {
```
philbrookes, line 99: This looks to be pulling the status from the Ingress to create entries in the DNS Record? I'm not sure the webhook is the best place for this? I think a lot of the logic in here should be instead run from the `pkg/controllers/traffic/traffic_controller.go`?

I have done a similar thing by calling out the traffic_controller logic from the syncer, in my syncer PR, see here: https://github.com/Kuadrant/multi-cluster-traffic-controller/pull/11/files#diff-cf3ce1c5bad0247b8bcf2c5d5eb789a53b4f752bc1684449472eca358dda2658R68

philbrookes: The same too for setting the TLS certificate and maybe even the managed host? It all will still ultimately be a response from the webhook, but by moving the logic into the traffic_controller we should get the same logic when reconciling the ingress (i.e. not just for the webhook)

sergioifg94: My understanding is that we want to have that logic in both places (the reconciler _and_ the webhook). It is a bit redundant for now but I think the webhook will have more complex logic later on when we implement custom domain verification.

Maybe @maleck13 can confirm that

maleck13: So domain verification, host / dns record registration can be duplicated, but I would go for having the actual updating of the DNSRecord happen outside of the webhook if you follow?
Webhook can validate the ingress and create the initial DNSRecord and TLS certificte  to register a new host and mutate the ingress to add tls section

philbrookes: my suggestion is rather: let's move the "duplicated" code to the traffic_controller, and call that from the webhook, effectively removing the duplication, anything that is webhook-specific should not be moved and instead stay here.

philbrookes: or in other words:
```
webhook handle -> pass ingress to traffic_controller -> modify ingress with webhook specific logic -> return ingress on webhook response.
```

maleck13: traffic controller right now modifiies the ingress directly

pkg/admission/ingress/webhook_handler.go
```diff
@@ -0,0 +1,147 @@
...
+	// when certificate ready copy secret (need to add event handler for certs)
+	// only once certificate is ready update DNS based status of ingress
+	secret, err := h.Certificates.GetCertificateSecret(ctx, managedHostRecord.Name)
+	if err != nil && !k8serrors.IsNotFound(err) {
+		return false, err
+	}
+
+	trafficAccessor.AddTLS(managedHostRecord.Name, secret)
+
+	activeDNSTargetIPs := map[string][]string{}
```
maleck13, line 126: yeah I thnk we want to remove this part

pkg/admission/ingress/webhook_handler.go
```diff
@@ -0,0 +1,147 @@
...
+
+}
+
+func (h *Handler) handle(ctx context.Context, ingress *networkingv1.Ingress) (bool, error) {
+	trafficAccessor := traffic.NewIngress(ingress)
+
+	// verify host is correct
+	// no managed host assigned assign one
+	// create empty DNSRecord with assigned host
+	managedHostRecord, err := h.Hosts.EnsureManagedHost(ctx, trafficAccessor)
```
maleck13, line 105: there is a slight change to logic in my PR now in that it returns an array. As thre could be more than one managed host assigned. (most of the time prob only one, but might as well allow for their being more than one)

pkg/admission/ingress/webhook_handler.go
philbrookes: yes, I don't think that makes this incompatible? It doesn't send it to the API, just updates it in memory.

sergioifg94: I think there's still a few differences between the logic in the reconciler and the logic in the webhook. My understanding is that the webhook:
* Only creates (but never updates) the DNSRecords and Certificates for the Ingress
* Doesn't copy the TLS secret once it's created by cert-manager. It only sets the name in the TLS section of the Ingress

We could still probably refactor it to reduce the duplicated logic to a minimum but I'd leave that for another time I think

philbrookes: ok, we can address this later.

david-martin:
Some cmds I used to get the controller running locally:

```bash
# Patch imagepullpolicy
kubectl -n multi-cluster-traffic-controller-system patch deployment mctc-controller-manager -p '{"spec": {"template": {"spec":{"containers":[{"name":"manager","imagePullPolicy":"Never"}]}}}}' -n multi-cluster-traffic-controller-system

# Create Configmap from file
kubectl -n multi-cluster-traffic-controller-system create configmap controller-config $(for envvar in $(cat controller-config.env); do echo -n "--from-literal=$envvar ";done)

# Create Secret from file
kubectl -n multi-cluster-traffic-controller-system create secret generic aws-credentials $(for envvar in $(cat aws-credentials.env); do echo -n "--from-literal=$envvar ";done)
```

@sergioifg94 I'm getting an admission error when trying to create an ingress
```
admission webhook "validate.nginx.ingress.kubernetes.io" denied the request: host "_" and path "/" is already defined in ingress multi-cluster-traffic-controller-system/mctc-ingress-mctc
```

The ingress yaml:
```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  labels:
    argocd.argoproj.io/instance: mctc-control-plane-echo-app
  name: echo
  namespace: echo-app
spec:
  rules:
    - host: ''
      http:
        paths:
          - backend:
              service:
                name: echo
                port:
                  number: 80
            path: /
            pathType: Prefix
```

sergioifg94:
@david-martin That's right, I forgot to update the verification steps after a recent change. The reason for that is that I initially didn't use the empty host for the webhooks ingress, and the Nginx webhook validates that the empty host can only be used once. I think for now you can use any different host name for the echo ingress

main.go
```diff
@@ -107,7 +111,7 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "DNSRecord")
 		os.Exit(1)
 	}
-	dnsService := dns.NewService(mgr.GetClient(), dns.NewDefaultHostResolver(), defaultCtrlNS)
+	dnsService := dns.NewService(mgr.GetClient(), dns.NewSafeHostResolver(dns.NewDefaultHostResolver()), defaultCtrlNS)
```
sergioifg94, line 114: In case anyone wonders. The reason to wrap it with the SafeHostResolver is because it'll be share both by the Controller and the Webhook server so it avoids concurrency issues

david-martin (approved):

david-martin:
@philbrookes @maleck13 I think we can land this now, if you can ack again.

@sergioifg94 Have you captured all followups?
I'm thinking:
- webhook creation for each new cluster
- what's the extent of potential problems with ingress admission hooks preventing creation e.g. wildcards not allowed by nginx

sergioifg94:
@david-martin Thank you, I pushed as a new commit the changes for the webhook creation for each new cluster. I think it all makes more sense now.
I also created that separate issue to investigate the conflict around the wildcards (#25), and I'll create a separate one to investigate potential conflicts with other traffic controllers

david-martin (approved):
Latest change verified locally with 2 workload clusters.

The webhook is created on all clusters:

```
2023-01-31T10:27:56Z	INFO	new cluster added 	{"name": "mctc-workload-1"}
2023-01-31T10:27:56Z	INFO	getting webhook configurations
2023-01-31T10:27:56Z	INFO	create/update validating webhooks
2023-01-31T10:27:56Z	INFO	create/update mutating webhooks
2023-01-31T10:27:56Z	INFO	new cluster added 	{"name": "mctc-workload-2"}
2023-01-31T10:27:56Z	INFO	getting webhook configurations
2023-01-31T10:27:56Z	INFO	create/update validating webhooks
2023-01-31T10:27:56Z	INFO	create/update mutating webhooks
2023-01-31T10:27:56Z	INFO	new cluster added 	{"name": "mctc-control-plane"}
2023-01-31T10:27:56Z	INFO	getting webhook configurations
2023-01-31T10:27:56Z	INFO	create/update validating webhooks
2023-01-31T10:27:56Z	INFO	create/update mutating webhooks
```

david-martin (approved):
Rebased

philbrookes (approved):
LGTM

PR #11, opened 2023-01-23
ingress syncer agent

philbrookes:
introduces a syncer agent that runs on the workload cluster. This wraps around the same reconciler used by the pull method.

This code does have some expectations, in particular that by the time an ingress exists on the workload cluster, it will have a matching dnsrecord created on the control plane cluster, via the webhook that is in progress.

It also assumes that the host will be set by the webhook, currently it is hardcoded.

To see this working: 
- run make local-setup (this will provision 2 kind clusters, the control plane and the workload cluster)
- update the `samples/syncer/secret.yaml` with values to access the control plane 
- update the `samples/syncer/ingress.yaml` to use a valid host for your setup
- run the `samples/syncer/demo.sh` to create the various resources on these clusters
- in one terminal run: `make run-controller` **wait for it to start**
- in another terminal run: `make run-agent`
- see the dnsrecord is annotated with a specdiff patch and that patch is applied to the spec
- see the dnsrecord created in Route53 (if running the controller with the right credentials for the requested host)

ping @david-martin @mikenairn @maleck13

I have also tested that this code will function as expected if more than one specdiff annotation is present on a DNSRecord CR.


main.go
```diff
@@ -55,8 +59,14 @@ func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
 	var probeAddr string
+	var controlPlaneConfigSecretName string
+	var controlPlaneConfigSecretNamespace string
+	var runAsSyncer bool
```
maleck13, line 64: hmm not sure why we would use the same main ? Is it just for the POC, tbh I kind of expected a new `cmd/agent/main.go`

philbrookes: I considered this, but as the kubebuilder expects the main.go to be in the root of the repo, it would cause problems for our toolchain if I moved it (ping @mikenairn experienced this). So the startup flag seemed the least intrusive. 

Having said that, I don't think it has introduced an enormous amount of complexity into main (certainly less than I anticipated) so I am happy to go with either solution now.

mikenairn: I'd  move the main personally, will make building 2 separate binaries easier/possible which is what we want. We have already deviated from the kubebuilder default pattern anyway by moving controllers into the pkg dir, so no reason just to move this also, and it makes more sense in the context of this project i think.

philbrookes: cool, I'll do that and update this PR

PR #20, opened 2023-01-26
Add lint ci check

mikenairn:
Add the golangci-lint check to the Makefile, config for it, a github action to run it on PRs back to main, and fixes for current issues.

Adds a fix for the Dockerfile and removes external dns from the local setup.

GH Workflow tested here https://github.com/mikenairn/multi-cluster-traffic-controller/actions/runs/4014580657/jobs/6895274887

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,8 +13,7 @@ RUN go mod download
 
 # Copy the go source
 COPY main.go main.go
-COPY api/ api/
-COPY pkg/controllers/ controllers/
+COPY pkg/ pkg/
 
 # Build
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -13,8 +13,7 @@ RUN go mod download
 
 # Copy the go source
 COPY main.go main.go
-COPY api/ api/
-COPY pkg/controllers/ controllers/
+COPY pkg/ pkg/
```
philbrookes, line 16: 🥳

philbrookes (approved):

PR #21, opened 2023-01-26
Make docker-build removes builder image

sergioifg94:
## Description

Currently, the docker-build Make target uses the multi-stage build feature. This results in a lighter runnable image, but still keeps the previous build image in the local repository. The build image takes 2GB of space which can accumulate quite quickly.

Add a command to the make target to prune the build image immediately after building

## Verification steps

1. Build the image
    ```sh
    make docker-build
    ```
2. Ensure only one image remains in the repository
    ```sh
    docker images | head
    REPOSITORY                                               TAG                           IMAGE ID       CREATED         SIZE
    controller                                               latest                        25dc47a6e736   7 minutes ago   67.9MB
    ...
    ```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,6 +82,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 .PHONY: docker-build
 docker-build: test ## Build docker image with the manager.
 	docker build -t ${IMG} .
+	docker image prune -f --filter label=stage=mctc-builder
 
 .PHONY: docker-push
 docker-push: ## Push docker image with the manager.
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
 # Build the manager binary
 FROM golang:1.19 as builder
+LABEL stage=mctc-builder
 ARG TARGETOS
 ARG TARGETARCH
 


Makefile
```diff
@@ -82,6 +82,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 .PHONY: docker-build
 docker-build: test ## Build docker image with the manager.
 	docker build -t ${IMG} .
+	docker image prune -f --filter label=stage=mctc-builder
```
philbrookes, line 85: does this mean that image would need to be rebuild each time? Doesn't that take a while? Would it be better if we could not prune the latest one?

mikenairn: `make docker-build` is only really used for local development, images will be built elsewhere for latest tags etc.. What is the actual size of teh images being produced because they really shouldn't be that big

mikenairn: Not sure why we need to remove images. Are you saying it produces multiple builder images?

sergioifg94: @philbrookes It would only delete the builder image that is not used. That would be rebuilt anyways if there's a change in the code

@mikenairn Yes. The reason is that when testing changes that require to re-build the image it's easy to end up with lots of those intermediate images taking space in the computer

philbrookes: yeah ok, I don't see why you'd ever rebuild if you'd not changed the code.

maleck13:
@philbrookes @mikenairn @sergioifg94 are we keeping this PR?

sergioifg94:
I'd be partial to keep it. If anyone has had to build the image multiple times to test small changes they'll be able to see how much space it ends up taking just because of the builder image. With the changes of this PR you'd end up only with the small ~60MB image that runs the controller

philbrookes:
I'm going to close this for now as it has a merge conflict and hasn't been very active. Feel free to re-open when it's addressed.

PR #23, opened 2023-01-26
provide a sample applicationset

makslion:
## What 
Adding make target to deploy a sample applicationset. It would target the `mctc-control-plane` and will deploy an echo application from `samples/echo-service/echo.yaml`. 

## Verification 
1. `make local-setup`
2. `make deploy-sample-applicationset`
3. See the application created in the ArgoCD UI


david-martin (approved):
Verified locally.
Works as expected.
image

Thanks @makslion

PR #26, opened 2023-01-31
Rename DNS service functions

mikenairn:


maleck13:
@philbrookes how does this impact on what your doing at the moment? I am happy to merge this, but the main purpose of it is to rename the dns functions so perhaps we can apply it once your work is done?

mikenairn:
There's no rush on this, it was just me making sense of what it was doing. I'm  changing stuff in these functions anyway for the MZ work, so probably can just leave this for now.

PR #31, opened 2023-02-01
Fix webhook reconciliation when controller runs locally

sergioifg94:
## Description

The changes added in #9 include the setup of an Ingress that exposes the webhook server, and the reconciliation of webhook configuration pointing to that ingress. This setup works when the controller runs in-cluster but currently fails if the controller runs locally and the webhook ingress is not found.

Add the logic to handle these errors and allow the controller to continue running if no webhook ingress is created

## Verification steps

Run the controller locally and see no errors are shown regarding the webhook config reconciliation

### DIFF
--- a/hack/local-setup.sh
+++ b/hack/local-setup.sh
@@ -25,7 +25,7 @@ KIND_CLUSTER_PREFIX="mctc-"
 KIND_CLUSTER_CONTROL_PLANE="${KIND_CLUSTER_PREFIX}control-plane"
 KIND_CLUSTER_WORKLOAD="${KIND_CLUSTER_PREFIX}workload"
 
-BASE_PROXY_PORT=8085
+BASE_PROXY_PORT=9200
 
 INGRESS_NGINX_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/ingress-nginx
 CERT_MANAGER_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/cert-manager
@@ -136,11 +136,10 @@ deployDashboard() {
 
   kubectl proxy --context kind-${clusterName} --port ${port} &
   proxyPID=$!
+  echo $proxyPID >> /tmp/dashboard_pids
 
   echo -ne "\n\n\tAccess Kubernetes Dashboard\n\n"
   echo -ne "\t\t\t* The dashboard is available at http://localhost:$port/api/v1/namespaces/kubernetes-dashboard/services/https:kubernetes-dashboard:/proxy/\n"
-  echo -ne "\t\t\t* To kill the proxy background process, run:\n"
-  echo -ne "\t\t\t\tkill $proxyPID\n"
   echo -ne "\t\tAccess the dashboard using the following Bearer Token: $token\n"
 }
 
@@ -158,6 +157,19 @@ deployWebhookConfigs(){
   kubectl apply -f $WEBHOOK_PATH/webhook-configs.yaml 
 }
 
+stopProxies() {
+  if [[ -f /tmp/dashboard_pids ]]; then
+    echo "Stopping existing proxies"
+    while read p; do
+      kill $p
+    done </tmp/dashboard_pids
+    rm /tmp/dashboard_pids
+  fi
+}
+
+# Stop proxies to the existing clusters
+stopProxies
+
 #Delete existing kind clusters
 clusterCount=$(${KIND_BIN} get clusters | grep ${KIND_CLUSTER_PREFIX} | wc -l)
 if ! [[ $clusterCount =~ "0" ]] ; then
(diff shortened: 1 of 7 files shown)


config/webhook/ingress.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
-  name: ingress-mctc
+  name: ingress
```
mikenairn, line 4: Would webhook be a more descriptive name? We know it's an ingress from the kind of resource.

sergioifg94: Agreed. Just pushed a commit with the change

david-martin:
Getting this error locally when running `make run`
```
2023-02-02T10:51:27Z	ERROR	setup	problem running manager	{"error": "listen tcp :8082: bind: address already in use"}
```

8082 is the hostPort for kind cluster
Should a different port number be used, or am I missing some config?

mikenairn:
> Getting this error locally when running `make run`
> 
> ```
> 2023-02-02T10:51:27Z	ERROR	setup	problem running manager	{"error": "listen tcp :8082: bind: address already in use"}
> ```
> 
> 8082 is the hostPort for kind cluster Should a different port number be used, or am I missing some config?

Same, I've just been disabling it entirely for now `./bin/manager --webhooks-port=0` but we should try and make all those default ports numbers not conflict with the ones we are using for the kind cluster locally.

sergioifg94:
As a curiosity, the reason we went for `8082` is that both `8080` and `8081` were conflicting with the proxy container that runs alongside the controller when running it in-cluster

david-martin:
> As a curiosity, the reason we went for `8082` is that both `8080` and `8081` were conflicting with the proxy container that runs alongside the controller when running it in-cluster

@sergioifg94 The `argocdAddCluster` function in local-setup will start from 8082 and use up ports for each new workload cluster.

Might be best to jump to a different number far enough away by default, and one that's unlikely to conflict with other things locally.

main.go
```diff
@@ -73,12 +74,17 @@ func main() {
...
 	var WebhookPortNumber int
+	var namespace string
+
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
-	flag.IntVar(&WebhookPortNumber, "webhooks-port", 8082, "The port of the webhooks server. Set to 0 disables the webhooks server")
+	flag.IntVar(&WebhookPortNumber, "webhooks-port", 9090, "The port of the webhooks server. Set to 0 disables the webhooks server")
```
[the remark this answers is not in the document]
sergioifg94: Just pushed a commit following your suggestion. Kept the webhook service in 8082, updated the cluster ports to start in 9090, and the proxy ports to 9200 (the proxy runs `kubectl proxy` to access the dashboard locally without having to expose it)

hack/local-setup.sh
```diff
@@ -135,22 +135,34 @@ deployDashboard() {
...
+
+# Stop proxies to the existing clusters
+stopProxies
+
 #Delete existing kind clusters
 clusterCount=$(${KIND_BIN} get clusters | grep ${KIND_CLUSTER_PREFIX} | wc -l)
 if ! [[ $clusterCount =~ "0" ]] ; then
   echo "Deleting previous kind clusters."
   ${KIND_BIN} get clusters | grep ${KIND_CLUSTER_PREFIX} | xargs ${KIND_BIN} delete clusters
 fi
```
mikenairn, line 163: Would a separate script make sense `local-cleanup.sh` that can be called at the start of this script, and also at any time to tear down resources/processes created by local-setup? It could delete the clusters as well as stop any dashboard proxies it started.

sergioifg94: I like that idea. I have created an initial version of that in another PR #67. I'll add the step of killing the proxy processes on top of it

pkg/_internal/controller/run.go
```diff
@@ -13,3 +15,31 @@ func IsRunningLocally() bool {
...
+
+// GetNamespace returns the namespace where the controller is running. If the
+// controller is running in the local host, and the --namespace flag has not been
+// passed, returns false
+func GetNamespace() (string, bool) {
+	return namespace, namespace != ""
+}
+
+func FindNamespace() (string, bool) {
+	ns, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
```
mikenairn, line 27: Seems a bit hacky. Is this the best way to get the current namespace?

pkg/controllers/cluster/reconciler.go
```diff
@@ -18,6 +20,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+var (
+	IngressName = "mctc-webhooks"
```
mikenairn, line 24: Is this the correct default name now?

sergioifg94:
Closing this as the main reason of this PR was relevant to the dynamic reconciliation of the webhook configurations, which has been removed in favour of them being created statically in the local setup.

I'll move the changes related to stopping the proxies to #67

PR #36, opened 2023-02-01
HCG-171 add make targets for prow

makslion:
## What 
Adding:
* `make code/fix` - to format and fix the code 
* `make code/check` to enable prow ensuring the code in PR is formatted 
* `make vendor/fix` to clean/fix up vendors 
* `make vendor/check` to enable prow ensuring vendors are fine 

## Verification 
1. Ensure commands are working as expected 
2. Ensure vendor changes in this PR aren't causing any trouble.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ endif
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
-include ./hack/make/agent.make ./hack/make/controller.make ./hack/make/syncer.make ./hack/make/dependencies.make
+include ./hack/make/*.make
 
 .PHONY: all
 all: build
--- a/OWNERS
+++ b/OWNERS
@@ -1,0 +1,20 @@
+# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
+
+approvers:
+- david-martin
+- mikenairn
+- philbrookes
+- sergioifg94
+- maleck13
+
+reviewers:
+- david-martin
+- mikenairn
+- philbrookes
+- sergioifg94
+- R-Lawton
+- roivaz
+- pmccarthy
+- maleck13
+- makslion
+- jasonmadigan
--- a/hack/make/check.make
+++ b/hack/make/check.make
@@ -1,0 +1,5 @@
+##@ Check
+## Targets to be used to ensure code quality
+.PHONY: verify-code
+verify-code: vet lint ## Verify code formatting
+	@diff -u <(echo -n) <(gofmt -d `find . -type f -name '*.go' -not -path "./vendor/*"`)


Makefile
```diff
@@ -55,8 +65,18 @@ vet: ## Run go vet against code.
...
+
+.PHONY: vendor/check
+vendor/check: vendor/fix
+	git diff --exit-code vendor/
+	git diff --exit-code go.sum
+
+.PHONY: vendor/fix
+vendor/fix:
+	go mod tidy
+	go mod vendor
```
mikenairn, line 79: We haven;t vendored module sup to now, do we really want to?

Makefile
```diff
@@ -46,6 +46,16 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
+.PHONY: code/check
+code/check:
```
mikenairn, line 50: Never convinced by these phony make targets using file paths. We haven't done this for any others, so personally i woudln't start doing it now for these. You might also want to take inspiration from how we have setup individual make files in other projects like here https://github.com/Kuadrant/kuadrant-operator/blob/main/make/verify.mk

Makefile
```diff
@@ -46,6 +46,16 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
...
+	@diff -u <(echo -n) <(gofmt -d `find . -type f -name '*.go' -not -path "./vendor/*"`)
+	golint ./... | grep -v  "comment on" | grep -v "or be unexported"
+	go vet ./...
+
+.PHONY: code/fix
+code/fix:
+	@gofmt -w `find . -type f -name '*.go' -not -path "./vendor/*"`
```
mikenairn, line 57: There's a format target below, what's the difference?

Makefile
```diff
@@ -55,8 +65,18 @@ vet: ## Run go vet against code.
 	go vet ./...
 
 .PHONY: lint
-lint: ## Run golangci-lint against code.
-	golangci-lint run ./...
+lint: ## Run golint against code.
+	golint ./...
```
mikenairn, line 69: Any reason to change this?

Makefile
makslion: We would eventually, right? 🤔

Makefile
makslion: Fair, I'll update.

Makefile
makslion: Would we want to extract formatting and vendor to another folder? 🤔 The makefile right now looks segmented with `General`, `Build`, `Deployment` and `Build Dependencies` sections. I could go further and extract those into separate files  as they tend to grow immensely 🤔  WDYT?

Makefile
makslion: Container running it might now have golangci-lint and to keep consistent with `code/check` 🤔 . Actually, since `golint` archived probably better to use the `golangci-lint` everywhere and ensure prow know what is that

Makefile
makslion: Closely followed the provided example :thinking:

mikenairn: I think the main thing here is that we just don't deviate from the naming that we already are using, so verify-* targets would make more sense to me. If we ere actually using make to manage files then sure file paths obviously make sense. 

Individual Makefiles, it's really if it makes sense. The Makefile can get quite unwieldy if we just keep adding loads of targets, if we have several verify-* type targets that are specifically used to verify stuff sis an expected state and we plan to run thos eon CI, maybe they could be grouped together somewhere.

Makefile
mikenairn: golangci-lint afaik, is the preferred linter these days, so personally I'd stick with it. Also there is a config in this repo for it, that is currently getting used both locally and by the GH action that we are using to run it (leaves comments on the PRs when linters stuff fails). so you'd need to change that if you were replacing it.

RE. enabling openshift CI, do we see us moving all our CI tasks to there or are we planning to keep GH actions? I see a lot of advantages in enabling prow for better management of changes getting back to main (tide managing merges etc..) , but we actually don't have to run all the tests there unless we really want to.

Makefile
```diff
@@ -55,9 +58,14 @@ vet: ## Run go vet against code.
 	go vet ./...
 
 .PHONY: lint
-lint: ## Run golangci-lint against code.
+lint: ## Run golint against code.
 	golangci-lint run ./...
```
mikenairn, line 62: I'd undo the comment change

Makefile
mikenairn: I have no strong opinions on whether we vendor now or not, pros and cons to it. Eventually I'd imagine we will be asked to, but it's not IMO essential right now, and we are only really hitting cons (messy commit history with vendor changes).

Point here is though, we aren't vendoring, and you are adding a check to check the vendor directory. So either this PR needs to add the vendor directory as well, and we vendor from now on, or we shoudl just remove the vendor related targets.

@maleck13 @philbrookes Any opinions on vendoring?

maleck13: lots of opinions on vendoring :) but I agree not something we need to do at this point

makslion: then I'll go with a simple formating check and will drop vendors :thinking:

Makefile
makslion: undone

Makefile
```diff
@@ -16,6 +16,9 @@ endif
 SHELL = /usr/bin/env bash -o pipefail
 .SHELLFLAGS = -ec
 
+# Include additional files
+include ./make/*.mk
```
mikenairn, line 20: I can't really remember why now, but there was some reason we added this at the bottom of the Makefile https://github.com/Kuadrant/kuadrant-operator/blob/main/Makefile#L392

make/check.mk
```diff
@@ -0,0 +1,5 @@
+##@ Check
+## Targets to be used to ensure code quality
+.PHONY: verify-code
+verify-code: vet lint
```
mikenairn, line 4: Consider adding a comment so it shows up in the help see https://github.com/Kuadrant/kuadrant-operator/blob/main/make/verify.mk#L17 `make help`

Makefile
makslion: Seems like it somehow connected to the MAKEFILE_LISTS variable - the one that Make creates just before execution to hold the list of all the files :thinking:  [This](https://stackoverflow.com/a/64394417/8716469) looks like the problem that you might've had along with the solution. Would you have a strong preference on having `include` statements in the end of the file?

make/check.mk
makslion: So that's how it works? Neat! The more we know. Changed, thanks!

OWNERS
```diff
@@ -0,0 +1,25 @@
...
+- david-martin
+- mikenairn
+- philbrookes
+- sergioifg94
+- R-Lawton
+- roivaz
+- pmccarthy
+- maleck13
+- makslion
+- jasonmadigan
```
mikenairn, line 25: @maleck13 Any opinions on what this should look like? Approvers are basically able to merge PRs

maleck13: Lets go with 

```
- david-martin
- mikenairn
- philbrookes
- sergioifg94
- maleck13
```

initially we can add others as knowledge grows etc

makslion: As it stands right now all those people have the ability to hit the "merge" button :thinking: If we narrow the `approvers` list we should, probably, review what people can do in this repo. Reduced the list to people mentioned above

makslion:
Rebased and squashed commits

maleck13 (approved):

PR #39, opened 2023-02-01
Update kindest node image to 1.26

david-martin:
2 reasons:
- keep up to date
- fix issue below seen on some local dev environments

```
 ✓ Ensuring node image (kindest/node:v1.24.7) 🖼
 ✗ Preparing nodes 📦
ERROR: failed to create cluster: could not find a log line that matches "Reached target .*Multi-User System.*|detected cgroup v1"
make: *** [local-setup] Error 1
```

### DIFF
--- a/hack/.kindUtils
+++ b/hack/.kindUtils
@@ -9,7 +9,7 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
 - role: control-plane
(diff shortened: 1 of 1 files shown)


mikenairn:
From what i remember, i used this version as it was the same as the latest version used by OpenShift. Since i don't think we really need to care too much about that, I'm happy enough we can just keep it up to date with the latest.

mikenairn (approved):
Works for me.

PR #55, opened 2023-02-03
WIP Add RUN_IN_CLUSTER_SETUP option to local-setup

mikenairn:
local-setup is intended for local development, and mostly you would be running the controller locally. Currently this is not possible because it's assuming the controller will be running in a kind cluster with internal access (--internal option used when extracting the kubeconfig).

The RUN_IN_CLUSTER_SETUP option allows you do either, but defaults to false to allow local development to work as normal.

To run local-setup when you intend to run the controller on the cluster, use:
```
RUN_IN_CLUSTER_SETUP=true make local-setup
```

Also removes externalDNS from local setup again.

Doesn't completely resolve local-setup issues, but along with changes in this PR https://github.com/Kuadrant/multi-cluster-traffic-controller/pull/31, hopefully it will.

### DIFF
--- a/hack/local-setup.sh
+++ b/hack/local-setup.sh
@@ -35,6 +35,10 @@ ISTIO_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/istio
 
 GATEWAY_API_RESOURCES_URL="github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.6.0"
 
+# If you intend to run the controller on the control plane cluster (`make deploy`) set this to true.
+# If you are running the controller locally in dev (`make run`) set this to false (default).
+: ${RUN_IN_CLUSTER_SETUP:=false}
+
 set -e pipefail
 
 deployIngressController () {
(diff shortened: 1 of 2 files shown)


hack/local-setup.sh
```diff
@@ -35,6 +35,10 @@ ISTIO_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/istio
 
 GATEWAY_API_RESOURCES_URL="github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.6.0"
 
+# If you intend to run the controller on the control plane cluster (`make deploy`) set this to true.
+# If you are running the controller locally in dev (`make run`) set this to false (default).
+: ${RUN_IN_CLUSTER_SETUP:=false}
```
mikenairn, line 40: If anyone can think of a better name for this, happy to change it.

philbrookes (approved):
the name is a bit awkward alright, but nothing better springs to mind!

roivaz:
I think argocd won't be able to talk to the other kind workload clusters (in the case that there are any) without using the internal kubeconfig. We should probably write that down somewhere so people don't hit that wall, maybe a comment in the script itself, wdyt?

mikenairn:
> I think argocd won't be able to talk to the other kind workload clusters (in the case that there are any) without using the internal kubeconfig. We should probably write that down somewhere so people don't hit that wall, maybe a comment in the script itself, wdyt?

I'll check. TBH, if it can then there might not be a need to use `--internal` for any deployment.

mikenairn:
Added WIP as it needs more thought. It should be possible to configure this in such a way that it's able to communicate with these kind clusters regardless of whether it's on or off cluster, we had a similar setup with kcp and it's syncers.

PR #57, opened 2023-02-03
Fixup GatewayAPI Roles

david-martin:


mikenairn (approved):
Hit this myself. I think if we removed this from this file it will stop adding the extra `kuadrant.io` when we use the generator https://github.com/Kuadrant/multi-cluster-traffic-controller/blob/main/PROJECT#L1

david-martin:
> if we removed this from this file it will stop adding the extra `kuadrant.io` when we use the generator

Yes, that works. `make manifests` adds the right rules to the ClusterRole then.
I've included the PROJECT change

mikenairn (approved):

PR #58, opened 2023-02-03
webhook refactor

R-Lawton:
# Verification
. `make local-setup` with `MCTC_WORKLOAD_CLUSTERS_COUNT` set to bring up workload cluster(s)
. `make deploy`

What should be created:
* tls certs in the tls directory in webhook-config/control
* Webhook Service, Ingress and TLS secret in the control cluster
* webhook configs should be created in the workload clusters with the matching CA bundle as the new created certs (when decoded should be the tls.crt duplicated)


hack/local-setup.sh
```diff
@@ -143,6 +145,21 @@ deployDashboard() {
...
+  clusterName=${1}
+  echo "Deploying the webhook configuration to (${clusterName})"
+
+  kubectl config use-context kind-${clusterName}
+
+  plain=$( cat $TLS_CERT_PATH/tls.crt $TLS_CERT_PATH/tls.crt  )
+  echo $WEBHOOK_PATH/webhook-configs.yaml
+  endocded=$( echo "$plain" | base64 )
+
+  yq -e  -i ".webhooks[0].clientConfig.caBundle =\"$endocded\"" $WEBHOOK_PATH/webhook-configs.yaml
```
sergioifg94, line 158:
```suggestion
  encoded=$( echo "$plain" | base64 )

  yq -e  -i ".webhooks[0].clientConfig.caBundle =\"$encoded\"" $WEBHOOK_PATH/webhook-configs.yaml
```
Tiny typo

hack/local-setup.sh
```diff
@@ -171,12 +188,17 @@ deployExternalDNS ${KIND_CLUSTER_CONTROL_PLANE}
 #6. Deploy argo cd
 deployArgoCD ${KIND_CLUSTER_CONTROL_PLANE}
 
-#7. Deploy Dashboard
+#7. Deploy webhook conifgs
+deployWebhookConfigs ${KIND_CLUSTER_CONTROL_PLANE}
```
sergioifg94, line 192: I might be wrong but my understanding is that we will only need the webhook configs in the workload clusters

philbrookes: Yeah, this makes sense to me. Webhook in the workload clusters, endpoint in the control plane. (Though we might want webhooks for gateways on the control plane in the future)

hack/local-setup.sh
R-Lawton: thank you nice catch

hack/local-setup.sh
R-Lawton: ok cool that makes sense, i misunderstood the flow of the script ill change that

philbrookes (approved):

PR #59, opened 2023-02-03
Gateway reconcile

### LINKED ISSUE #40: Add new controller for reconciling Gateway resources

### LINKED ISSUE #41: Add support for registering supported GatewayClasses

david-martin:
Closes #40 
Closes #41

Adds initial GatewayClass and Gateway controller implementation & tests.

To verify, run `make test`.

Alternatively, run the controller locally with the `run` target,
then create the sample gatewayclass `kubectl apply -f ./config/samples/gatewayclass.yaml`
and sample gateway `kubectl apply -f ./config/samples/gateway.yaml`


config/samples/gatewayclass.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: GatewayClass
+metadata:
+  name: mctc-gw-istio-instance-per-cluster
```
maleck13, line 4: maybe mctc-gw-istio-external?

gateway_conditions.md
```diff
@@ -0,0 +1,134 @@
...
+```yaml
+    - lastTransitionTime: "1970-01-01T00:00:00Z"
+      status: "True"
+      type: Accepted
+      reason: Accepted
+      message: Handled by kuadrant.io/mctc-gw-controller
+      observedGeneration: <corresponding metadata.generation>
+```
+
+If the Gateway is not valid or an unsupported class is specified, the condition changes to:
```
maleck13, line 31: I think we should be able to filter down to only gateways of a given class in the event handlers

config/samples/gatewayclass.yaml
david-martin: I wonder if 'external' has connotations with the gateway being outside the data plane clusters?
E.g. a service by the cloud provider or a 3rd party.
In our POC, the gateway will be internal to the data plane clusters.

If we extrapolate, the different types would be:

* Internal, instance per cluster
* Internal, 1 dedicated config cluster
* Internal, multi cluster, single region, completely managed and hidden from user
* Internal, multi cluster, multi region, completely managed and hidden from user
* External, provided by 3rd party

pkg/controllers/gateway/gatewayclass_controller.go
```diff
@@ -46,9 +59,67 @@ type GatewayClassReconciler struct {
...
+	if err != nil {
+		if err := client.IgnoreNotFound(err); err == nil {
+			return ctrl.Result{}, nil
+		} else {
+			log.Error(err, "Unable to fetch GatewayClass")
+			return ctrl.Result{}, err
+		}
+	}
+
+	if previous.Spec.ControllerName != ControllerName {
```
maleck13, line 75: as we have this filtered below prob can remove?

pkg/controllers/gateway/gatewayclass_controller.go
```diff
@@ -46,9 +59,67 @@ type GatewayClassReconciler struct {
...
+		log.Info("Not a GatewayClass for this controller", "ControllerName", previous.Spec.ControllerName)
+		return ctrl.Result{}, nil
+	}
+
+	if previous.Status.Conditions != nil &&
```
maleck13, line 80: nit: this might be more readable as a function `GatewayIsAccepted`

pkg/controllers/gateway/gatewayclass_controller.go
```diff
@@ -18,14 +18,27 @@ package gateway
...
+	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 )
 
+const (
+	ControllerName = "kuadrant.io/mctc-gw-controller"
+)
+
+func getSupportedClasses() []string {
```
maleck13, line 38: probably want to consider externalising these in the future

pkg/controllers/gateway/suite_test.go
```diff
@@ -71,10 +85,308 @@ var _ = BeforeSuite(func() {
...
+						{
+							Name:     "test-listener-1",
+							Port:     8443,
+							Protocol: gatewayv1beta1.HTTPSProtocolType,
+						},
+					},
+				},
+			}
+
+			// // Mock Cluster Secret
```
maleck13, line 270: prob can remove if not used?

pkg/controllers/gateway/suite_test.go
```diff
@@ -71,10 +85,308 @@ var _ = BeforeSuite(func() {
...
+		It("should accept a gatewayclass for this controller", func() {
+			Expect(k8sClient.Create(ctx, gatewayclass)).To(BeNil())
+			createdGatewayclass := &gatewayv1beta1.GatewayClass{}
+			gatewayclassType := types.NamespacedName{Name: gatewayclass.Name, Namespace: gatewayclass.Namespace}
+
+			// Exists
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, gatewayclassType, createdGatewayclass)
+				return err == nil
+			}, time.Second*10, time.Millisecond*250).Should(BeTrue())
```
maleck13, line 154: I imagine we might find we repeat ourselves a long with these durations. In GLBC we had something like `testtime.Short, ` etc that were just set to common durations

pkg/controllers/gateway/suite_test.go
```diff
@@ -71,10 +85,308 @@ var _ = BeforeSuite(func() {
...
+					log.Log.Error(err, "No errors expected")
+					Fail("No errors expected")
+				}
+				condition = createdGatewayclass.Status.Conditions[0]
+				return condition.Type == string(gatewayv1beta1.GatewayClassConditionStatusAccepted) && condition.Status == metav1.ConditionUnknown
+			}, time.Second*10, time.Millisecond*250).Should(BeTrue())
+		})
+
+		It("should NOT accept a gatewayclass that is 'unsupported'", func() {
+			gatewayclass.Name = "test-class-name-1"
```
maleck13, line 202: so if we filter out these in the events, I guess we expect the status to stay empty?

pkg/controllers/gateway/suite_test.go
```diff
@@ -71,10 +85,308 @@ var _ = BeforeSuite(func() {
...
+			Expect(k8sClient.Create(ctx, gatewayclass)).To(BeNil())
+			createdGatewayclass := &gatewayv1beta1.GatewayClass{}
+			gatewayclassType := types.NamespacedName{Name: gatewayclass.Name, Namespace: gatewayclass.Namespace}
+
+			// Exists
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, gatewayclassType, createdGatewayclass)
+				return err == nil
+			}, time.Second*10, time.Millisecond*250).Should(BeTrue())
```
maleck13, line 211: I see we already reuse the same time over and over :)

maleck13 (approved):
minor comments. I think the main thing was already called out by Sergio. We can filter out events from gatewayclasses we are not interested in. Prob should have some form of validating webhook that just rejects them at some point

config/samples/gatewayclass.yaml
maleck13: hmm. I had thought of external as allowing traffic into the cluster from public internet

pkg/controllers/gateway/suite_test.go
david-martin: There's 2 levels to a GatewayClass being Accepted.

The first is the controllerName matching a controller. This piece is in the event filter already so the controller only considers GatewayClasses meant for it. If the controllerName doesn't match, we do nothing.

The second is the name of the GatewayClass itself. We could filter it out early if the name doesn't match one that we 'implement'. However, it seems more helpful to set Accepted to false and set a message saying it's not supported.
This could help with versioning in future.

That being said, if the POC is taken further, a nicer pattern is for the controller to create/reconcile all GatewayClasses it supports itself.

config/samples/gatewayclass.yaml
david-martin: Yes, that lines up better with existing terminology around load balancers & gateways.

What about `mctc-gw-istio-external-instance-per-cluster`

Other types could be:
* `mctc-gw-istio-external-config-cluster`
* `mctc-gw-amazonapigw-external`

maleck13: yeah I am ok with that for now

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -46,17 +58,123 @@ type GatewayReconciler struct {
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.13.1/pkg/reconcile
 func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = log.FromContext(ctx)
+	log := log.FromContext(ctx)
+
+	previous := &gatewayv1beta1.Gateway{}
+	err := r.Client.Get(ctx, client.ObjectKey{Namespace: req.Namespace, Name: req.Name}, previous)
+	if err != nil {
```
maleck13, line 65: might be able to simplify this a little 

```
if err != nil {
		if err := client.IgnoreNotFound(err); err != nil {
			log.Error(err, "Unable to fetch Gateway")
			return ctrl.Result{}, err
		}
		return ctrl.Result{}, err
	}

```

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -46,17 +58,123 @@ type GatewayReconciler struct {
...
+			log.Error(err, "Unable to fetch Gateway")
+			return ctrl.Result{}, err
+		}
+	}
+
+	// Check if the class name is one of ours
+	gatewayClass := &gatewayv1beta1.GatewayClass{}
+	err = r.Client.Get(ctx, client.ObjectKey{Namespace: previous.Namespace, Name: string(previous.Spec.GatewayClassName)}, gatewayClass)
+	if err != nil {
+		if err := client.IgnoreNotFound(err); err == nil {
```
maleck13, line 78: same as above

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -46,17 +58,123 @@ type GatewayReconciler struct {
...
+	for _, condition := range conditions {
+		if condition.Type == string(conditionType) {
+			return &condition
+		}
+	}
+	return nil
+}
+
+func selectClusters(gateway gatewayv1beta1.Gateway) []string {
+	selector := gateway.ObjectMeta.Annotations[GatewayClusterLabelSelectorAnnotation]
```
maleck13, line 160: annotations could be nil here, 

```
if gateway.Annotations == nil{
return []string{}
} ```

File: pkg/controllers/gateway/gatewayclass_controller.go [hunk 13]
```diff
@@ -46,17 +59,76 @@ type GatewayClassReconciler struct {
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.13.1/pkg/reconcile
 func (r *GatewayClassReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = log.FromContext(ctx)
+	log := ctrllog.FromContext(ctx)
+
+	previous := &gatewayv1beta1.GatewayClass{}
+	err := r.Client.Get(ctx, client.ObjectKey{Namespace: req.Namespace, Name: req.Name}, previous)
+	if err != nil {
+		if err := client.IgnoreNotFound(err); err == nil {
```
maleck13, line 67: as above

pkg/controllers/gateway/gatewayclass_controller.go
david-martin: Agreed.
The pattern of the controller creating & managing it's own classes would hit on this.

pkg/controllers/gateway/suite_test.go
david-martin: In the first case, the Accepted status will stay as the default from the CRD
https://github.com/kubernetes-sigs/gateway-api/blob/v0.6.0/config/crd/standard/gateway.networking.k8s.io_gatewayclasses.yaml#L130-L134

In the second case, the GatewayClass was purposed set to use our controller, but it's not a supported name so the status will look something like:
```yaml
    - lastTransitionTime: "1970-01-01T00:00:00Z"
      status: "False"
      type: Accepted
      reason: InvalidParameters
      message: Invalid Parameters - Unsupported class name test-class-name-1. Must be one of [mctc-gw-istio-external-instance-per-cluster]
      observedGeneration: 1
```

maleck13 (approved):
test pass locally

PR #60, opened 2023-02-07
Add ManagedZone API and reconciler

### LINKED ISSUE #35: Define ManagedZone CRD and controller
**What**

Define the managed zone CRD. We should keep this very simple for the time being. The spec currently likely only needs the root domain for the zone. 
The status will need ready conditions, and a place to report nameservers back to the requester so that they can update their own DNS Provider to point at ours. 
We may also want to record additional information in the spec/status such which provider this Zone is hosted in and the number of records currently assigned to the zone

**Done**
- Simple API Resource that specifies a root domain for a new zone to be used by the tenant e.g. apps.mycompany.com
- New Managed Zone controller within the traffic controller
- The new zone is assigned to a provider by the traffic controller 
(only 1 provider at this stage)
- The zone is then created in a provider via the traffic controller using provider credentials 
- The zone status is updated with the nameservers for this new zone and a ready condition
- Unit tests in place

mikenairn:
closes #35 

Adds a new ManagedZone CRD that manages a dns zone in the configured provider. We currently only support route53, and a managed zone has a 1:1 mapping to a route53 hosted zone.

There are 3 managed zone use cases currently covered by this PR:

1. I want to use an existing route53 hosted zone in my AWS account.
1. I want to create and use a new hosted zone in my AWS account.
1. I want to add a hosted zone that is a child of another hosted zone.

For M2 we only need 1, but the work for the other use cases has been done, and will be required at a later stage.

All DNSRecords now belong to a ManagedZone, therefore a ManagedZone must exist in MCTC before any DNSREcords can be created. MCTC will look for a ManagedZone in the default control ns (multi-cluster-traffic-controller-system)

### Add new MZ for existing route53 hosted zone

mymanagedzone.yaml
```yaml
kind: ManagedZone
metadata:
  name: <my zone name>
spec:
  id: <aws zone id>
  domainName:  <my zone name>
  description:  <description for my zone>
```
```bash
kubectl apply -f mymanagedzone.yaml -n multi-cluster-traffic-controller-system
$ kubectl get managedzones -A
NAMESPACE                                 NAME             DOMAIN NAME      ID                                  RECORD COUNT   NAMESERVERS                                                                                         READY
multi-cluster-traffic-controller-system   mn.hcpapps.net   mn.hcpapps.net   /hostedzone/Z04114632NOABXYWH93QU   2              ["ns-2005.awsdns-58.co.uk","ns-627.awsdns-14.net","ns-1160.awsdns-17.org","ns-263.awsdns-32.com"]   True
```

### Add new MZ for a new route53 hosted zone

Same as above, but don't add a provider id in the spec. 

mymanagedzone.yaml
```yaml
kind: ManagedZone
metadata:
  name: <my zone name>
spec:
  domainName:  <my zone name>
  description:  <description for my zone>
```
```bash
kubectl apply -f mymanagedzone.yaml -n multi-cluster-traffic-controller-system
$ kubectl get managedzones -A
NAMESPACE                                 NAME                  DOMAIN NAME           ID                                  RECORD COUNT   NAMESERVERS                                                                                         READY
multi-cluster-traffic-controller-system   mn.hcpapps.net        mn.hcpapps.net        /hostedzone/Z04114632NOABXYWH93QU   2              ["ns-2005.awsdns-58.co.uk","ns-627.awsdns-14.net","ns-1160.awsdns-17.org","ns-263.awsdns-32.com"]   True
multi-cluster-traffic-controller-system   mntest1.hcpapps.net   mntest1.hcpapps.net   /hostedzone/Z09055493T16VE4KEOSXJ   2              ["ns-1930.awsdns-49.co.uk","ns-1193.awsdns-21.org","ns-569.awsdns-07.net","ns-213.awsdns-26.com"]   True
```

### Add new MZ for a new route53 hosted zone with a parent zone

Same as above, but include a parent ref:

mymanagedzone.yaml
```yaml
kind: ManagedZone
metadata:
  name: <my zone name>
spec:
  domainName:  <my zone name>
  description:  <description for my zone>
  parentManagedZone:
    name: <my parent zone name>
```
```bash
kubectl apply -f mymanagedzone.yaml -n multi-cluster-traffic-controller-system
$ kubectl get managedzones -A
NAMESPACE                                 NAME                   DOMAIN NAME            ID                                  RECORD COUNT   NAMESERVERS                                                                                         READY
multi-cluster-traffic-controller-system   mn.hcpapps.net         mn.hcpapps.net         /hostedzone/Z04114632NOABXYWH93QU   3              ["ns-2005.awsdns-58.co.uk","ns-627.awsdns-14.net","ns-1160.awsdns-17.org","ns-263.awsdns-32.com"]   True
multi-cluster-traffic-controller-system   mntest1.hcpapps.net    mntest1.hcpapps.net    /hostedzone/Z09055493T16VE4KEOSXJ   2              ["ns-1930.awsdns-49.co.uk","ns-1193.awsdns-21.org","ns-569.awsdns-07.net","ns-213.awsdns-26.com"]   True
multi-cluster-traffic-controller-system   test1.mn.hcpapps.net   test1.mn.hcpapps.net   /hostedzone/Z09292451HZ2HEL2Q12HV   2              ["ns-1148.awsdns-15.org","ns-1578.awsdns-05.co.uk","ns-102.awsdns-12.com","ns-883.awsdns-46.net"]   True```
```

Note the record count on the parent zone has updated. This is because a new NS type DNSRecord has been created owned by the parent zone:

```bash
$ kubectl get dnsrecord -A
NAMESPACE                                 NAME                   READY
multi-cluster-traffic-controller-system   test1.mn.hcpapps.net   True
$ kubectl tree managedzones mn.hcpapps.net -n multi-cluster-traffic-controller-system
NAMESPACE                                NAME                                READY  REASON           AGE 
multi-cluster-traffic-controller-system  ManagedZone/mn.hcpapps.net          True   ProviderSuccess  9m1s
multi-cluster-traffic-controller-system  ├─DNSRecord/test1.mn.hcpapps.net    True   ProviderSuccess  110s
multi-cluster-traffic-controller-system  └─ManagedZone/test1.mn.hcpapps.net  True   ProviderSuccess  111s
```


docs/managed_zones.md
```diff
@@ -0,0 +1,52 @@
...
+
+The `id` is the provider ID for an existing hosted zone in route53. Note, if you do not have a hosted zone, you can leave this blank and it will create one for `domainName` assuming the aws account has that domain registered in route53.
+
+```
+kubectl get managedzones -A
+NAMESPACE   NAME             DOMAIN NAME      ID                                  RECORD COUNT   NAMESERVERS                                                                                         READY
+mctc        dev.hcpapps.net  dev.hcpapps.net   /hostedzone/xy....zz               2              ["ns-2005.awsdns-58.co.uk","ns-627.awsdns-14.net","ns-1160.awsdns-17.org","ns-263.awsdns-32.com"]   True
+```
+
+### Create a managed zone
```
maleck13, line 50: looks left over?

pkg/apis/v1/managedzone_types.go
```diff
@@ -0,0 +1,154 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ManagedZoneReference holds a reference to a ManagedZone
+type ManagedZoneReference struct {
+	// `namespace` is the namespace of the managed zone
+	// Required
+	Namespace string `json:"namespace"`
```
maleck13, line 27: Wondering if this is needed at this point? or would same namespace suffice for now?

pkg/controllers/managedzone/managedzone_controller.go
```diff
@@ -0,0 +1,371 @@
...
+
+func (r *ManagedZoneReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	_ = log.FromContext(ctx)
+
+	previous := &v1.ManagedZone{}
+	err := r.Client.Get(ctx, client.ObjectKey{Namespace: req.Namespace, Name: req.Name}, previous)
+	if err != nil {
+		if err := client.IgnoreNotFound(err); err == nil {
+			return ctrl.Result{}, nil
+		} else {
```
maleck13, line 62: prob can remove this else?

pkg/controllers/managedzone/managedzone_controller.go
```diff
@@ -0,0 +1,371 @@
...
+		message = fmt.Sprintf("The DNS provider failed to ensure the managed zone: %v", err)
+
+		err = r.Status().Update(ctx, managedZone)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	// Create the parent zone NS record
+	err = r.createParentZoneNSRecord(ctx, managedZone)
```
maleck13, line 121: Is the usecase for the parent zone, if we want to setup a new zone under our default zone we do this by creating the new managed zone resource and pointing it at the parent and this will result in a NS record in the parent and a new zone for the child?

pkg/controllers/managedzone/managedzone_controller.go
```diff
@@ -0,0 +1,371 @@
...
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *ManagedZoneReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&v1.ManagedZone{}).
+		Owns(&v1.DNSRecord{}).
```
maleck13, line 160: Interesting does this trigger a reconcile if something it owns changes? Do we need to own and for ManagedZone?

maleck13:
Given this a once over, it is looking reasonable to me. One question I have is around the parent namespace ref. I assume setting this is considered very privileged. I think we would need to think carefully how that and by whom that was set also how we would verify that the target of the ref was allowed to be targetted.

docs/managed_zones.md
mikenairn: This was accidentally committed, will be completly rewritten

pkg/apis/v1/managedzone_types.go
mikenairn: I think we can probably assume everything is in the same ns as the MZ now. If that requirement changes we can deal with it later.

pkg/controllers/managedzone/managedzone_controller.go
mikenairn: Yes. This was going to be required when we were looking at a zone per tenant, and i had already written the code. We could probably get away without it for M2 now that we are just going for a single MZ tbh, but it is handy for creating zones for testing purposes.

config/crd/bases/kuadrant.io_dnsrecords.yaml
```diff
@@ -15,7 +15,12 @@ spec:
     singular: dnsrecord
   scope: Namespaced
   versions:
-  - name: v1
+  - additionalPrinterColumns:
+    - description: DNSRecord ready.
+      jsonPath: .status.conditions[?(@.type=="Ready")].status
+      name: Ready
+      type: string
+    name: v1
```
david-martin, line 23: should we bump this up to v2,
or bump it down to something like v1alpha1 to reflect it's subject to change?

config/crd/bases/kuadrant.io_managedzones.yaml
```diff
@@ -0,0 +1,146 @@
...
+      type: string
+    - description: The NameServers assigned by the provider for this zone.
+      jsonPath: .status.nameServers
+      name: NameServers
+      type: string
+    - description: Managed Zone ready.
+      jsonPath: .status.conditions[?(@.type=="Ready")].status
+      name: Ready
+      type: string
+    name: v1
```
david-martin, line 39: maybe v1alpha1 here too as it's feels too soon to be locking down a version

pkg/apis/v1/dnsrecord_types.go
```diff
@@ -105,15 +105,22 @@ func (e *Endpoint) String() string {
...
 // DNSRecordStatus defines the observed state of DNSRecord
 type DNSRecordStatus struct {
-	// zones are the status of the record in each zone.
-	Zones []DNSZoneStatus `json:"zones,omitempty"`
+
+	// conditions are any conditions associated with the record in the managed zone.
+	//
+	// If publishing the record fails, the "Failed" condition will be set with a
+	// reason and message describing the cause of the failure.
+	Conditions []DNSRecordCondition `json:"conditions,omitempty"`
```
david-martin, line 123: Any reason not to use `Conditions []metav1.Condition` here

pkg/apis/v1/managedzone_types.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// ManagedZoneStatus defines the observed state of a Zone
+type ManagedZoneStatus struct {
+	// List of status conditions to indicate the status of a ManagedZone.
+	// Known condition types are `Ready`.
+	// +listType=map
+	// +listMapKey=type
+	// +optional
+	Conditions []ManagedZoneCondition `json:"conditions,omitempty"`
```
david-martin, line 51: Same here. `metav1.Condition` ?

cmd/controller/main.go
```diff
@@ -60,7 +62,7 @@ func init() {
 
 const (
 	//(cbrookes) This will be removed in the future when we have many tenant ns and way to map to them
-	defaultCtrlNS       = "argocd"
+	defaultCtrlNS       = "multi-cluster-traffic-controller-system"
```
mikenairn, line 65: @maleck13 Why was this "argocd" up to now, and not the ns that we intend the controller to run in? Just trying to understand what this was for previously.

config/crd/bases/kuadrant.io_dnsrecords.yaml
mikenairn: I don't think we should make it v2, but yeah, all these APIs should probably be v1alpha1 for now

cmd/controller/main.go
maleck13: It was because we were using the secrets in the argocd ns

config/crd/bases/kuadrant.io_dnsrecords.yaml
mikenairn: Chnaged to v1alpha1

config/crd/bases/kuadrant.io_managedzones.yaml
mikenairn: Changed to v1alpha1

pkg/apis/v1/dnsrecord_types.go
mikenairn: Updated to use metav1.Condition

pkg/apis/v1/managedzone_types.go
mikenairn: Updated to use metav1.Condition

cmd/controller/main.go
mikenairn: Does it still need to be argocd? Do the secrets have to be in that ns?

david-martin:
First pass trying out the functionality looks good.

```
kubectl --context kind-mctc-control-plane  get managedzones -A
NAMESPACE                                 NAME                       DOMAIN NAME                ID                                  RECORD COUNT   NAMESERVERS                                                                                         READY
multi-cluster-traffic-controller-system   davem.dev.hcpapps.net      davem.dev.hcpapps.net      /hostedzone/<zoneid>   2              ["ns-1930.awsdns-49.co.uk","ns-373.awsdns-46.com","ns-1428.awsdns-50.org","ns-682.awsdns-21.net"]   True
multi-cluster-traffic-controller-system   dm.dev.hcpapps.net         dm.dev.hcpapps.net         /hostedzone/<zoneid>   30             ["ns-778.awsdns-33.net","ns-1860.awsdns-40.co.uk","ns-354.awsdns-44.com","ns-1281.awsdns-32.org"]   True
multi-cluster-traffic-controller-system   test1.dm.dev.hcpapps.net   test1.dm.dev.hcpapps.net   /hostedzone/<zoneid>   2              ["ns-1776.awsdns-30.co.uk","ns-362.awsdns-45.com","ns-626.awsdns-14.net","ns-1343.awsdns-39.org"]   True
```

```
kubectl --context kind-mctc-control-plane  get managedzone test1.dm.dev.hcpapps.net -n multi-cluster-traffic-controller-system -o=jsonpath='{.metadata.ownerReferences}'
[{"apiVersion":"kuadrant.io/v1alpha1","blockOwnerDeletion":true,"controller":true,"kind":"ManagedZone","name":"dm.dev.hcpapps.net","uid":"53038631-4d5e-460e-a67a-2ae578b44f9d"}]
```

```
kubectl --context kind-mctc-control-plane  get dnsrecords -A
NAMESPACE                                 NAME                       READY
multi-cluster-traffic-controller-system   test1.dm.dev.hcpapps.net   True
```

I'll take a look through the code next.

README.md
```diff
@@ -52,9 +51,8 @@ When deploying the multi cluster traffic controller using the make targets the f
     ```
 
 ## 2. Running the controller locally:
-1. Create two env files:
-    * One called `aws-credentials.env` containing **AWS_ACCESS_KEY_ID** and **AWS_SECRET_ACCESS_KEY**
-    * One called `controller-config.env` containing **AWS_DNS_PUBLIC_ZONE_ID** and **ZONE_ROOT_DOMAIN**
+1. Create env files:
+    * One called `aws-credentials.env` containing **AWS_ACCESS_KEY_ID**, **AWS_SECRET_ACCESS_KEY** and **AWS_REGION**
```
david-martin, line 55: Is the controller-config.env file needed anymore?
If not, we can remove it from the steps below 
```
```

pkg/_internal/conditions/conditions.go
```diff
@@ -0,0 +1,52 @@
...
+func HasCondition(existingConditions []metav1.Condition, c metav1.Condition) bool {
+	for _, cond := range existingConditions {
+		if c.Type == cond.Type && c.Status == cond.Status {
+			return true
+		}
+	}
+	return false
+}
+
+func SetCondition(conditions []metav1.Condition, generation int64, conditionType string, status metav1.ConditionStatus, reason, message string) []metav1.Condition {
```
david-martin, line 23: This is useful.
I'll work it in to the gateway reconcile pr around [here](https://github.com/Kuadrant/multi-cluster-traffic-controller/pull/84/files#diff-6afa7d73841a5c0ebdb78eaf859ce5a400203a037f871440940b5347c3a97376R256-R304)

pkg/dns/aws/dns.go
```diff
@@ -49,120 +47,155 @@ const (
...
+func (p *Route53DNSProvider) Delete(record *v1alpha1.DNSRecord, managedZone *v1alpha1.ManagedZone) error {
+	return p.change(record, managedZone, deleteAction)
+}
+
+func (p *Route53DNSProvider) EnsureManagedZone(zone *v1alpha1.ManagedZone) (ManagedZoneOutput, error) {
+	var zoneID string
+	if zone.Spec.ID != "" {
+		zoneID = zone.Spec.ID
+	} else {
+		zoneID = zone.Status.ID
```
david-martin, line 115: Is there an edge case where the ID is not set in either?

pkg/dns/service.go
```diff
@@ -288,20 +288,30 @@ func (s *Service) PatchTargets(ctx context.Context, targets, hosts []string, clu
...
-// this is temporary and will be replaced in the future by CRD resources
-type zone struct {
-	v1.DNSZone
-	RootDomain string
-	Default    bool
-}
+// GetManagedZone returns the most suitable ManagedZone to publish the given traffic resource into.
+// Currently, this just returns the first ManagedZone found in the traffic resources own namespace, or if none is found,
+// it looks for the first one listed in the default ctrl namespace.
+func (s *Service) GetManagedZone(ctx context.Context, t traffic.Interface) (*v1alpha1.ManagedZone, error) {
```
david-martin, line 294: I'm thinking about how to work this into Gateways, and wondering if the current implementation is what we need for both Ingress and Gateways.
Both will have potentially multiple hosts (Ingress.spec.rules and Gateway.spec.listeners), so more than 1 MZ may be required.
I think for an initial implementation, what's here would work for both, with the limitation that you have to keep all hosts in the same 'zone'.

david-martin (commented):
I'm leaning on the side of approving the PR as is.
I've left a few comments to better understand the changes first.

pkg/dns/service.go
maleck13: Yeah we might want to adjust this to be GetManagedZones? or have this take the specific host from the listener/spec?

david-martin: Per host/listener is more useful than getting back a list of MZs.
I'm thinking a function that takes a list of hosts (strings, so it works for Ingress & Gateways), and returns a map of hosts to MZ.
A per host lookup function may be useful too, but it might be too easy to overuse it in a loop and hit a performance penalty as each loop may be doing a lot of the same data lookup and mapping.

david-martin: No need to change this right now though btw.
Let's refactor it to satisfy the Gateway use case in #84

README.md
mikenairn: I just added this back. It's not actually need by the controller, but i have added a config to generate a ManagedZone as part of local-setup using the contents of that file see : 

```
./bin/kustomize --reorder none --load-restrictor LoadRestrictionsNone build config/local-setup/controller | kubectl apply -f -
```

hack/make/agent.make
```diff
@@ -1,3 +1,5 @@
+##@ Agent
```
mikenairn, line 1: @philbrookes fyi, adding these makes the help nicer, see :

```
$ make help

Usage:
  make <target>

General
  help             Display this help.

Development
  clean            Clean up temporary files.
  manifests        Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
  generate         Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
  fmt              Run go fmt against code.
  vet              Run go vet against code.
  lint             Run golangci-lint against code.
  test             Run tests.
  local-setup      Setup multi cluster traffic controller locally using kind.
  local-cleanup    Cleanup resources created by local-setup
  build            Build all binaries.

Deployment
  install          Install CRDs into the K8s cluster specified in ~/.kube/config.
  uninstall        Uninstall CRDs from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
  dev-tls          Generate dev tls webhook cert if necessary.

Agent
  build-agent      Build agent binary.
  docker-build-agent  Build docker image with the agent.
  docker-push-agent  Push docker image with the agent.
  deploy-agent     Deploy controller to the K8s cluster specified in ~/.kube/config.
  undeploy-agent   Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.

Controller
  build-controller  Build controller binary.
  docker-build-controller  Build docker image with the controller.
  docker-push-controller  Push docker image with the controller.
  deploy-controller  Deploy controller to the K8s cluster specified in ~/.kube/config.
  undeploy         Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.

Syncer
  build-syncer     Build syncer binary.
  docker-build-syncer  Build docker image with the syncer.
  docker-push-syncer  Push docker image with the syncer.
  deploy-syncer    Deploy controller to the K8s cluster specified in ~/.kube/config.
  undeploy-syncer  Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.

Build Dependencies
  kustomize        Download kustomize locally if necessary. If wrong version is installed, it will be removed before downloading.
  controller-gen   Download controller-gen locally if necessary. If wrong version is installed, it will be overwritten.
  envtest          Download envtest-setup locally if necessary.
  kind             Download kind locally if necessary.
```

hack/make/agent.make
```diff
@@ -7,8 +9,8 @@ build-agent: manifests generate fmt vet ## Build agent binary.
 .PHONY: run-agent
 run-agent: manifests generate fmt vet install
 	go run ./cmd/agent/main.go \
-	    --metrics-bind-address=:8082 \
-	    --health-probe-bind-address=:8083 \
+	    --metrics-bind-address=:8084 \
+	    --health-probe-bind-address=:8085 \
 	    --control-plane-config-namespace=mctc-system
```
mikenairn, line 14: @philbrookes This was causing issues when running the controller and the agent locally since the controller uses 8082 for the webhook

README.md
david-martin: Nice

cmd/controller/main.go
mikenairn: @david-martin Will changing this cause any issues? We shoudl probably just be adding resources (DNSRecord/ManagedZones) into the same NS the controller is running in, but changing this was the quickest way to do it, and works when you are running it locally.

david-martin: No.
I had changed that too on the gateway reconciler branch https://github.com/Kuadrant/multi-cluster-traffic-controller/pull/84/files#diff-369ede0a9a64763f9a3f47aa07fa8a9d1062b3df2783458f02e49cf7f1f3554dR63

pkg/dns/aws/dns.go
mikenairn: If `zone.Status.ID` isn't set then we need to create one, and after we create it updates the status with the ID.

pkg/dns/service.go
mikenairn: Yes, this will need updated once it's merged to work with Gateways. My understanding is the logic will be slightly different, requiring a lookup of managed zones based on the host(s) in the gateway spec.

david-martin (approved):

PR #61, opened 2023-02-08
Kuadrant operator

R-Lawton:
# Verification
1. Run make local setup 
`make local-setup MCTC_WORKLOAD_CLUSTERS_COUNT=<NUMBER_OF_CLUSTERS>`

Check  in the workload cluster that the following happens:
Note make sure you switch context to the right workload clusters
` kubectl config use-context kind-mctc-workload-1`
1. Namespace called `kuadrant-system` gets created
3. Operator gets installed
4. Kuadrant API CRDs should be available
`kubectl get crds | grep kuadrant`
```
authconfigs.authorino.kuadrant.io           2023-02-14T13:19:51Z
authorinos.operator.authorino.kuadrant.io   2023-02-14T13:19:51Z
authpolicies.kuadrant.io                    2023-02-14T13:19:47Z
kuadrants.kuadrant.io                       2023-02-14T13:19:47Z
limitadors.limitador.kuadrant.io            2023-02-14T13:19:51Z
ratelimitpolicies.kuadrant.io               2023-02-14T13:19:47Z
```

### DIFF
--- a/hack/local-setup.sh
+++ b/hack/local-setup.sh
@@ -119,6 +119,14 @@ installGatewayAPI() {
   ${KUSTOMIZE_BIN} build ${GATEWAY_API_KUSTOMIZATION_DIR} | kubectl apply -f -
 }
 
+deployKuadrant(){
+  clusterName=${1}
+  kubectl config use-context kind-${clusterName} 
+
+  echo "Installing Kuadrant in ${clusterName}"
+  ${KUSTOMIZE_BIN} build config/kuadrant | kubectl apply -f -
+}
+
 deployDashboard() {
   clusterName=${1}
   portOffset=${2}
@@ -191,8 +199,6 @@ deployDashboard $KIND_CLUSTER_CONTROL_PLANE 0
 #8. Add the control plane cluster
 argocdAddCluster ${KIND_CLUSTER_CONTROL_PLANE} ${KIND_CLUSTER_CONTROL_PLANE}
 
-
-
 #9. Add workload clusters if MCTC_WORKLOAD_CLUSTERS_COUNT environment variable is set
 if [[ -n "${MCTC_WORKLOAD_CLUSTERS_COUNT}" ]]; then
   for ((i = 1; i <= ${MCTC_WORKLOAD_CLUSTERS_COUNT}; i++)); do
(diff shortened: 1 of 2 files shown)


hack/local-setup.sh
```diff
@@ -120,6 +124,27 @@ installGatewayAPI() {
...
+      git reset --hard
+      git fetch && git checkout $GIT_TAG_KUAD
+    fi
+  fi  
+  echo "Kuadrant operator is installed and at the correct version"
+  kubectl create namespace kuadrant-system 
+  cd ..
+  make install -C kuadrant-operator/ && make deploy -C kuadrant-operator/
+}
+
```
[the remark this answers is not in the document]
R-Lawton: Updated to use kustomize

mikenairn:
@R-Lawton This all looks good to me, but you might want to check it is installing everything as expected. It doesn't look like any of those kustomize files have changed much since i added them, which makes me wonder if they are still working as intended. Maybe just check everything that gets installed as part of a `make deploy`, is also being installed when doing it this way.

They are also referencing main of all the kuadrant dependencies. This might be fine for our current dev needs, but eventually we will likely want a kuadrant install that is a fixed version, including the version of it's dependencies (authorino, limitador). Just something to think about, might need to work with the kuadrant team to make sure these manifests are updated as expected during a release.

R-Lawton:
@mikenairn Yes, I thought the same thing so I have already made sure things were the same with what was created with the make targets vs kustomize. The make target calls the deploy directory Kustomize yaml which is what I have the local setup script use. The only extra thing the make deploy target does is change the image if specified but by default, the image is set to latest. The make install target installs the crds which via kustomize gets installed via the deploy kustomization file.

hack/local-setup.sh
```diff
@@ -191,21 +199,19 @@ deployDashboard $KIND_CLUSTER_CONTROL_PLANE 0
...
-
-
 #9. Add workload clusters if MCTC_WORKLOAD_CLUSTERS_COUNT environment variable is set
 if [[ -n "${MCTC_WORKLOAD_CLUSTERS_COUNT}" ]]; then
   for ((i = 1; i <= ${MCTC_WORKLOAD_CLUSTERS_COUNT}; i++)); do
     kindCreateCluster ${KIND_CLUSTER_WORKLOAD}-${i} $((${port80} + ${i})) $((${port443} + ${i}))
     installGatewayAPI ${KIND_CLUSTER_WORKLOAD}-${i}
     deployIngressController ${KIND_CLUSTER_WORKLOAD}-${i}
     deployIstio ${KIND_CLUSTER_WORKLOAD}-${i}
+    deployKuadrant ${KIND_CLUSTER_CONTROL_PLANE}-${i}
```
david-martin, line 209: Getting an error
```
error: no context exists with the name: "kind-mctc-control-plane-1"
make: *** [local-setup] Error 1
````

Should it be this?
```suggestion
    deployKuadrant ${KIND_CLUSTER_WORKLOAD}-${i}
```

david-martin (changes requested):
Working with the suggested change above.
```
kubectl --context kind-mctc-workload-1 get crd|grep -i kuadrant
authconfigs.authorino.kuadrant.io           2023-02-15T10:47:22Z
authorinos.operator.authorino.kuadrant.io   2023-02-15T10:47:22Z
authpolicies.kuadrant.io                    2023-02-15T10:47:22Z
kuadrants.kuadrant.io                       2023-02-15T10:47:22Z
limitadors.limitador.kuadrant.io            2023-02-15T10:47:22Z
ratelimitpolicies.kuadrant.io               2023-02-15T10:47:22Z
```

Using `main` is fine for dev at this time, as Mike said, unless there's a version anyone on the kuadrant team slack would recommend

hack/local-setup.sh
R-Lawton: yes thank you copy paste error

david-martin (approved):

PR #62, opened 2023-02-09
Local deploy using env files

### LINKED ISSUE #19: Add AWS config & credentials to the 'deploy' target
Currently if you want to deploy the controller to a cluster (local or otherwise), you need to patch the Deployment to either have env vars for AWS access, or pull them in from configmap &/or secret.

In glbc, a local aws-credentials.env and controller-config.env file, which are in the gitignore, are used to store these values and then pulled in via kustomize https://github.com/kcp-dev/kcp-glbc/blob/main/config/deploy/local/kcp-glbc/kustomization.yaml#L16-L26
The .env files have a template, with some sensible defaults so a developer can copy it.
This model could be used here unless there's a better model we can come up with.

sergioifg94:
Closes #19 

Add a kustomization that creates a ConfigMap and Secret from the local env files and mounts them in the controller Pod

## Verification steps

> Ensure you have the correct values in the local env files `controller-config.env` and `aws-credentials.env`

1. Run local setup
    ```sh
    make local-setup
    ```
2. Build and load the image
    ```sh
    make docker-build
    kind load docker-image controller:latest --name mctc-control-plane  --nodes mctc-control-plane-control-plane    
    ```
3. Deploy the controller
    ```sh
    make deploy
    ```
4. Verify that the controller container runs correctly

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,7 +143,7 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified 
 .PHONY: deploy-controller
 deploy-controller: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${CONTROLLER_IMG}
-	$(KUSTOMIZE) build config/default | kubectl apply -f -
+	$(KUSTOMIZE) --load-restrictor LoadRestrictionsNone build config/deploy/local | kubectl apply -f -
 
 .PHONY: deploy-agent
 deploy-agent: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
--- a/README.md
+++ b/README.md
@@ -31,25 +31,9 @@ When deploying the multi cluster traffic controller using the make targets the f
     make local-setup MCTC_WORKLOAD_CLUSTERS_COUNT=<NUMBER_WORKLOAD_CLUSTER>
     ```
 
-1. Update the manager to contain the necessary environment variable and update the image policy (N.B. Do not copy/paste this, the hyphens from github are bad for the YAML parser):
-    ```
-    containers:
-        - command:
-            - /controller
-            args:
-            - --leader-elect
-            image: controller:latest
-            imagePullPolicy: Never
-            env:
-            - name: AWS_ACCESS_KEY_ID
-            value: <AWS_ACCESS_KEY_ID>
-            - name: AWS_SECRET_ACCESS_KEY
-            value: <AWS_SECRET_ACCESS_KEY>
-            - name: AWS_DNS_PUBLIC_ZONE_ID
-            value: <AWS_DNS_PUBLIC_ZONE_ID>
-            - name: ZONE_ROOT_DOMAIN
-            value: <ZONE_ROOT_DOMAIN>
-    ```
+1. Create two env files:
+    * One called `aws-credentials.env` containing **AWS_ACCESS_KEY_ID** and **AWS_SECRET_ACCESS_KEY**
+    * One called `controller-config` containing **AWS_DNS_PUBLIC_ZONE_ID** and **ZONE_ROOT_DOMAIN**
 
 1. Build the controller image and load it into the control plane
     ```sh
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -121,7 +121,7 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 .PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	$(KUSTOMIZE) build config/default | kubectl apply -f -
+	$(KUSTOMIZE) --load-restrictor LoadRestrictionsNone build config/deploy/local | kubectl apply -f -
```
mikenairn, line 124: What does `--load-restrictor LoadRestrictionsNone` do?

config/deploy/local/kustomization.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+configMapGenerator:
+  - name: controller-config
+    envs:
+      - ../../../controller-config.env
+
+secretGenerator:
+  - name: aws-credentials
+    envs:
+      - ../../../aws-credentials.env
```
mikenairn, line 24: I think in kcp-glbc we just had these always be generated for all deployments and merged in local values. Do we see any value in just doing that here, since presumably we need to inject these config/credential values for all deployments anyway, not just loical.

Makefile
sergioifg94: It disables the restriction that the env files have to be under the kustomization directory, so we can reference them from the root of the project

mikenairn: Interesting, good to know.

config/deploy/local/kustomization.yaml
mikenairn: Presumably these files need to exist, does anything make sure of that, or does the deploy just fail and we have it in the readme or soemthing?

mikenairn:
This is fine for local dev. Although i don't think it fully resolves what was being asked for in https://github.com/Kuadrant/multi-cluster-traffic-controller/issues/19. Even though that does mention `make deploy`, the solution we used for kcp-glbc would work for all deployments and not just local deployments which this does.

Do you see a deployment kustomize overlay having to inject all of this in this way? Is there ever a time when a deployment is not going to need an aws credentials secret?

sergioifg94:
I updated the PR to have the mounting of the environment variables by default in `config/manager/manager.yaml`, and I updated the README to reflect that the deployment simply needs to have the env files instead of modifying the manifest.

I didn't add the functionallity to merge environment variables for the sake of simplicity as it might become confusing to know where what values come from

mikenairn (approved):
Looks fine. Small query, but nothing that needs to hold up a merge.

config/deploy/local/manager_config_patch.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Deployment
+metadata:
+  name: controller-manager
+  namespace: system
+spec:
+  template:
+    spec:
+      containers:
+      - name: manager
+        imagePullPolicy: Never
```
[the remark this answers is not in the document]
sergioifg94: Yes. I assume that the local deploy will be built for the `make deploy` target, and everything will run in the kind clusters. If there's a reason to assume otherwise I'm happy to change it

README.md
```diff
@@ -31,25 +31,9 @@ When deploying the multi cluster traffic controller using the make targets the f
...
-            - name: AWS_SECRET_ACCESS_KEY
-            value: <AWS_SECRET_ACCESS_KEY>
-            - name: AWS_DNS_PUBLIC_ZONE_ID
-            value: <AWS_DNS_PUBLIC_ZONE_ID>
-            - name: ZONE_ROOT_DOMAIN
-            value: <ZONE_ROOT_DOMAIN>
-    ```
+1. Create two env files:
+    * One called `aws-credentials.env` containing **AWS_ACCESS_KEY_ID** and **AWS_SECRET_ACCESS_KEY**
+    * One called `controller-config` containing **AWS_DNS_PUBLIC_ZONE_ID** and **ZONE_ROOT_DOMAIN**
```
mikenairn, line 36: Just realised this should be `controller-config.env`. Anyway, can be updated in another PR.

PR #64, opened 2023-02-09
Update Readme for phase 1

R-Lawton:


README.md
```diff
@@ -1,80 +1,90 @@
...
+    make local_setup MCTC_WORKLOAD_CLUSTERS_COUNT=<NUMBER_WORKLOAD_CLUSTER>
+    ```
+
+3. Deploy the controller to the control plane cluster
+    ```sh
+    make deploy
+    ```
+4. Deploy the operator
+    ```sh
+    (export $(cat ./controller-config.env | xargs) && export $(cat ./aws-credentials.env | xargs) && make build install run
```
mikenairn, line 85: Does `make deploy` not deploy the operator on the cluster using the kustomize config? This is for running it locally no?

README.md
```diff
@@ -1,80 +1,90 @@
...
+            env:
+            - name: AWS_ACCESS_KEY_ID
+            value: <AWS_ACCESS_KEY_ID>
+            - name: AWS_SECRET_ACCESS_KEY
+            value: <AWS_SECRET_ACCESS_KEY>
+            - name: AWS_DNS_PUBLIC_ZONE_ID
+            value: <AWS_DNS_PUBLIC_ZONE_ID>
+            - name: ZONE_ROOT_DOMAIN
+            value: <ZONE_ROOT_DOMAIN>
+    ```
```
mikenairn, line 50: I don;t think we should have to do this. @sergioifg94 Had a PR to add in a local deploy which would have these already set. If this is what you have to do now, fair enough, but i;d suggest we try and get rid of it asap and have a nicer way of deploying it locally.

README.md
```diff
@@ -1,80 +1,90 @@
...
-make manifests
-```
 
-**NOTE:** Run `make --help` for more information on all potential `make` targets
+## Prerequisites:
+* AWS
+* Kind 
+    * `make kind`
+* yq 
+    * `make yq`
```
mikenairn, line 22: Why yq?

README.md
```diff
@@ -1,80 +1,90 @@
...
-```sh
-make docker-build docker-push IMG=<some-registry>/multi-cluster-traffic-controller:tag
-```
+## Description:
+When deploying the multi cluster traffic controller using the make targets the following will be created: 
+* Kind cluster(s)
+* Gateway API CRDs in the control plane cluster
+* Ingress controller
+* Cert manager
+* External DNS
```
mikenairn, line 9: We should remove this again, got added back for some reason, external dns is not being used.

README.md
R-Lawton: Im hoping Sergios pr gets merged before we tag phase 1 but right now it's not so I just wrote it as is. if it does get merged soon ill update it

README.md
R-Lawton: its needed for the webhook config to inject the ca bundle

README.md
R-Lawton: Cool

README.md
R-Lawton: Ooops copy paste error

README.md
mikenairn: This is fine. I would say though that when i see yq or jq being added a dependency of a deploy, it usually tells me we are doing something wrong. Ideally the kustomizations that we are using should override anything required without manipulating yaml directly.

mikenairn (approved):

PR #67, opened 2023-02-10
Specify local docker network

sergioifg94:
## Description

Currently, some of the configuration resources assume the IP of the control cluster being `172.17.0.2`. This is correct in the case of the local machine having no previous docker networks, as the creation of the kind clusters results in the creation of a bridge network and the IP range automatically assigned. However, in the scenario where there are other previous networks in the machine, the IP of the cluster will not match the IP specified in the resources.

As a first step to ensure consistency regardless of the current networks, modify the `local-setup` script to pre-create a network in the `172.32.0.0` subnet, and configure kind to use that network instead of automatically creating one. Update the resources that reference the control cluster IP to the new subnet

## Verification steps

In order to verify this, run the `local-setup` script and verify that the clusters are functioning correctly

Verify as well that the `mctc` docker network has been created
```sh
docker network inspect mctc
```

> A new `make` directive `local-cleanup` has been added to clean up resources that are created when running `local-setup`

Verify the `local-cleanup` by running `make local-cleanup` and ensuring the kind clusters and the network are deleted

### DIFF
--- a/hack/.cleanupUtils
+++ b/hack/.cleanupUtils
@@ -1,0 +1,21 @@
+# shellcheck shell=bash
+
+cleanClusters() {
+	# Delete existing kind clusters
+	clusterCount=$(${KIND_BIN} get clusters | grep ${KIND_CLUSTER_PREFIX} | wc -l)
+	if ! [[ $clusterCount =~ "0" ]] ; then
+		echo "Deleting previous kind clusters."
+		${KIND_BIN} get clusters | grep ${KIND_CLUSTER_PREFIX} | xargs ${KIND_BIN} delete clusters
+	fi	
+}
+
+cleanNetwork() {
+    # Delete the network
+    echo "Deleting mctc network"
+    docker network rm mctc || true
+}
+
+cleanup() {
+	cleanClusters
+	cleanNetwork
+}
--- a/hack/.cleanupUtils
+++ b/hack/.cleanupUtils
@@ -10,12 +10,23 @@ cleanClusters() {
 }
 
 cleanNetwork() {
-    # Delete the network
-    echo "Deleting mctc network"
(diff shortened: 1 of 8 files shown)


hack/.cleanupUtils
```diff
@@ -0,0 +1,21 @@
...
+cleanNetwork() {
+    # Delete the network
+    echo "Deleting mctc network"
+    docker network rm mctc || true
+}
+
+cleanup() {
+	cleanClusters
+	cleanNetwork
+}
```
mikenairn, line 21: Can this cleanup those local proxies for the dsashboards as well?

sergioifg94: I added the ability to stop the proxies in #31. We can get this merged first and then I update that PR to add a call to the function in the cleanup

mikenairn:
This looks fine to me, but you might want to check with someone on a Mac.

hack/.cleanupUtils
```diff
@@ -0,0 +1,32 @@
...
+      kill $p
+    done </tmp/dashboard_pids
+    rm /tmp/dashboard_pids
+  fi
+}
+
+cleanup() {
+  stopProxies
+	cleanClusters
+	cleanNetwork
```
mikenairn, line 31: Formatting looks a bit off

sergioifg94: Interesting. It looks good on my text editor. I'm rebasing and hopefully it fixes itself automatically

mikenairn: tabs maybe?

mikenairn (approved):
Looks fine to me. Assuming it works OK for those Max folks /lgtm

david-martin (approved):
Verified on macos.

PR #77, opened 2023-02-16
Development experience improvements

sergioifg94:
## Description

Changes to _slightly_ improve the development experience

### 1. Webhook base template

The webhook configuration is generated by setting the CA bundle into an existing template and applying it into the workload clusters. Currently the file is checked in version control which means the changed file might accidentally be commited. In order to avoid this, keep a separate base template checked in, and the effectively used one in .gitignore

### 2. Automatically generate the control plane secret for the agent

Currently the secret needs to be created manually. There is some existing functionality to create these type of secrets so refactor that functionality to be able to use it when deploying the agent into the cluster

## Verification steps

In order to verify these changes, follow the steps in the [README](https://github.com/Kuadrant/multi-cluster-traffic-controller/blob/9f43933c44e798cfa12458f0e936b8203a69874b/README.md) to deploy the agent and verify that:
1. Git doesn't detect any changes related to the webhook configuration
2. The secret for the agent has been created successfully
    ```sh
    kubectl get secret/control-plane-cluster -n mctc-system -o yaml
    ```


hack/.clusterUtils
```diff
@@ -0,0 +1,47 @@
...
+    {
+      "tlsClientConfig":
+        {
+          "insecure": false,
+          "caData": "${caData}",
+          "certData": "${certData}",
+          "keyData": "${keyData}"
+        }
+    }
+  name: ${clusterName}
```
philbrookes, line 31: it looks like this value will be "mctc-control-plane" but it should be the name of the workload cluster it's being put into (e.g. "mctc-workload-1").

sergioifg94: Hmm that is not as simple to automate as I thought it would be, since it means the secret will change for every cluster that it's created for. I think the only way to automate it will be to take it from the context or the kubeconfig. Any ideas?

philbrookes: in my PR, I updated the logic in `#8. Add workload clusters if MCTC_WORKLOAD_CLUSTERS_COUNT environment variable is set` to create the secret on each workload cluster, maybe this could do the same?

mikenairn: This is kind of related to the issues i was thinking kustomize would help us with, or at least we should be generating a set of k8s resources per cluster, that can be applied directly using `kustomize build x | kubectl apply -f`. Probably using an overlay per cluster that uses base resources in our config dir.

I'd say for now though, if we can just get this working, we can look at something like that later.

philbrookes: I expect to open my PR today, if you like we can just go with my solution. It's also not ideal, but saves you the effort of changing this PR @sergioifg94

sergioifg94: I'm following your suggestion for now @philbrookes and adding the secret in the local-setup. I'm testing it and I'll update the PR when I'm sure it works

sergioifg94: Just pushed the changes. The secret is now created automatically in the local-setup with the correct name

README.md
```diff
@@ -88,7 +86,10 @@ When deploying the multi cluster traffic controller using the make targets the f
 ```sh
 ```
-1. Update the secret in `config/agent/secret.yaml` with the correct credentials for the control plane `tmp/kubeconfigs/mctc-control-plane.kubeconfig` . (**N.B.** The server should be changed to the kind cluster address e.g. `https://127.0.0.1:46419`)
+1. Update the secret in `config/agent/secret.yaml` with the correct credentials for the control plane
+```sh
+make agent-secret
+```
```
mikenairn, line 92: This probably won't work in all cases. The host needs to change depending on where you are running the controller. If the controller is deployed, then this will work, but if it's running locally, you need to use the non internal one e.g. https://127.0.0.1:46419.

We should at least leave the comment here explaining that.

mikenairn:
@sergioifg94 You probably need to remove the `clear-dev-tls` target here https://github.com/Kuadrant/multi-cluster-traffic-controller/blob/main/Makefile#L78 since it's now committed.

README.md
philbrookes: what I did was create both secrets (mctc-workload-1 and mctc-workload-1-external) then you pass in which one to use as an argument to the agent (external for deployment and regular for local dev)

philbrookes: though actually, I'd prefer to merge this now, and then I can open the spec syncing part of the dumb-agent on top of this, and it can be tweaked  after.

sergioifg94: I added a commit to allow passing the `LOCAL_ACCESS` environment variable so, if it's set, it uses the `127.0.0.1` address (doesn't add the `--internal` flag), and updated the README to include it when running the agent locally

philbrookes (approved):

philbrookes:
alright, let's get this merged and we can iterate on it from there, it's a good starting point, I think.

PR #81, opened 2023-02-17
Add MetalLB to local workload clusters for Services of type LoadBalancer

david-martin:
This change enables Services of type `LoadBalancer` in a local development environment to be assigned an IP Address.
This is required for local istio Gateways running in workload clusters, as it uses a Service of type `LoadBalancer`

To verify, bring up a local development environment with 2 workload clusters (2 to verify the IPAddressPool subnet works for each new workload cluster)
```
MCTC_WORKLOAD_CLUSTERS_COUNT=2 make local-setup
```
Deploy a sample app to both workload clusters, including a Service of type `LoadBalancer`
```
kubectl --context kind-mctc-workload-1 apply -n default -f https://gist.githubusercontent.com/david-martin/d30bd544436be7244d8ecaefc848cfd7/raw/ca59823ac856e6a3815c42726c0c556a8cd586f0/sample_app_with_loadbalancer.yaml
kubectl --context kind-mctc-workload-2 apply -n default -f https://gist.githubusercontent.com/david-martin/d30bd544436be7244d8ecaefc848cfd7/raw/ca59823ac856e6a3815c42726c0c556a8cd586f0/sample_app_with_loadbalancer.yaml
```

Verify both Services have an external IP Address assigned in the corresponding subnet range e.g. `172.32.200.0` and `172.32.201.0`
```
kubectl --context kind-mctc-workload-1 -n default get svc echo -o=jsonpath='{.status.loadBalancer.ingress[0].ip}'
kubectl --context kind-mctc-workload-2 -n default get svc echo -o=jsonpath='{.status.loadBalancer.ingress[0].ip}'
```
Verify L2 routing is working. You should get a HTTP 200 response.
```
curl $(kubectl --context kind-mctc-workload-1 -n default get svc echo -o=jsonpath='{.status.loadBalancer.ingress[0].ip}') -v
curl $(kubectl --context kind-mctc-workload-2 -n default get svc echo -o=jsonpath='{.status.loadBalancer.ingress[0].ip}') -v
```

### DIFF
--- a/hack/local-setup.sh
+++ b/hack/local-setup.sh
@@ -27,6 +27,7 @@ KIND_CLUSTER_CONTROL_PLANE="${KIND_CLUSTER_PREFIX}control-plane"
 KIND_CLUSTER_WORKLOAD="${KIND_CLUSTER_PREFIX}workload"
 
 INGRESS_NGINX_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/ingress-nginx
+METALLB_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/metallb
 CERT_MANAGER_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/cert-manager
 EXTERNAL_DNS_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/external-dns
 ARGOCD_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/argocd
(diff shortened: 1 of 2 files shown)


hack/local-setup.sh
```diff
@@ -46,6 +47,34 @@ deployIngressController () {
...
+  namespace: metallb-system
+spec:
+  addresses:
+  - 172.32.${metalLBSubnet}.0/24
+---
+apiVersion: metallb.io/v1beta1
+kind: L2Advertisement
+metadata:
+  name: empty
+  namespace: metallb-system
```
roivaz, line 74: just curious, what is this for?

david-martin: Good question.
I tried these changes locally without the L2Advertisement CR.
The curl request to the service loadbalancer IP Address times out.

My understanding is this resource instructs metallb to use Layer 2 protocols (ARP in this case) to advertise that the IP address is available on the network. ARP (Address Resolution Protocol) is used by L2 devices/interfaces to discover the MAC address for a given IP so it can construct the appropriate L2/Ethernet frame header with the destination MAC address.

From https://metallb.universe.tf/concepts/
> After MetalLB has assigned an external IP address to a service, it needs to make the network beyond the cluster aware that the IP “lives” in the cluster. MetalLB uses standard networking or routing protocols to achieve this, depending on which mode is used: ARP, NDP, or BGP.

BGP is also supported, but I haven't tried that (I'm also not sure if it's necessary as I believe it's kind of a flat network that's established, so no routing is required)

roivaz: Thanks for looking that up! It's actually very little setup to have l4 load balancers working in local dev envs, very interesting!

roivaz:
Works like a charm!

roivaz (approved):
lgtm

hack/local-setup.sh
mikenairn: Not a big deal, but could these resources be added to the kustomize config itself instead of applying them directly here. The use of the `metalLBSubnet` probably makes that harder, but it would tidy this up a bit.

mikenairn (approved):
Looks fine, small comment but nothing to stop a merge.

hack/local-setup.sh
david-martin: Capturing the outcome of offline conversation.
We've used the `replacements` field before in kustomize, but that pattern doesn't apply here.
Generally, kustomize doesn't help when working with variables (deliberately so, as it pushes for file based configuration).
It would be possible to have a base IPAddressPool resource in the config dir, and then use yq to overwrite `spec.addresses` for each install of metallb. However, given this change is almost certainly going to be local dev only (load balancers are available in cloud providers), that change doesn't offer anything additional outside of what it already does.

Issue #52, opened 2023-02-02
Support a placement label on the gateway resource


david-martin:

maleck13:
covered now by OCM

maleck13:
https://github.com/Kuadrant/multicluster-gateway-controller/pull/175

maleck13:
https://github.com/Kuadrant/multicluster-gateway-controller/pull/175

Issue #44, opened 2023-02-02
Cluster Registration


maleck13:
**What**

We want to be able to register clusters with the control plane within a given tenant NS . These clusters will be represented by secrets. These cluster secrets will be the place where additional information and context is added about the cluster. 

The setup of this can be done via a script initially. 

This should also spit our yaml for the sync agent and webhook configuration

**Constraints**

There isn't a requirement to setup a the tenant as part of this work, this is just focused on registering the cluster within a tenant.

- [ ] #68
- [ ] Add logic to the registration script that uses a KUBECONFIG to create a new service account and create a rolebinding in the tenant ns
- [ ] Add logic that uses the SA and outputs a deployment etc to setup the syncer

maleck13:
@philbrookes is there specific things you see as needed for the sync agent deployment yaml?

philbrookes:
I'd say this is a good starting point: https://github.com/Kuadrant/multi-cluster-traffic-controller/tree/main/config/agent

It's what is used to deploy the current ingress-agent to a cluster.

Issue #43, opened 2023-02-02
Basic Sync Support


maleck13:
**What**

We are adding a sync agent to reflect certain resources defined at the control plane into the data plane. This syncer will also be responsible for reflecting the status of those resources back into the control plane and applying transforms added via the controllers in the control plane. This syncer will lean heavily on what we have learned through use of the KCP sync component

At the end of this work the control plane should no longer talk to the data plane.

**Open Questions**
- how do we choose which namespace on the cluster to sync to? Do we generate a namespace like kcp did based on the tenant namespace?
- Initially we can sync to a known namespace (IE istio-system) 

**Outcome**

We should have basic sync capabilities for gateways and TLS secrets based on access provided by service account secrets. We should capture, patterns, concepts and remaining questions as a Design doc

- [x] Sync DNSRecords based on Ingress #70 
- [x] Remove cluster watch support #70 
- [x] Remove webhook reconcile #70 
- [x] Add capability to sync gateways and secrets
- [x] Add capability to report back the status of a given gateway as a multi-cluster status

maleck13:
I think we need a mechanism for the control plane to say
1) this Gateway is ready to sync
2) this gateway is for your cluster

So the gateway admin will choose a placement label but we then need to reflect the cluster choice on the gateway resource somehow have the sync component see this and then sync down the spec of the gateway

@philbrookes

maleck13:
marking this as done in m2 and moving additonal MVP work to #108

PR #84, opened 2023-02-20
Reconcile Istio Gateway

### LINKED ISSUE #45: Gateway certificate sync

### LINKED ISSUE #46: Add support for reconciling Gateway Listeners
**What**

The gateway controller in the traffic controller, will need to reconcile the listeners defined in the gateway. For each listener the following requirements have been identified

- Check if the host is part of a managed zone for that tenant
- If not then it can be ignored
- ensure there is a certificate for the host specified (if it is a wildcard host the cert should be a wildcard cert)
- ensure there is a DNSRecord based on the LB status in the gateway (only once the certificate is ready)

david-martin:
closes #45, closes #46 

*Steps to verify locally using Route53*

**IMPORTANT** These steps assume your local ./aws-credentials.env file has all values (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_REGION`) set and ./controller-config.env file has the right values for `ZONE_ROOT_DOMAIN` and `AWS_DNS_PUBLIC_ZONE_ID` that match dev.hcpapps.net.

**IMPORTANT** Change the hostname in the Gateway and HttpRoute to avoid reusing `myapp.dev.hcpapps.net` e.g. Use `<myname>.dev.hcpapps.net`

```
cp config/samples/gateway.yaml ./gateway.yaml
cp config/samples/echo_httproute.yaml ./echo_httproute.yaml
vi gateway.yaml
vi echo_httproute.yaml
```

**Terminal 1** Local setup & start the controller in the control plane cluster
```
(export $(cat ./controller-config.env | xargs) && export $(cat ./aws-credentials.env | xargs) && make build-controller install run-controller
```

**Terminal 2** Start the syncer in the workload cluster
```
make kind-load-syncer
make deploy-syncer
```

**Terminal 3** Create the GatewayClass & Gateway
```
export KUBECONFIG=./tmp/kubeconfigs/mctc-control-plane.kubeconfig
kubectl apply -f config/samples/gatewayclass.yaml
kubectl apply -n mctc-tenant -f ./gateway.yaml
```

**Terminal 2** Deploy the sample application with a HttpRoute
```
kubectl apply -n mctc-downstream -f ./echo_httproute.yaml
```

**Terminal 3** Wait for the Gateway to be programmed, then curl the hostname specified in the HttpRoute. You should get a 200 response.

```
kubectl wait --for=condition=Programmed --timeout=60s gateway example-gateway -n mctc-tenant
curl -k -s -o /dev/null -w "%{http_code}"  https://<myname>.dev.hcpapps.net
```

TODO: 

- [x] Rebase after #60 (#35)- Refactor zone selection/validation 
- [x] Rebase after sync agent spec #89 (#43) - Implement sync annotation setting
- [x] Rebase after sync agent status #96 (#43) - Implement status retrieval
- [x] Working integration test
- [x] Decouple from ingress controller that the ingress agent uses (initally re-used some of that code for a steel thread)
- [x] Fetch Gateway IP address(es) from Gateway status in `GetDNSTargets`
- [x] Check Gateway status for attached routes before creating DNS record (internal discussion link about attachedRoutes field in GatewayStatus https://redhat-internal.slack.com/archives/C047YHJJG2D/p1676550730816259)

Out of scope/follow up
- #85
- #52 
- Rebase after cluster registration changes #44 - Gateway cluster selector (sync to all clusters initially)

### DIFF
--- a/pkg/controllers/gateway/gateway_controller.go
+++ b/pkg/controllers/gateway/gateway_controller.go
@@ -21,10 +21,17 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+	"time"
 
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/_internal/conditions"
 	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/_internal/slice"
+	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/apis/v1alpha1"
+	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/syncer"
+	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/traffic"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -38,25 +45,36 @@ const (
 	GatewayClusterLabelSelectorAnnotation = "kuadrant.io/gateway-cluster-label-selector"
 )
 
+type HostService interface {
+	CreateDNSRecord(ctx context.Context, hostKey string, managedZone *v1alpha1.ManagedZone) (*v1alpha1.DNSRecord, error)
+	GetManagedZone(ctx context.Context, t traffic.Interface) (*v1alpha1.ManagedZone, error)
+	AddEndPoints(ctx context.Context, t traffic.Interface) error
+	RemoveEndpoints(ctx context.Context, t traffic.Interface) error
+}
+
+type CertificateService interface {
+	EnsureCertificate(ctx context.Context, host string, owner metav1.Object) error
+	GetCertificateSecret(ctx context.Context, host string, namespace string) (*corev1.Secret, error)
+}
+
+type GatewayHelper interface {
+	GetGatewayStatuses(ctx context.Context) []gatewayv1beta1.GatewayStatus
+	GetListenerNameByHost(host string) string
+	GetListenerStatusesByListenerName(ctx context.Context, listenerName string) []gatewayv1beta1.ListenerStatus
+}
+
 // GatewayReconciler reconciles a Gateway object
 type GatewayReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	Scheme       *runtime.Scheme
+	Certificates CertificateService
+	Host         HostService
 }
 
 //+kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gateways,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gateways/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gateways/finalizers,verbs=update
 
-// Reconcile is part of the main kubernetes reconciliation loop which aims to
-// move the current state of the cluster closer to the desired state.
-// TODO(user): Modify the Reconcile function to compare the state specified by
-// the Gateway object against the actual cluster state, and then
-// perform operations to make the cluster state reflect the state specified by
-// the user.
-//
-// For more details, check Reconcile and its Result here:
-// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.13.1/pkg/reconcile
 func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	log := log.FromContext(ctx)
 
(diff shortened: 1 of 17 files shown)


pkg/traffic/traffic.go
```diff
@@ -25,10 +25,9 @@ type Interface interface {
 	GetNamespaceName() types.NamespacedName
 	AddTLS(host string, secret *corev1.Secret)
 	HasTLS() bool
-	GetTLS() []TLSConfig
```
david-martin, line 28: Unused

pkg/tls/service.go
```diff
@@ -31,7 +31,7 @@ func NewService(controlClient client.Client, defaultCtrlNS, defaultIssuer string
 }
 
 func (s *Service) EnsureCertificate(ctx context.Context, host string, owner metav1.Object) error {
-	cert := s.certificate(host, s.defaultIssuer, s.defaultCtrlNS)
+	cert := s.certificate(host, s.defaultIssuer, owner.GetNamespace())
```
david-martin, line 34: use the tenant ns

david-martin:
I'd appreciate your review & verification here.

- @philbrookes in particular, the changes to the syncer controller to hardcode a couple of transforms
- @mikenairn I've included your commit to get MZs working, but I've also changed things so that only hosts in a gateway listener that match an MZ will be considered for further reconciling. I've also made a change to allow assigning endpoints for an individual host, as a gateway may have multiple listeners with only some hosts ready/attached.

Things not included in this PR are #85, #52, integration with #44, and handling of Gateway deletion.
I would seem them landing in priority order after this PR merges.

pkg/syncer/spec/controller.go
```diff
@@ -246,6 +246,30 @@ func (c *Controller) applyToDownstream(ctx context.Context, gvr schema.GroupVers
 	}
 
 	//TODO jsonpatch applied
+	// HARDCODED PATCHES - REMOVE THESE
```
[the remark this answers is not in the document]
philbrookes: this looks like a suitable workaround to me. I'm going to be working on the annotation transforms this week, so hopefully we'll be able to migrate this shortly.

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -70,88 +88,237 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		// TODO: Handle any cleanup if there were previously clusters
+		return metav1.ConditionFalse, clusters, false, nil
 	}
+	trafficAccessor := traffic.NewGateway(gateway)
+	gatewayHelper := trafficAccessor.(GatewayHelper)
+	allHosts := trafficAccessor.GetHosts()
+	hostsWithMZ := []string{}
+
+	// First, filter out hosts that are *not* in a Managed Zone
+	// Those can be left alone
```
philbrookes, line 161: It looks like this comment might be out of date? I think this for block is now generating an empty DNS record for each subdomain that exists in a valid managed zone?

philbrookes: oh I see it is also building hostsWithMZ at the end.

pkg/dns/service.go
```diff
@@ -314,6 +322,33 @@ func (s *Service) GetManagedZone(ctx context.Context, t traffic.Interface) (*v1a
...
 }
 
+// GetManagedZoneForDomain returns a ManagedZone for the given domain if one exists.
+// Currently, this returns the first matching ManagedZone found in the traffic resources own namespace, or if none is found,
+// it looks for the first matching one listed in the default ctrl namespace.
+func (s *Service) GetManagedZoneForDomain(ctx context.Context, domain string, t traffic.Interface) (*v1alpha1.ManagedZone, error) {
+	var managedZones v1alpha1.ManagedZoneList
+
+	if err := s.controlClient.List(ctx, &managedZones, client.InNamespace(t.GetNamespace()), client.MatchingFields{"spec.domainName": domain}); err != nil {
+		log.Log.Error(err, "unable to list managed zones in traffic resource NS")
```
philbrookes, line 332: Should we be able to pull the controller-runtime log logger from the context here?

pkg/traffic/gateway.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/_internal/slice"
+	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/apis/v1alpha1"
+	status "github.com/Kuadrant/multi-cluster-traffic-controller/pkg/syncer/status"
+)
+
+func NewGateway(g *gatewayv1beta1.Gateway) Interface {
+	return &Gateway{Gateway: g}
+}
+
+type Gateway struct {
```
philbrookes, line 26: It looks like we haven't defined an interface for this object? Should we be using one? Are gateways abstracted like this, for the same reason as ingresses (i.e. routes might come in future).

pkg/controllers/gateway/gateway_controller.go
david-martin: Yes, exactly

pkg/dns/service.go
david-martin: Yes, I'll update this

pkg/dns/service.go
```diff
@@ -79,13 +79,25 @@ func (s *Service) GetDNSRecords(ctx context.Context, traffic traffic.Interface)
...
 				Name:      host,
-				Namespace: s.defaultCtrlNS,
+				Namespace: traffic.GetNamespace(),
 			},
 		}
 		if err := s.controlClient.Get(ctx, client.ObjectKeyFromObject(record), record); err != nil {
 			if k8serrors.IsNotFound(err) {
-				log.Log.V(10).Info("no dnsrecord found for host ", "host", record.Name)
-				continue
+				log.Log.V(10).Info("no dnsrecord found for host ", "host", record.Name, "namespace", record.Namespace)
```
david-martin, line 87: @mikenairn I've updated this to check the tenant namespace first for a matching MZ, then fallback to the default ns

pkg/traffic/gateway.go
david-martin: I've attempted to use the traffic Interface for the Gateway resource as much as possible.
Is there some wiring that I've missed?

pkg/dns/service.go
mikenairn: Hmm, do we know what managed zone is being used by this point? It should only be looking in that ns i think since the record will always be in the MZ ns.

mikenairn: but it can't check the defaultNS that's for sure

pkg/traffic/gateway.go
philbrookes: I missed the reused traffic interface, that's cool.

philbrookes (approved):

pkg/dns/service.go
david-martin: GetDNSRecords is called from 3 places:
- EnsureManagedHost (ingress specific)
- AddEndPoints
- RemoveEndpoints

RemoveEndpoints is not used anywhere yet, but I'd imagine you'd call it when a Gateway (or Ingress) is deleted, or maybe when a specific listener block is removed from a Gateway.
In those cases the MZ would have to be looked up from the host. That being said, there's an edge case where an MZ could be added after the fact that matches the host, but the DNSRecord was created against a different MZ.

AddEndPoints isn't used by Ingress, as there's patching that handles that.
For Gateways, the MZ is known. If we were to pass it in to GetDNSRecords, the logic in the gateway reconciler would need to be refactored somewhat to map hosts to MZs early on and keep that map around until AddEndPoints has to be called for each host.

I'm not sure about EnsureManagedHost.

david-martin: @mikenairn There's a typo in my first comment here.
~check the tenant namespace first for a matching MZ~
should be 'check the tenant namespace first for a matching DNSRecord by host'

david-martin: I discovered a bug here where it returns too early instead of adding the DNSRecord to the array. I've pushed a small change https://github.com/Kuadrant/multi-cluster-traffic-controller/pull/84/commits/748dcd0aa5466d102ab3e1294c971308029983bb

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -70,88 +88,237 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	for _, host := range allHosts {
+		hostParts := strings.SplitN(host, ".", 2)
```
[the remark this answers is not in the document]
maleck13: We  might want to use a trie tree structure, example https://github.com/detailyang/domaintree-go. Also I think it might be worth encapsulating in the `GetManagedZoneForDomain` method

pkg/dns/service.go
mikenairn: I created a follow on issue for this https://github.com/Kuadrant/multi-cluster-traffic-controller/issues/111

pkg/controllers/gateway/gateway_controller.go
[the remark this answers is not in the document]
mikenairn: Good point. Probably worth moving that logic into the method for ease of testing it. Maybe rename it `GetManagedZoneForHost` and just pass the host straight it.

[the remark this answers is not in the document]
mikenairn: The logic btw will work for a co.uk domain, or any host with multiple subdomains, but it is simplistic in that it will only knock off the first label and look for an exactly matching MZ for the rest e.g. foo.bar.baz.cbrooks.co.uk, would look for an MZ for `bar.baz.cbrooks.co.uk` and create a record for `foo` in it.

If we want to be able to work down the tree and for example find an MZ for `cbrooks.co.uk` and create a record for `foo.bar.baz` then we would need to update the logic in some way to allow this.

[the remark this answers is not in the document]
maleck13: This can be added as a follow up

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -70,88 +88,237 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		}
+		subDomain := hostParts[0]
+		parentDomain := hostParts[1]
+
+		log.V(2).Info("getting managed zone for parent domain of host", "host", host, "parentDomain", parentDomain, "subDomain", subDomain)
+		managedZone, err := r.Host.GetManagedZoneForDomain(ctx, parentDomain, trafficAccessor)
+		if err != nil {
+			return metav1.ConditionUnknown, clusters, false, err
+		}
+		if managedZone == nil {
```
maleck13, line 175: This is probably something we would want to record in the status lets add a follow up all around the CP gateway status

pkg/controllers/gateway/gateway_controller.go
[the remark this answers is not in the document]
mikenairn: Created a follow up https://github.com/Kuadrant/multi-cluster-traffic-controller/issues/112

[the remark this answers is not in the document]
maleck13: https://github.com/Kuadrant/multi-cluster-traffic-controller/issues/112

pkg/syncer/spec/controller.go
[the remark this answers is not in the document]
maleck13: Do we need an issue to track bringing the control plane upto date once the syncer supports transforms?

maleck13 (approved):
I believe we have captured all the needed follow up work. As this has been reviewed and verified. I think it is fine to merge at this point

david-martin:
Thanks @philbrookes @mikenairn @maleck13 

Summary of follow ups:
- #85
- #52
- #111
- #112 
- #113 
- #114

PR #86, opened 2023-02-22
GH-32: Add proxy to enable webhooks when running locally

### LINKED ISSUE #32: Enable webhooks when running the controller locally
## What

The changes introduced in #9 allow the webhook system to be set up only when the controller is running in-cluster. The webhook configuration will not be reconciled when running locally

## Why

In order to continously verify and develop the webhook system, it would be better to include the ability for the webhooks to work when the controller runs locally

## How

Investigate an approach to expose the webhook server (which will run in the local host) with a valid CA bundle that be referenced in the webhook configuration.

### Possible implementations

1. One alternative to implement this is exposing a container using the webhook TLS certificates, which proxies the requests to `172.32.0.1` (the IP address to the local host). This would require:
    * A container image that can be configured to proxy requests to a given address
    * Configuring the webhook ingress to expose that container image instead of the controller
2. Another alternative is modifying the controller to receive the TLS certificates location as a parameter and, in that case, run the webhook server with them. Then all that needs to be done is configuring the WebhookConfiguration to point to `172.32.0.1`. In that case the host of the certificates will need to change depending on whether we're running the webhook server in-cluster or in the localhost

sergioifg94:
## Description

> Closes #32 

Add a command to deploy a simple proxy in the control cluster, which redirects requests to the local host, and is exposed as the webhook endpoint. This allows the workload clusters to maintain the same webhook configuration while running the controller in the localhost

## Verification

Follow the instructions in the [README](https://github.com/Kuadrant/multi-cluster-traffic-controller/blob/4b9e56c32526ee5ca127a9b8711aaca774214d98/README.md#2-running-locally) to run the controller locally and ensure the webhooks are reachable

### DIFF
--- a/config/webhook-setup/proxy/kustomization.yaml
+++ b/config/webhook-setup/proxy/kustomization.yaml
@@ -1,0 +1,23 @@
+namespace: multi-cluster-traffic-controller-system
+
+namePrefix: mctc-
+
+resources:
+- namespace.yaml
+- deployment.yaml
+- service.yaml
+- ../control/ingress.yaml
+
+configMapGenerator:
+  - name: webhook-proxy-conf
+    files:
+      - nginx.conf
+
+secretGenerator:
+- name: ingress-tls
+  files:
+    - ../control/tls/tls.crt
+    - ../control/tls/tls.key
+
+generatorOptions:
+  disableNameSuffixHash: true
--- a/Makefile
+++ b/Makefile
@@ -177,6 +177,10 @@ clear-dev-tls:
 	-rm -f $(DEV_TLS_CRT)
 	-rm -f $(DEV_TLS_KEY)
 
+.PHONY: webhook-proxy
+webhook-proxy: kustomize
+	$(KUSTOMIZE) --load-restrictor LoadRestrictionsNone build config/webhook-setup/proxy | kubectl apply -f -
+
 ##@ Build Dependencies
 
 ## Location to install dependencies to
--- a/config/webhook-setup/proxy/deployment.yaml
+++ b/config/webhook-setup/proxy/deployment.yaml
@@ -1,0 +1,32 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: webhook-proxy
+  namespace: system
+spec:
+  selector:
+    matchLabels:
+      control-plane: webhook-proxy
+  replicas: 1
+  template:
+    metadata:
+      annotations:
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -63,6 +63,13 @@ When deploying the multi cluster traffic controller using the make targets the f
     make local_setup MCTC_WORKLOAD_CLUSTERS_COUNT=<NUMBER_WORKLOAD_CLUSTER>
     ```
 
+1. To ensure the webhook server is reachable from the workload cluster. Deploy
+   the webhook proxy
+
+   ```sh
+   make webhook-proxy
+   ```
```
mikenairn, line 71: This would be better being installed as part of the local-setup script since it's intended purpose is to setup a dev env locally using kind. Presumably though we don't actually want this deployed if you are actually running the controller on the cluster? If just had the proxy listen on a different port from the default(8082) could we not just install this all the time on the kind control cluster?

sergioifg94: I thought of it but, as you say, the main reason against it is the case that we'd want to run the controller in-cluster.
The problem of using a different port is that in that case we'd have to go and modify the port in the webhook configuration if we want to run it in cluster.

I haven't tested it, but I think it might still be possible to deploy the proxy by default in the local-setup and, if running `make deploy`, it would update the service to point to the controller instead of the proxy, so it would still work without any manual changes

sergioifg94: I just tested it and it does work, running `make deploy` after the proxy has been deployed. It configures the service to point to the controller, and everything else can remain the same. The command just fails to create the ingress but that's ok because the spec is the same as the proxy. I'm going to figure out if there's a way to not fail the command and include the deployment of the proxy in the local-setup. How does that sound?

config/webhook-setup/proxy/kustomization.yaml
```diff
@@ -0,0 +1,21 @@
...
+- service.yaml
+- ../control/ingress.yaml
+
+configMapGenerator:
+  - name: webhook-proxy-conf
+    files:
+      - nginx.conf
+
+secretGenerator:
+- name: ingress-tls
```
mikenairn, line 15: Looks like this will have to be called `mctc-ingress-tls`. The default kustomize config uses a name prefix of `mctc`, but this one isn;t, and isn;t called from another one, so you need to be explicit with the name here if you are sharing the same ingress setup for both.

mikenairn (approved):

PR #87, opened 2023-02-22
Local setup and readme updates

mikenairn:
Updates to local setup and docs related to running the controller and agent locally.

* Use tmp dir for temporary files
* Remove externaldns
* Remove samples/syncer (Same as echo-service and the secret should be created elsewhere as part of the agent deployment)
* Add clean make target (Removes all temp files and binaries creates from make commands and scripts)
* Add build make target (Builds both the controller and agent)
* Use Make as intended for file generation.


Makefile
```diff
@@ -164,14 +163,19 @@ undeploy-agent: ## Undeploy controller from the K8s cluster specified in ~/.kube
...
+DEV_TLS_DIR = config/webhook-setup/control/tls
+DEV_TLS_CRT ?= $(DEV_TLS_DIR)/tls.crt
+DEV_TLS_KEY ?= $(DEV_TLS_DIR)/tls.key
+
 .PHONY: dev-tls
-dev-tls:
-	test -s config/webhook-setup/control/tls/tls.crt || openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout config/webhook-setup/control/tls/tls.key -out config/webhook-setup/control/tls/tls.crt -subj "/C=IE/O=Red Hat Ltd/OU=HCG/CN=webhook.172.32.0.2.nip.io" -addext "subjectAltName = DNS:webhook.172.32.0.2.nip.io"
+dev-tls: $(DEV_TLS_CRT) ## Generate dev tls webhook cert if necessary.
+$(DEV_TLS_CRT):
+	openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout $(DEV_TLS_KEY) -out $(DEV_TLS_CRT) -subj "/C=IE/O=Red Hat Ltd/OU=HCG/CN=webhook.172.32.0.1.nip.io" -addext "subjectAltName = DNS:webhook.172.32.0.1.nip.io"
```
mikenairn, line 173: No real reason to use `test` here, we can just use a non phony target that expects a file to be generated. I see other targets using `test`, but they appear to do other tests such as check the version etc...

README.md
```diff
@@ -37,8 +37,8 @@ When deploying the multi cluster traffic controller using the make targets the f
 
 1. Build the controller image and load it into the control plane
     ```sh
-    export KUBECONFIG=./hack/kubeconfigs/mctc-control-plane.kubeconfig
-    make kind-load-controller
+   kubectl config use-context kind-mctc-control-plane 
+   make kind-load-controller
```
mikenairn, line 41: Sticking with 1 KUBECONFIG, the users default, and just switching context is probably going to cause less issues. If you run `local-setup` in a terminal you have previously changed the KUBECONFIG in, it wouldn't work.

README.md
```diff
@@ -82,19 +83,21 @@ When deploying the multi cluster traffic controller using the make targets the f
...
+1. Apply the control plane secret
 ```sh
-kubectl create -f samples/syncer/secret.yaml
+kubectl create ns mctc-system
+kubectl create -f config/agent/secret.yaml
 ```
 1. Run the agent locally:
 ```sh
-make run-agent
+make build-agent run-agent
```
mikenairn, line 100: There are actually a few other things required to make all this work locally, but I'm not doing that in this PR.

mikenairn: Related https://github.com/Kuadrant/multi-cluster-traffic-controller/pull/86

Makefile
```diff
@@ -3,13 +3,6 @@
 TAG ?= latest
 CONTROLLER_IMG ?= controller:$(TAG)
 AGENT_IMG ?= agent:$(TAG)
-AWS_ENVVARS_FILE ?= aws-credentials.env
-CONTROLLER_ENVVARS_FILE ?= controller-config.env
-
-include $(CONTROLLER_ENVVARS_FILE)
-include $(AWS_ENVVARS_FILE)
-export AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_DNS_PUBLIC_ZONE_ID ZONE_ROOT_DOMAIN
-
```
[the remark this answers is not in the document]
philbrookes: it was just a QOL of change so I didn't have to keep remembering to export this files when running make commands. I don't think they're required any longer.

README.md
philbrookes: It's annoying using context when you want to have  one terminal on a workload cluster and another on the control plane cluster, as changing context in one changes it in both.

philbrookes: oh, I see, we're keeping the kubeconfig files, just not using them from the makefile

philbrookes (approved):

README.md
mikenairn: true, if you want to configure 2 terminals one for each cluster, switching contexts isn't ideal. This is just docs changes, so not a huge issue either way but had it mess up a few times if you run local-setup after setting KUBECONFIG to one that gets pulled out of a kind cluster.

PR #89, opened 2023-02-23
dumb sync - spec

philbrookes:
### setup
- `export LOCAL_ACCESS="true"`
- run a fresh: `make local-setup MCTC_WORKLOAD_CLUSTERS_COUNT=1`
- open 3 terminals
  - in terminal 1 target the control plane: `export KUBECONFIG=./tmp/kubeconfigs/mctc-control-plane.kubeconfig`
  - in terminal 2 and 3 target the workload cluster: `export KUBECONFIG=./tmp/kubeconfigs/mctc-workload-1.kubeconfig`
- in terminal 2, run: `make run-syncer`

### assert annotation is required
- in terminal 1: `kubectl create -f config/samples/gateway.yaml`
- **expectation:** nothing should happen

### assert annotation is observed
- in terminal 1: edit the gateway and add the annotation: `mctc-sync-agent/kind-mctc-workload-1: "true"`
- there should be activity in terminal 2
- in terminal 3: `kubectl get gateway example-gateway -n mctc-downstream -o yaml`
- **expectation:** The  gateway should be present in the downstream cluster

### assert updates are applied
- in terminal 1: edit the spec of the gateway
- there should be activity in terminal 2
- in terminal 3: `kubectl get gateway example-gateway -n mctc-downstream -o yaml`
- **expectation:** the gateway spec should update to match the change

## Notes
- deletion is not implemented
- patch annotations are not implemented
- the downstream namespace is hardcoded `mctc-downstream`
- status syncing up is not implemented
- A lot of this code is airlifted from the KCP syncer, and modified, in some cases to simplify it, in some cases to refactor away from the expectation of workspaces.

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -58,9 +58,9 @@ func main() {
 	var controlPlaneConfigSecretName string
 	var controlPlaneConfigSecretNamespace string
 
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8084", "The address the metric endpoint binds to.")
-	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8085", "The address the probe endpoint binds to.")
-	flag.StringVar(&controlPlaneConfigSecretName, "control-plane-cluster", "control-plane-cluster", "The name of the secret with the control plane client configuration")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	flag.StringVar(&controlPlaneConfigSecretName, "control-plane-config-name", "control-plane-cluster", "The name of the secret with the control plane client configuration")
 	flag.StringVar(&controlPlaneConfigSecretNamespace, "control-plane-config-namespace", "default", "The namespace containing the secret with the control plane client configuration")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
--- a/cmd/syncer/main.go
+++ b/cmd/syncer/main.go
@@ -1,0 +1,191 @@
+/*
+Copyright 2022 The MultiCluster Traffic Controller Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/dynamic/dynamicinformer"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	corev1 "k8s.io/api/core/v1"
+	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
+	// to ensure that exec-entrypoint and run can make use of them.
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
+	//+kubebuilder:scaffold:imports
+
+	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/_internal/clusterSecret"
+	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/syncer"
+	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/syncer/spec"
+)
+
+var (
+	scheme            = runtime.NewScheme()
+	setupLog          = ctrl.Log.WithName("setup")
+	NEVER_SYNCED_GVRs = []string{"pods"}
+)
+
+const (
+	DEFAULT_RESYNC = 0
+	SYNCER_THREADS = 1
+)
+
+type arrayFlags []string
+
+func (i *arrayFlags) String() string {
+	return strings.Join(*i, ",")
+}
+
+func (i *arrayFlags) Set(value string) error {
+	*i = append(*i, value)
+	return nil
+}
+
+func init() {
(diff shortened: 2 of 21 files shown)


david-martin:
First pass with trying it out locally seems OK.

1 liner to annotate the gateway `kubectl annotate gateway example-gateway mctc-sync-agent/kind-mctc-workload-1=true`

logs
```
INFO[0001] starting syncer: kind-mctc-workload-1
INFO[0074] processing key default/example-gateway, gvr: gateway.networking.k8s.io/v1beta1, Resource=gateways
INFO[0074] got upstream namespace: default
INFO[0074] Created downstream namespace for upstream namespace
INFO[0074] Updated upstream resource with syncer finalizer
INFO[0074] processing key default/example-gateway, gvr: gateway.networking.k8s.io/v1beta1, Resource=gateways
INFO[0074] got upstream namespace: default
INFO[0074] Upstream object is intended to be removedintendedToBeRemovedFromLocationfalsestillOwnedByExternalActorForLocationfalse
INFO[0074] Upserted upstream resource to downstream
INFO[0390] processing key default/example-gateway, gvr: gateway.networking.k8s.io/v1beta1, Resource=gateways
INFO[0390] got upstream namespace: default
INFO[0390] Upstream object is intended to be removedintendedToBeRemovedFromLocationfalsestillOwnedByExternalActorForLocationfalse
INFO[0390] Upserted upstream resource to downstream
```

This entry looks odd `removedintendedToBeRemovedFromLocationfalsestillOwnedByExternalActorForLocationfalse`

cmd/agent/main.go
```diff
@@ -58,9 +58,9 @@ func main() {
 	var controlPlaneConfigSecretName string
 	var controlPlaneConfigSecretNamespace string
 
-	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8084", "The address the metric endpoint binds to.")
-	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8085", "The address the probe endpoint binds to.")
-	flag.StringVar(&controlPlaneConfigSecretName, "control-plane-cluster", "control-plane-cluster", "The name of the secret with the control plane client configuration")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
```
david-martin, line 61: why the change of port?
Won't this conflict with the controller if running that locally too?

cmd/agent/main.go
```diff
@@ -58,9 +58,9 @@ func main() {
...
+	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	flag.StringVar(&controlPlaneConfigSecretName, "control-plane-config-name", "control-plane-cluster", "The name of the secret with the control plane client configuration")
```
david-martin, line 63: Why is this changing for the agent?
I wasn't expecting any agent changes in this PR

hack/.argocdUtils
```diff
@@ -9,3 +9,48 @@ argocdAddCluster() {
     setLabel argocd.argoproj.io/secret-type cluster | 
     kubectl apply --context kind-${hubCluster} -f -
 }
+
+
+CreateAgentSecret() {
```
david-martin, line 14: This isn't used anywhere that I can see.
Also, it doesn't seem to be argocd related

cmd/syncer/main.go
```diff
@@ -0,0 +1,196 @@
...
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/sirupsen/logrus"
```
david-martin, line 23: why logrus?
`sigs.k8s.io/controller-runtime/pkg/log` is used elsewhere

david-martin (changes requested):
Code wise, the syncing implementation looks reasonable (it's inspired by kcp right?)
I've left a few comments unrelated to the syncing piece.

1 other thing, a slight concern about local dev.
Ideally a developer can run each component locally to develop either the control plane piece, the agent or syncer in the data plane, or test some things out with argocd.
I've highlighted a port conflict thing above, but there is another thing.
When you export `LOCAL_ACCESS="true"` and run `local-setup`, the argocd cluster secrets are affected where the server urls use something like `https://127.0.0.1:63441` instead of the internal kind hostname e.g. mctc-workload-1
Is there a way to run the syncer locally *and* have the argocd cluster registration still work for local dev?

cmd/syncer/main.go
david-martin: Looks like it's using go-lgr, so we could import that directly if you didn't want to use controller-runtime here?
```
All logging in controller-runtime is structured, using a set of interfaces defined by a package called logr (https://pkg.go.dev/github.com/go-logr/logr)
```

cmd/agent/main.go
philbrookes: yes, my thoughts were to set the values in the makefile and deployment.yaml, and leave the hardcoded values sane

cmd/agent/main.go
philbrookes: I was standardizing the parameter name to something more readable - as it's also used in the syncer agent.

hack/.argocdUtils
philbrookes: oh yeah, this an artifact from rebasing with Sergio's changes. I'll remove it.

cmd/syncer/main.go
philbrookes: I didn't think too hard about this, honestly. Happy to change it.

philbrookes:
> Code wise, the syncing implementation looks reasonable (it's inspired by kcp right?) I've left a few comments unrelated to the syncing piece.
> 
> 1 other thing, a slight concern about local dev. Ideally a developer can run each component locally to develop either the control plane piece, the agent or syncer in the data plane, or test some things out with argocd. I've highlighted a port conflict thing above, but there is another thing. When you export `LOCAL_ACCESS="true"` and run `local-setup`, the argocd cluster secrets are affected where the server urls use something like `https://127.0.0.1:63441` instead of the internal kind hostname e.g. mctc-workload-1 Is there a way to run the syncer locally _and_ have the argocd cluster registration still work for local dev?

Yes, @sergioifg94 @mikenairn and I did discuss this. We really  want to improve this whole area in the future and just settled on something that at least improves the local dev experience for now. I think a short term solution for this is to create both secrets, and point the deployment.yaml at the internal secret and the make run-syncer target at the external secret.

philbrookes:
> `removedintendedToBeRemovedFromLocationfalsestillOwnedByExternalActorForLocationfalse`

I've found the cause of this, will push a fix.

cmd/agent/main.go
david-martin: OK.
So as it stands, the metrics-bind-address default value is 8080 for the ingress agent and the controller, while the syncer is 8084.
The health-probe-bind-address default is 8081 for ingress, 8081 for controller and 8085 for syncer.
Should the defaults be changed for the ingress agent or controller so they don't clash,
or should the make target(s) for either overwrite those args?

david-martin:
> I think a short term solution for this is to create both secrets, and point the deployment.yaml at the internal secret and the make run-syncer target at the external secret.

@philbrookes If that means that the argocd cluster setup works out of the box,
and you can run the syncer locally either with the `run-syncer` target or as an image in the cluster, that sounds like a win-win.
What needs to change in the PR to ensure that?

philbrookes:
> > I think a short term solution for this is to create both secrets, and point the deployment.yaml at the internal secret and the make run-syncer target at the external secret.
> 
> @philbrookes If that means that the argocd cluster setup works out of the box, and you can run the syncer locally either with the `run-syncer` target or as an image in the cluster, that sounds like a win-win. What needs to change in the PR to ensure that?

My initial thoughts on this were to do it in a separate PR, but actually I think it's not a big change, so I'll do it here.

david-martin (approved):
Re-verified.
👍

PR #90, opened 2023-02-23
new installation method of istio

R-Lawton:
1. Download istio CLI
`make istioctl -f hack/make/dependencies.make`
2. Bring up local setup
3. When everything up switch context to workload 1
`kubectl config use-context kind-mctc-workload-1`
4. Create gateway
`kubectl --context kind-mctc-workload-1 apply -n istio-system -f ./config/samples/gateway_istio.yaml`
5. Apply kuadrant cr
`kubectl apply -f config/samples/kuadrant.yaml`
6. Clone https://github.com/R-Lawton/mctc_phase2-demo
`git clone https://github.com/R-Lawton/mctc_phase2-demo`
7. Deploy applicationset which contains echo deployment and service and rate limit policy
`kubectl --context kind-mctc-control-plane apply -n argocd -f argo`
8. Curl echo application 
`curl  -v -H 'Host: test.example.com' http://localhost:9080/demo`
you should get two status 200 when making 2 request in 10 seconds if you create more you should get message about hitting limit.
9. If you check the pod running in the mctc-demo namespace in the workload cluster you should see the logs messages for each get request.


david-martin (changes requested):
Got the example app working e2e with rate limiting.
A few things.

Though the steps in the PR description say to run `make istioctl -f hack/make/dependencies.make`, it's only necessary to run `make istioctl` as the dependencies.make file is includes in the main make file. No code changes needed for this.

Can you add istioctl as a dependency of the `local-setup` target, given that it's required now (in the same manner as kind, kustomize and helm)?

The curl example didn't work for me. I'm not sure where the port 9080 came from.
However, I was able to hit the route (via the gateway) on the metallb service loadbalancer IP  for istio *and* see the rate limiting in action. So, no code changes needed here, just an observance.
```
while true;do curl -s -o /dev/null -w "%{http_code}\n" -H 'Host: test.example.com' http://172.32.200.0/demo && sleep 3; done
200
200
429
429
200
200
429
```

R-Lawton:
Sorry my mistake i forgot a port forwarding step but yet using the gateway IP works the same to

david-martin (approved):

PR #93, opened 2023-02-28
#44: Cluster registration

### LINKED ISSUE #44: Cluster Registration

sergioifg94:
## Description

> Closes #44 

Cluster registration script

## Verification steps

1. Run local-setup with only the control cluster
	```sh
	make local-setup
	```

1. Create the tenant namespace in the control cluster
	```sh
	kubectl create namespace my-tenant
	```

1. Create the tenant cluster
	```sh
	# Import the util scripts
	source hack/.setupEnv
	source hack/.kindUtils

	kindCreateCluster tenant-cluster 9091 8446
	```

1. Install Gateway API in the tenant cluster
	```sh
	export KUBECONFIG=$(pwd)/tmp/kubeconfigs/tenant-cluster.kubeconfig
	./bin/kustomize build config/gateway-api | kubectl apply -f -
	```

1. Run the register cluster script
	```sh
	export KUBECONFIG=$(pwd)/tmp/kubeconfigs/mctc-control-plane.kubeconfig
	kubectl config use-context kind-mctc-control-plane
	./hack/register-cluster.sh $(pwd)/tmp/kubeconfigs/internal/mctc-control-plane.kubeconfig $(pwd)/tmp/kubeconfigs/tenant-cluster.kubeconfig my-tenant kind-tenant-cluster > agent-manifests.yaml
	```

3. Deploy the syncer in the tenant cluster
	```sh
	# Build and load the syncer image
	make docker-build-syncer
	kind load docker-image syncer:latest --name tenant-cluster  --nodes tenant-cluster-control-plane

	export KUBECONFIG=$(pwd)/tmp/kubeconfigs/tenant-cluster.kubeconfig	

	kubectl create namespace mctc-system
	kubectl apply -f agent-manifests.yaml
	```

### DIFF
--- a/config/syncer-control/role.yaml
+++ b/config/syncer-control/role.yaml
@@ -1,0 +1,25 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  creationTimestamp: null
+  name: syncer
+  namespace: system
+rules:
+- apiGroups:
+  - gateway.networking.k8s.io
+  resources:
+  - gateways
+  verbs:
+  - get
+  - update
+  - watch
+  - list
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
+  - list
+  - watch
--- a/hack/.clusterUtils
+++ b/hack/.clusterUtils
@@ -1,22 +1,15 @@
 # shellcheck shell=bash
 
-makeSecretForCluster() {
-  local clusterName=$1
-  local targetCluster=$2
+makeSecretForKubeconfig() {
+  local kubeconfig=$1
+  local clusterName=$2
+  local targetCluster=$3
   local targetClusterName=${targetCluster:=$clusterName}
-  local localAccess=${3}
 
-  if [ "$localAccess" != "true" ]; then
-    internalFlag="--internal"
-  fi
-
-  local tmpfile=$(mktemp /tmp/kubeconfig-internal.XXXXXX)
-  ${KIND_BIN} export kubeconfig -q $internalFlag --name ${clusterName} --kubeconfig ${tmpfile}
-  local server=$(kubectl --kubeconfig ${tmpfile} config view -o jsonpath="{$.clusters[?(@.name == 'kind-${clusterName}')].cluster.server}")
-  local caData=$(kubectl --kubeconfig ${tmpfile} config view --raw -o jsonpath="{$.clusters[?(@.name == 'kind-${clusterName}')].cluster.certificate-authority-data}")
-  local certData=$(kubectl --kubeconfig ${tmpfile} config view --raw -o jsonpath="{$.users[?(@.name == 'kind-${clusterName}')].user.client-certificate-data}")
-  local keyData=$(kubectl --kubeconfig ${tmpfile} config view --raw -o jsonpath="{$.users[?(@.name == 'kind-${clusterName}')].user.client-key-data}")
-  rm -f ${tmpfile}
+  local server=$(kubectl --kubeconfig ${kubeconfig} config view -o jsonpath="{$.clusters[?(@.name == '${clusterName}')].cluster.server}")
+  local caData=$(kubectl --kubeconfig ${kubeconfig} config view --raw -o jsonpath="{$.clusters[?(@.name == '${clusterName}')].cluster.certificate-authority-data}")
+  local certData=$(kubectl --kubeconfig ${kubeconfig} config view --raw -o jsonpath="{$.users[?(@.name == '${clusterName}')].user.client-certificate-data}")
+  local keyData=$(kubectl --kubeconfig ${kubeconfig} config view --raw -o jsonpath="{$.users[?(@.name == '${clusterName}')].user.client-key-data}")
 
   cat <<EOF
 kind: Secret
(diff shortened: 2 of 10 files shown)


hack/.clusterUtils
```diff
@@ -41,6 +41,45 @@ type: Opaque
...
+          "keyData": "${keyData}"
+        }
+    }
+  name: ${targetClusterName}
+  server: ${server}
+type: Opaque
+EOF
+
+}
+
```
mikenairn, line 82: This looks almost identical to `makeSecretForCluster`, can one not call the other?

sergioifg94: Yes, I have it in mind to refactor it so `makeSecretForCluster` calls `makeSecretForKubeconfig`

config/syncer-control/role.yaml
```diff
@@ -0,0 +1,25 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
```
maleck13, line 3: is this the role applied in the control plane or the data plane?

maleck13: ah I see it seems to be the control plane one. Looks good. shame we can't limit it to just annotation updates

maleck13:
trying this out locally. Changes look reasonable

config/syncer-control/role.yaml
sergioifg94: Yes I checked if there was something similar to `/status` for annotations but didn't find anything

maleck13:
I think we may be able to limit the syncer to a particular namespace on the data plane cluster. This will simplify thing quite a bit, as a tenant cp ns maps to a data plane ns on a set of clusters. 
I mention it as we have a cluster role and cluster role binding currently. Don't need to change it as of now, but if we do move to a 1:1 then we should def change it

maleck13:
@sergioifg94 looks like might need to update the flags to the syncer
```
flag provided but not defined: -control-plane-namespace
```


```
Usage of /syncer:
  -control-plane-config-name string
    	The name of the secret with the control plane client configuration (default "control-plane-cluster")
  -control-plane-config-namespace string
    	The namespace containing the secret with the control plane client configuration (default "mctc-system")
  -health-probe-bind-address string
```

sergioifg94:
@maleck13 Oh yes for now I'm just removing the flag from the generated manifest, but when #96 is merged it should work

maleck13:
ok so is this blocked by that PR? wondering how to check the registration worked?

maleck13:
everything seemed to work as expected, but just the start up of the syncer failed.

maleck13:
Just a note. When I ran `make local-setup` a second time I got the following error

```
make local-setup
test -s /Users/maleck13/projects/src/github.com/kuadrant/multi-cluster-traffic-controller/bin/yq || GOBIN=/Users/maleck13/projects/src/github.com/kuadrant/multi-cluster-traffic-controller/bin go install github.com/mikefarah/yq/v4@v4.30.8
./hack/local-setup.sh
Deleting mctc network
Error response from daemon: error while removing network: network mctc id 3e25dc23e481a5e7b8ba2b1f9a7a71169a4b31510b99b03fe73c07f602ce48bb has active endpoints
Error response from daemon: network with name mctc already exists
```

philbrookes:
> looks like there's an error in the steps to setup:
> 
> ```
> pbrookes@pbrookes-mac multi-cluster-traffic-controller % source hack/.setupEnv
> source hack/.kindUtils
> 
> kindCreateCluster tenant-cluster 9091 8446
> kindCreateCluster:5: no such file or directory: /Users/pbrookes/Projects/src/github.com/Kuadrant/multi-cluster-traffic-controller/../bin/kind
> kindCreateCluster:26: no such file or directory: /Users/pbrookes/Projects/src/github.com/Kuadrant/multi-cluster-traffic-controller/../bin/kind
> kindCreateCluster:27: no such file or directory: /Users/pbrookes/Projects/src/github.com/Kuadrant/multi-cluster-traffic-controller/../bin/kind
> ```

I've manually worked around it for now. Not sure if this needs correcting though?

sergioifg94:
@philbrookes Are you using bash? I have zsh and I get similar errors so I need to explicitly open a bash terminal for it

philbrookes:
> 

oh, yes I am using zsh, I'll try switching to bash. Can we not make this work for zsh? Considering that a few of us are using Macs?

sergioifg94:
I think all the scripts are meant to be run in bash. See the `#!/bin/bash` at the first line. So when executing the script even if you're on zsh it will run it on bash and that's how it's always worked fine. The only thing is that if we `source` the script from zsh as we're doing in the verification steps it will fail because some stuff is not compatible.

I think that as this is only for the sake of verifying the PR and not a common step it's fine to leave it the way it is

philbrookes (approved):
As we discussed, it'd be good to update local-setup to use this cluster registration for the workload clusters eventually, so we can spot issues with either that process, or the tenant permissions earlier in development, but that can wait for another PR.

PR #96, opened 2023-03-01
sync status back to control plane

philbrookes:
Adds a spec syncer which will sync spec from upstream to downstream.
This also adds an argument to specify which namespace in the control plane to sync spec from and status to.
The code is not ideal, and I will be submitting a PR to improve it, I just want to get a working POC in, so that MS2 is not blocked.

##Verification
`make kind-load-syncer`
`export KUBECONFIG=./tmp/kubeconfigs/mctc-workload-1.kubeconfig`
`make deploy-syncer`
`export KUBECONFIG=./tmp/kubeconfigs/mctc-control-plane.kubeconfig`
`kubectl create namespace mctc-tenant`
`kubectl create -f config/samples/gateway.yaml -n mctc-tenant`
`kubectl annotate gateway example-gateway mctc-sync-agent/all=true -n mctc-tenant`
`kubectl get gateway example-gateway -n mctc-tenant -o yaml`
See that the spec is added as an annotation to the gateway object.

### DIFF
--- a/cmd/syncer/main.go
+++ b/cmd/syncer/main.go
@@ -42,6 +42,7 @@ import (
 	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/_internal/clusterSecret"
 	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/syncer"
 	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/syncer/spec"
+	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/syncer/status"
 )
 
 var (
@@ -51,8 +52,8 @@ var (
 )
 
 const (
-	DEFAULT_RESYNC = 0
-	SYNCER_THREADS = 1
+	DEFAULT_RESYNC     = 0
+	dataPlaneNamespace = "mctc-downstream"
 )
 
 type arrayFlags []string
@@ -78,12 +79,14 @@ func main() {
 	var controlPlaneConfigSecretName string
 	var controlPlaneConfigSecretNamespace string
 	var syncedResources arrayFlags
+	var controlPlaneNS string
 
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.StringVar(&controlPlaneConfigSecretName, "control-plane-config-name", "control-plane-cluster", "The name of the secret with the control plane client configuration")
 	flag.StringVar(&controlPlaneConfigSecretNamespace, "control-plane-config-namespace", "mctc-system", "The namespace containing the secret with the control plane client configuration")
 	flag.Var(&syncedResources, "synced-resources", "A list of GVRs to sync (e.g. ingresses.v1.networking.k8s.io)")
+	flag.StringVar(&controlPlaneNS, "control-plane-namespace", "mctc-tenant", "The name of the upstream namespace to sync resources from")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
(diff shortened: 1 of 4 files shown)


philbrookes:
ping @david-martin @R-Lawton @sergioifg94

david-martin:
First pass trying out the PR on its own looks promising.
I'm going to rebase #59 on this branch and see if I can get it integrated before giving the thumbs up.

1 thing I did spot already though is the status annotation name.
`mctc-status-syncer-status-kind-mctc-workload-1`
For Gateways I want to easily get the statuses of all gateways in all clusters it has been synced to.
The naming convention makes it a little trickier than I'd like i.e. I'll have to know which clusters the gateway was synced to, then construct the annotation names from them to pull out each status.
Alternatively I could loop through all annotations and look for matches on the annotation name prefix.
It's doable, but I'm wondering if it can be made easier? 
I don't recall how it was done for kcp, but if its the same pattern, that's ok, we can leave it as is (as I suspect they did it that was for a reason)

philbrookes:
> First pass trying out the PR on its own looks promising. I'm going to rebase #59 on this branch and see if I can get it integrated before giving the thumbs up.
> 
> 1 thing I did spot already though is the status annotation name. `mctc-status-syncer-status-kind-mctc-workload-1` For Gateways I want to easily get the statuses of all gateways in all clusters it has been synced to. The naming convention makes it a little trickier than I'd like i.e. I'll have to know which clusters the gateway was synced to, then construct the annotation names from them to pull out each status. Alternatively I could loop through all annotations and look for matches on the annotation name prefix. It's doable, but I'm wondering if it can be made easier? I don't recall how it was done for kcp, but if its the same pattern, that's ok, we can leave it as is (as I suspect they did it that was for a reason)

In KCP the labels (we use annotations) used to mark a resource as should-be-synced for a particular agent (i.e. `mctc-sync-agent/mctc-workload-1`) had a value which was set to a value by the syncer to tell the control plane how the syncing was progressing (e.g. "Synced"). As it happens we didn't use that, as I recall in GLBC we just iterated through all the annotations that had the status prefix.

david-martin:
> I'm going to rebase #59 on this branch and see if I can get it integrated before giving the thumbs up.

Done. This helped unearth a couple of issues with the gateway reconcile logic, which If added fixes for to #59.
Approving based on functionality

david-martin (approved):

cmd/syncer/main.go
```diff
@@ -156,43 +170,93 @@ func startSyncers(ctx context.Context, GVRs []string, secretRef client.ObjectKey
...
 	specSyncConfig := syncer.Config{
-		UpstreamClientConfig: upstreamRestConfig,
-		GVRs:                 GVRs,
-		DownStreamClient:     mgr.GetClient(),
-		InformerFactory:      informerFactory,
-		ClusterID:            upstreamClusterConfig.Name,
-		NeverSyncedGVRs:      NEVER_SYNCED_GVRs,
+		GVRs:            GVRs,
+		InformerFactory: informerFactory,
+		ClusterID:       upstreamClusterConfig.Name,
```
maleck13, line 185: the cluster id where does that come from?

cmd/syncer/main.go
```diff
@@ -156,43 +170,93 @@ func startSyncers(ctx context.Context, GVRs []string, secretRef client.ObjectKey
...
+		NeverSyncedGVRs: NEVER_SYNCED_GVRs,
+		UpstreamNS:      controlPlaneNS,
+		DownstreamNS:    dataPlaneNamespace,
```
maleck13, line 188: so I think in the next milesone this could just use the downward API

cmd/syncer/main.go
philbrookes: it's the name data field in the cluster config secret.

cmd/syncer/main.go
```diff
@@ -156,43 +170,93 @@ func startSyncers(ctx context.Context, GVRs []string, secretRef client.ObjectKey
...
+	if err != nil {
+		setupLog.Error(err, "unable to create sync secret client")
+		os.Exit(1)
+	}
+	err = secretClient.Get(ctx, secretRef, controlPlaneSecret, &client.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("unable to retrieve control plane config secret '%v', error: %v", secretRef.String(), err)
+
+	}
+	downstreamClusterConfig, err := clusterSecret.ClusterConfigFromSecret(controlPlaneSecret)
```
maleck13, line 221: I am not following this, downstream is the cluster the sync agent is installed in?

maleck13: this ``` downstreamClient, err := clusterSecret.DynamicClientsetFromSecret(controlPlaneSecret) ``` reads strangely

cmd/syncer/main.go
```diff
@@ -156,43 +170,93 @@ func startSyncers(ctx context.Context, GVRs []string, secretRef client.ObjectKey
...
+	}
+	upstreamClient, err := dynamic.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		return nil, fmt.Errorf("unable to create client from mgr rest config: %v", err)
+	}
+	informerFactory := dynamicinformer.NewDynamicSharedInformerFactory(upstreamClient, DEFAULT_RESYNC)
+	informerFactory.Start(ctx.Done())
+	informerFactory.WaitForCacheSync(ctx.Done())
+
+	//for status: downstream is control-plane, upstream is data plane
```
maleck13, line 237: ahh. This is very confusing, is not simpler to think of CP as always being upstream and DP always being downstream?

pkg/syncer/spec/controller.go
```diff
@@ -26,8 +28,7 @@ const (
 	SyncerDeletionAnnotationPrefix      = "mctc-spec-syncer-deletion-timestamp-"
 	SyncerClusterStatusAnnotationPrefix = "mctc-spec-syncer-status-"
 	syncerApplyManager                  = "syncer"
-	downstreamNamespace                 = "mctc-downstream"
-	NUM_THREADS                         = 1
+	NUM_THREADS                         = 8
```
maleck13, line 31: any particular reason for this number?

cmd/syncer/main.go
philbrookes: Downstream is where the sync is sync to. Upstream is where it is syncing from.

For Spec: upstream = control plane, downstream = data plane
for Status: upstream = data plane, downstream = control plane

cmd/syncer/main.go
philbrookes: I intentionally avoided that, to keep the logic in the syncer from being so hardwired to our solution. Instead keeping the concept of upstream / downstream == from / to.

pkg/syncer/spec/controller.go
philbrookes: no, I just put in something.

cmd/syncer/main.go
philbrookes: I'm happy to discuss that further with you, to make sure we land on something that is both useful for us, and for other projects. Either way it can be discussed outside the scope of this PR though - I think?

PR #102, opened 2023-03-07
WIP Generate kustomize configs per cluster during local setup

mikenairn:
Update the local-setup logic to generate a set of kustomize configs for each kind cluster (control, or workload) being created during local setup. These configs use one of two different overlays (control-cluster or workload-cluster) depending on what type of cluster is being created, and this overlay is responsible for maintaining the set of resources that need to be added to that type of cluster.  Kustomize now manages the creation of each clusters secret used by both argocd and the mctc controller/agent/syncer itself. During generation of the clusters config a `cluster-secret` kustomization is generated containing config extracted from the running clusters kubeconfig. Note the `server` for these secrets now use the internal node ips instead for the mctc network, so there is no need for 2 secrets (internal/external).

In some cases during initial installation it is necessary to install some resources or services before adding others i.e. nginx, base/istio etc.. In order to allow this each overlay has an init kustomization that contains a list of resources that must be installed ahead of everything else. After initial creation the default clusters kustomization can be used to reconcile all changes as normal.

Generate a config per cluster IMO has a few advantages over the current approach of just directly applying individual kustomizations directly:

* It's possible to see exactly the resources that were applied to each cluster as you can run the kustomize build any time after setup: 
* You can reapply all the resource, modify them in the cluster config or delete them all as required.
* Secrets are now generated as part of the kustomize config and can be regenerted/viewed at any time:
* Works more like how argo will eventually do things.

Verification:
```
$ make local-cleanup && make local-setup MCTC_WORKLOAD_CLUSTERS_COUNT=1
$ tree config/local-setup/clusters
config/local-setup/clusters
├── mctc-control-plane
│   ├── argocd
│   │   └── kustomization.yaml
│   ├── cluster-secret
│   │   ├── cluster-config.env
│   │   ├── config
│   │   └── kustomization.yaml
│   ├── init
│   │   └── kustomization.yaml
│   ├── kustomization.yaml
│   └── mctc-control-plane.kubeconfig
└── mctc-workload-1
    ├── cluster-secret
    │   ├── cluster-config.env
    │   ├── config
    │   └── kustomization.yaml
    ├── init
    │   └── kustomization.yaml
    ├── kustomization.yaml
    └── mctc-workload-1.kubeconfig

7 directories, 13 files
```

Build specific cluster resources:
```
./bin/kustomize --load-restrictor LoadRestrictionsNone build config/local-setup/clusters/mctc-control-plane/ --enable-helm --helm-command ./bin/helm
```

Build specific cluster secret only:
```
./bin/kustomize --load-restrictor LoadRestrictionsNone build config/local-setup/clusters/mctc-control-plane/cluster-secret/ --enable-helm --helm-command ./bin/helm
```

Test argocd connection:

```
kubectl config use-context kind-mctc-control-plane
./bin/kustomize build config/local-setup/control-cluster/argocd/applications/ | kubectl apply -f -
applicationset.argoproj.io/echo-app created
```

Check in the UI stuff has happened (https://argocd.172.32.0.2.nip.io),. Note, it's expected that the echo app might fail to sync on the workload cluster if you don't have the controller running locally to server the webhook.


philbrookes:
I'm going to close this for now, as it's been quite inactive and needs a rebase.

PR #120, opened 2023-03-13
Improve MZ and record lookup for gateway reconciles

### LINKED ISSUE #111: Update GetDNSRecords log to query MZ ns only
The GetDNSRecords dns service method gets called to return existing records for a particular host. This needs to check the namespace of the managed zone for that particular host, instead of what it's currently doing of just checking the default ctrl ns, and with later changes the default and the traffic resource ns (see https://github.com/Kuadrant/multi-cluster-traffic-controller/pull/84).

GetDNSRecords is called from 3 places:

 * EnsureManagedHost (ingress specific)
 * AddEndPoints
 * RemoveEndpoints

mikenairn:
closes #111 

Updates the gateway reconciler to build up a list of managed hosts which includes a reference to the managed zone which can service it, and the DNSRecord that will exist or just have been created for it.  Later in the reconciliation when adding endpoints, we can now explicitly tell it what DNSRecord to add the endpoints to for each managed host.

The reconcilation logic between ingress and gateways is currently quite different. Ingresses expects a single host to be generated for a suitable manged zone and gateways do a lookup of multiple associated hosts for a managed zone capable of serving each. To make things easier this PR separates the logic for the two inside the dns service since trying to make it work for both is counterintuitive as we will more than likley remove the current ingress logic eventually.

### DIFF
--- a/pkg/dns/service.go
+++ b/pkg/dns/service.go
@@ -68,6 +68,9 @@ func (s *Service) resolveIPS(ctx context.Context, t traffic.Interface) ([]string
 	return activeDNSTargetIPs, nil
 }
 
+// GetDNSRecords returns all the existing DNSRecords for all the hosts in the traffic resource.
+
+// Ingress only - This is only used for traffic resources that currently expect a managed host to be generated.
 func (s *Service) GetDNSRecords(ctx context.Context, traffic traffic.Interface) ([]*v1alpha1.DNSRecord, error) {
 	// TODO improve this to use a label and list instead of gets
 	hosts := traffic.GetHosts()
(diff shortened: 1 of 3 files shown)


pkg/dns/service.go
```diff
@@ -236,14 +224,44 @@ func (s *Service) EnsureManagedHost(ctx context.Context, t traffic.Interface) ([
...
 
-func (s *Service) CreateDNSRecord(ctx context.Context, hostKey string, managedZone *v1alpha1.ManagedZone) (*v1alpha1.DNSRecord, error) {
-	managedHost := strings.ToLower(fmt.Sprintf("%s.%s", hostKey, managedZone.Spec.DomainName))
+// GetManagedZone returns the most suitable ManagedZone to publish the given traffic resource into.
+//
+// Currently, this just returns the first ManagedZone found in the traffic resources own namespace, or if none is found,
+// it looks for the first one listed in the default ctrl namespace.
+//
+// Ingress only - This is only used for traffic resources that currently expect a managed host to be generated.
+func (s *Service) GetManagedZone(ctx context.Context, t traffic.Interface) (*v1alpha1.ManagedZone, error) {
```
maleck13, line 233: maybe we want to rename to something like GetDefaultManagedZone?

maleck13 (approved):
changes look ok. I think once we have a gateway type that support ingresses we will be able to have the same model for both

PR #121, opened 2023-03-14
Redis

R-Lawton:
1. Bring up local setup
2. When everything up switch context to workload 1
`kubectl config use-context kind-mctc-workload-1`
3. Create gateway
`kubectl --context kind-mctc-workload-1 apply -n istio-system -f ./config/samples/gateway_istio.yaml`
4. Apply kuadrant cr
`kubectl apply -f config/samples/kuadrant.yaml`
5. Clone https://github.com/R-Lawton/mctc_phase2-demo
`git clone https://github.com/R-Lawton/mctc_phase2-demo`
6. Deploy applicationset which contains echo deployment and service and rate limit policy
`kubectl --context kind-mctc-control-plane apply -n argocd -f argo`
7. Curl echo application
`while true;do curl -s -o /dev/null -w "%{http_code}\n" -H 'Host: test.example.com' http://172.32.200.0/demo && sleep 1; done` you should see 2 status 200 and the rest non 200
8. Exec into redis pod terminal and check redis for keys
```
kubectl --context kind-mctc-control-plane exec -it $(kubectl --context kind-mctc-control-plane get pods --selector app=redis -o name -n redis) -n redis -- /bin/sh

redis-cli

keys *
```
you should see something like the following:

Screenshot 2023-03-13 at 16 57 09

### DIFF
--- a/hack/local-setup.sh
+++ b/hack/local-setup.sh
@@ -34,6 +34,9 @@ EXTERNAL_DNS_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/external-dns
 ARGOCD_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/argocd
 ISTIO_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/istio/istio-operator.yaml
 GATEWAY_API_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/gateway-api
+REDIS_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/kuadrant/redis
+LIMITADOR_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/kuadrant/limitador
+
 WEBHOOK_PATH=${LOCAL_SETUP_DIR}/../config/webhook-setup/workload
 TLS_CERT_PATH=${LOCAL_SETUP_DIR}/../config/webhook-setup/control/tls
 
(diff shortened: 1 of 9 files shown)


config/limitador/redis-server.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: redisconfig
+  namespace: kuadrant-system
+stringData:
+  URL: redis://172.32.0.2:30611
+type: Opaque
```
mikenairn, line 8: Consider using the kustomize secret generator (look for `secretGenerator`) for secrets and try and avoid using hard coded values specific to local setup in the base kustomizations.

config/redis/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+resources:
```
mikenairn, line 1: add `namespace: redis` so it can be overridden if needs be.

hack/make/dependencies.make
```diff
@@ -65,5 +64,4 @@ $(ISTIOCTL):
 	$(eval ISTIO_TMP := $(shell mktemp -d))
 	cd $(ISTIO_TMP); curl -sSL https://istio.io/downloadIstio | ISTIO_VERSION=$(ISTIOVERSION) sh -
 	cp $(ISTIO_TMP)/istio-$(ISTIOVERSION)/bin/istioctl ${ISTIOCTL}
-	-rm -rf $(TMP)	
-
+	-rm -rf $(TMP)
```
mikenairn, line 67: Can probably undo this

config/limitador/redis-server.yaml
R-Lawton: ok yes your right

hack/make/dependencies.make
R-Lawton: it was removed before by accident i think, do you we not need it to remove the tmp files i don't think the get deleted automatically

philbrookes:
local verification has worked  for me, after a few bumps!

hack/local-setup.sh
```diff
@@ -154,6 +152,15 @@ deployKuadrant(){
 
   echo "Installing Kuadrant in ${clusterName}"
   ${KUSTOMIZE_BIN} build config/kuadrant | kubectl apply -f -
+
+  if [[ "${REDIS}" == true ]]; then
```
mikenairn, line 156: What is this set to by default? I assume we really just want this all the time for a mctc setup no? Is there ever a case when we wouldn't want to have it setup with redis, at least for the time being until there is some othwer better option for multi cluster support? Having extra options will just make it more likely environments will drift and not be tested properly, unless it's really needed I'd just have it do one or the other (redis or no redis.)

hack/make/dependencies.make
mikenairn: I see, i think there was an extra tab and newline here that you removed. My editor is setup to automatically add a newline to the end of any file i edit, it's usually good practise to do that, that's what GH is complaining about.

hack/local-setup.sh
R-Lawton: ok ill change to it being mandatory

mikenairn (approved):

philbrookes (approved):

PR #122, opened 2023-03-14
Fix DNSRecord lookup in gateway reconcile

mikenairn:
Ignore not found errors when checking for existing DNSRecord.

Related https://github.com/Kuadrant/multi-cluster-traffic-controller/pull/120


david-martin (approved):
Thanks for fixing this @mikenairn

david-martin (approved):
Thanks for fixing this @mikenairn

PR #123, opened 2023-03-20
Syncer agent unit tests

philbrookes:
- Add unit tests to the mutators
- Add envtests for status syncer and spec syncer separately
- Fix logic around deletion events in control plane, to propagate to data plane.
## Verification
`make test`


david-martin:
👀

pkg/syncer/mutator/jsonpatch.go
```diff
@@ -26,7 +26,7 @@ func (m *JSONPatch) GetName() string {
 func (m *JSONPatch) Mutate(cfg syncer.MutatorConfig, obj *unstructured.Unstructured) error {
 	patchString := metadata.GetAnnotation(obj, JSONPatchAnnotationPrefix+cfg.ClusterID)
 	if patchString == "" {
-		return fmt.Errorf("no patch found for sync target '%v' on object: %v/%v", cfg.ClusterID, obj.GetNamespace(), obj.GetName())
+		return nil
```
david-martin, line 29: What's the reason for removing the error returned here?

pkg/syncer/spec/suite_test.go
```diff
@@ -0,0 +1,345 @@
...
+				err := k8sClient.Get(ctx, types.NamespacedName{Name: secret.Name, Namespace: dataPlaneNS}, &v1.Secret{})
+				return apierrors.IsNotFound(err)
+			}, TestTimeoutMedium, TestRetryIntervalMedium).Should(BeTrue())
+
+		})
+
+		It("should apply correctly annotated json mutations to a gateway and secret", func() {
+			metadata.AddAnnotation(gateway, syncer.MCTC_SYNC_ANNOTATION_PREFIX+clusterID, "true")
+			metadata.AddAnnotation(gateway, mutator.JSONPatchAnnotationPrefix+clusterID, `[
+			  {"op": "replace", "path": "/spec/gatewayClassName", "value": "istio"},
```
david-martin, line 328: Given there is currently a hardcoded transform of the `gatewayClassName` to `istio` [here](https://github.com/Kuadrant/multicluster-gateway-controller/blob/c2357b58be65b3be33210d8d81ae67823ab59399/pkg/syncer/spec/controller.go#L266), would it be better to change the value to something else?

pkg/syncer/status/suite_test.go
```diff
@@ -0,0 +1,268 @@
...
+				}
+				//ctrl.Log.Info("got data plane gateway", "status", dataplaneGateway.Status)
+
+				err = k8sClient.Get(ctx, gatewayNamespacedName, controlPlaneGateway)
+				if err != nil {
+					ctrl.Log.Error(err, "error getting control plane gateway")
+					return false
+				}
+				//ctrl.Log.Info("got control plane gateway", "annotations", controlPlaneGateway.Annotations)
+				return metadata.HasAnnotation(controlPlaneGateway, SyncerClusterStatusAnnotationPrefix+clusterID)
```
david-martin, line 261: Would it be better to verify the annotation contents here as well?

david-martin (changes requested):
A few comments, nothing major.
This is a big improvement on catching regressions and laying the foundation for more testing.

pkg/syncer/mutator/jsonpatch.go
philbrookes: When writing the unit tests, it felt unnatural for the mutator to error just because no patches were specified, as it seems valid that something might need to be synced without being patched.

pkg/syncer/spec/suite_test.go
philbrookes: I actually changed it to Istio because of that hardcoded change. Once that hardcoding is removed then this could be changed. Though it should still pass then anyway.

pkg/syncer/status/suite_test.go
philbrookes: I considered that alright, but wasn't sure if it'd get a bit noisy. I'll add it.

pkg/syncer/mutator/jsonpatch.go
david-martin: That sounds reasonable.
What's the value if you want to just sync the resource to a cluster? An empty string `""` ?

pkg/syncer/status/suite_test.go
philbrookes: I've added this now, deepequals was being a little weird with me, but I've got something that works.

PR #124, opened 2023-03-20
#118: Ability to handle parameters in GatewayClass

### LINKED ISSUE #118: GatewayClass parameters
Gateway Class from Gateway API supports a parameterRef. This can be any GVK. We want to leverage this to hold specific configuration that is tied to that gateway class.
Examples config:
- Default configuration such as number of replicas, and resource limits
- Which data plane gateway class and controller should instantiate the gateway

sergioifg94:
## Description 

> Closes #118 

Ability to get parameters from a resourced referenced in the `parametersRef` of the GatewayClass

## Verification steps

As the parameters are currently not used, in order to verify this functionality, ensure that the adequate status condition is reconciled when the `parametersRef` is misconfigured.

1. Start the local development clusters
    ```sh
    ```

3. Run the controller
    ```sh
    ```

1. Create the parameters ConfigMap
    ```sh
    kubectl create configmap gateway-params --from-file=params=config/samples/gatewayclass_params.json -n multi-cluster-traffic-controller-system
    ```

5. Create the GatewayClass
    ```yaml
    apiVersion: gateway.networking.k8s.io/v1beta1
    kind: GatewayClass
    metadata:
      name: mctc-gw-istio-external-instance-per-cluster
    spec:
      controllerName: "kuadrant.io/mctc-gw-controller"
      parametersRef:
        group: ""
        kind: "ConfigMap"
        namespace: "multi-cluster-traffic-controller-system"
        name: "gateway-params"
    ```

7. Create the Gateway with `.spec.gatewayClassName = mctc-gw-istio-external-instance-per-cluster`
    ```sh
    kubectl apply -n mctc-tenant -f gateway.yaml
    ```

1. Verify that once reconciled, the annotation is set
    ```sh
    kubectl get gateway/example-gateway -n mctc-tenant
    ```
    ```yaml
    metadata:
      name: example-gateway
      annotations:
        mctc-syncer-patch/kind-mctc-workload-1: [{"op":"replace","path":"/spec/gatewayClassName","value":"istio"}] 
    ```

### DIFF
--- a/hack/local-setup.sh
+++ b/hack/local-setup.sh
@@ -286,3 +286,8 @@ fi
 
 #10. Ensure the current context points to the control plane cluster
 kubectl config use-context kind-${KIND_CLUSTER_CONTROL_PLANE}
+
+ # Create configmap with gateway parameters for clusters
+kubectl create configmap gateway-params \
+  --from-file=params=config/samples/gatewayclass_params.json \
+  -n multi-cluster-traffic-controller-system
--- a/pkg/controllers/gateway/gateway_controller.go
+++ b/pkg/controllers/gateway/gateway_controller.go
@@ -26,7 +26,9 @@ import (
 	corev1 "k8s.io/api/core/v1"
 
 	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/_internal/conditions"
+	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/_internal/metadata"
 	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/_internal/slice"
+	syncutils "github.com/Kuadrant/multi-cluster-traffic-controller/pkg/_internal/sync"
 	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/apis/v1alpha1"
 	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/syncer"
 	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/traffic"
@@ -118,9 +120,32 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, nil
 	}
 
+	params, err := getParams(ctx, r.Client, gatewayClass)
+
+	// If the GatewayClass parametrers are invalid, update the status
+	// and stop reconciling
+	if err != nil && IsInvalidParamsError(err) {
+		gateway := previous.DeepCopy()
+		conditions.SetCondition(gateway.Status.Conditions, previous.Generation, string(gatewayv1beta1.GatewayConditionProgrammed), metav1.ConditionFalse, string(gatewayv1beta1.GatewayReasonPending), fmt.Sprintf("Invalid parameters in gateway class: %s", err.Error()))
+
+		if !reflect.DeepEqual(gateway, previous) {
+			err = r.Status().Update(ctx, gateway)
+			if err != nil {
+				log.Error(err, "Error updating Gateway status")
+				return ctrl.Result{}, err
+			}
+		}
+
+		return ctrl.Result{}, nil
+	}
+
+	if err != nil && !IsInvalidParamsError(err) {
+		return ctrl.Result{}, err
+	}
+
 	gateway := previous.DeepCopy()
 	acceptedStatus := metav1.ConditionTrue
-	programmedStatus, clusters, requeue, reconcileErr := r.reconcileGateway(ctx, *previous, gateway)
+	programmedStatus, clusters, requeue, reconcileErr := r.reconcileGateway(ctx, *previous, gateway, params)
 
 	// Update gateway spec/metadata
 	if !reflect.DeepEqual(gateway, previous) {
@@ -150,7 +175,7 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 
 // Configures Gateway tls & dns for each cluster it targets.
 // Returns the programmed status, a list of clusters that were programmed, if the gateway should be requeued, and any error
-func (r *GatewayReconciler) reconcileGateway(ctx context.Context, previous gatewayv1beta1.Gateway, gateway *gatewayv1beta1.Gateway) (metav1.ConditionStatus, []string, bool, error) {
+func (r *GatewayReconciler) reconcileGateway(ctx context.Context, previous gatewayv1beta1.Gateway, gateway *gatewayv1beta1.Gateway, params *Params) (metav1.ConditionStatus, []string, bool, error) {
 	log := log.FromContext(ctx)
 
 	clusters := selectClusters(*gateway)
@@ -163,6 +188,10 @@ func (r *GatewayReconciler) reconcileGateway(ctx context.Context, previous gatew
 	gatewayHelper := trafficAccessor.(GatewayHelper)
 	allHosts := trafficAccessor.GetHosts()
 	managedHosts := []ManagedHost{}
+
+	if err := r.reconcileParams(ctx, gateway, params); err != nil {
+		return metav1.ConditionUnknown, clusters, false, err
+	}
 
 	// Create a list of managed hosts.
 	// Find a suitable managed zone for each host, if one exists, ensure a DNSRecord exists for it.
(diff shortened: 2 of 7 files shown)


makslion:
Verified GatewayClass gets `InvalidParameters` field 
image

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -117,9 +117,32 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, nil
 	}
 
+	// TODO: Use return value
```
makslion, line 120: Aren't we using the return value at line 145? :thinking:

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -149,7 +172,7 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 
 // Configures Gateway tls & dns for each cluster it targets.
 // Returns the programmed status, a list of clusters that were programmed, if the gateway should be requeued, and any error
-func (r *GatewayReconciler) reconcileGateway(ctx context.Context, previous gatewayv1beta1.Gateway, gateway *gatewayv1beta1.Gateway) (metav1.ConditionStatus, []string, bool, error) {
+func (r *GatewayReconciler) reconcileGateway(ctx context.Context, previous gatewayv1beta1.Gateway, gateway *gatewayv1beta1.Gateway, params *Params) (metav1.ConditionStatus, []string, bool, error) {
```
makslion, line 175: _ *Params?

pkg/controllers/gateway/params.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func IsInvalidParamsError(err error) (is bool) {
+	_, is = err.(*InvalidParamsError)
+	return
+}
+
+type ParamsResolver func(context.Context, client.Client, gatewayv1beta1.ParametersReference) (*Params, error)
+
+var paramsResolvers map[schema.GroupKind]ParamsResolver = map[schema.GroupKind]ParamsResolver{}
```
makslion, line 37: the type here could be omitted as well, right?

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -117,9 +117,32 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 
+	// TODO: Use return value
+	params, err := getParams(ctx, r.Client, gatewayClass)
+	if err != nil && !IsInvalidParamsError(err) {
+		return ctrl.Result{}, nil
+	}
+
+	// If the GatewayClass parametrers are invalid, update the status
+	// and stop reconciling
+	if IsInvalidParamsError(err) {
```
maleck13, line 128: Nit: This might read better as 

``` 
if err !=nil && IsInvalidParamsError(err) {
...

}

if err != nil{
   return ctrl.Result{}, nil
}
```

Why do we ignore the err on line 122 should that be returned?

pkg/controllers/gateway/params.go
```diff
@@ -0,0 +1,89 @@
...
+	result := &Params{}
+	return result, nil
+}
+
+func init() {
+	configMapGroupKind := schema.GroupKind{
+		Group: corev1.GroupName,
+		Kind:  "ConfigMap",
+	}
+	paramsResolvers[configMapGroupKind] = fromConfigMap
```
maleck13, line 64: I wonder if we need this, I think we are only ever likely to have one params object type

pkg/controllers/gateway/params.go
```diff
@@ -0,0 +1,89 @@
...
+		Group: string(gatewayClass.Spec.ParametersRef.Group),
+		Kind:  string(gatewayClass.Spec.ParametersRef.Kind),
+	}
+
+	resolveParams, ok := paramsResolvers[groupKind]
+	if !ok {
+		return nil, &InvalidParamsError{fmt.Sprintf("unable to retrieve parameters for GroupKind %s", groupKind.String())}
+	}
+
+	params, err := resolveParams(ctx, client, *gatewayClass.Spec.ParametersRef)
```
maleck13, line 83: could this be simplified to

```
return resolveParams()
```

pkg/controllers/gateway/params.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+var _ error = &InvalidParamsError{}
+
+// Error implements error
+func (e *InvalidParamsError) Error() string {
+	return e.message
+}
+
+func IsInvalidParamsError(err error) (is bool) {
```
maleck13, line 30: I always wonder if the default for these should be

```
if err == nil{
return false
}
```

maleck13 (commented):
We have a valid use case for the parameters in adding in the supported data plane gateway controller to use (istio)
So we could actually use the code here and create a default configmap params with the gateway controller to use. This param is set as a transform on the gateway resource

maleck13:
couple of code comments and a suggested change to make the code do some actual work for us

pkg/controllers/gateway/gateway_controller.go
sergioifg94: I think that given that we intend to use that parameter in the future it makes sense to leave it named. I'd use the `_` when it's a parameter that we don't care about (we don't yet, but we will)

pkg/controllers/gateway/gateway_controller.go
sergioifg94: Makes sense, I'll make those changes. About the err being ignored, I missed it, it should indeed be returned

pkg/controllers/gateway/params.go
sergioifg94: The motivation for it is to be able to support multiple types in case we want to for example define a CRD for the parameters. So the user can have the option to define the parameters in a ConfigMap or in a CR and the controller will be able to handle it

pkg/controllers/gateway/params.go
sergioifg94: That's an interesting question, because it would be possible to define
```go
var err1 *InvalidParamsError = nil
var err2 error = nil
```

Both err1 and err2 would be `nil` but passing err1 to `IsInvalidParamsError` would return true, and passing err2 would return false.
I think it makes sense to leave it like this and leave the nil check responsibility to the caller of `IsInvalidParamsError`

sergioifg94:
Updated the PR with the suggestions, added some unit tests and updated the verification steps to verify that the patch annotation is set from the value specified in the parameters

maleck13 (approved):

maleck13:
@sergioifg94 just one question before merge. Prob worth having this params created as part of local set up so it is there and being used?

sergioifg94:
@maleck13 I added a commit to create the params in the local-setup and the `parametersRef` section filled by default in the sample gateway class

hack/local-setup.sh
```diff
@@ -224,6 +224,17 @@ initController() {
...
 
+# defaultGatewayParams adds a cluster to the config/samples/gatewayclass_params.json
+# file. This file can be used to create a ConfigMap referenced in the
+# parameterssRef section of the GatewayClass
+defaultGatewayParams() {
+  clusterName=${1}
+
+  echo "Adding ${clusterName} to gatewayclass params"
+
+  ${YQ_BIN} -i -P -o=json '.downstreamClasses["'$clusterName'"] = "istio"' config/samples/gatewayclass_params.json
```
maleck13, line 235: @sergioifg94 why are we keying it off the cluster name? The idea with the gatewayclass params is it is configuration to be applied to gateways of this class, so I would expect all gateways of this class would have the same downstream class

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -164,6 +188,15 @@ func (r *GatewayReconciler) reconcileGateway(ctx context.Context, previous gatew
 	allHosts := trafficAccessor.GetHosts()
 	managedHosts := []ManagedHost{}
 
+	// Set the annotations to sync the class name from the parameters
+	for cluster, downstreamClass := range params.GetDownstreamClasses() {
```
maleck13, line 192: yeah same here, I think for the downstream class, each instance would have the same class

maleck13 (commented):
I don't think the params should be keyed of the clustername, the idea with the params is it is additional config for a given gateway class and so generic

maleck13:
@mikenairn I see your PR depends on this one. We can probably merge this, I would just prefer the params weren't keyed against a cluster. The gatewayclass params, IMO are intended as class level configuration rather than cluster level configuration cc @sergioifg94

sergioifg94:
I was planning on waiting until patching annotations for all clusters is supported, but we can merge this as is for now and make those changes later

sergioifg94:
@maleck13 I just updated the PR to follow your suggestion, so the parameter is now `downstreamClass` and the patch is applied to all clusters

maleck13 (approved):

Issue #125, opened 2023-03-21
Use transform annotation for 'istio' gatewayClassName instead of hardcoded transform


david-martin:
Update the gateway controller to add a transform annotation, setting the `gatewayClassName` to `istio` when the Gateway is synced to workload clusters.
See https://github.com/Kuadrant/multicluster-gateway-controller/blob/c2357b58be65b3be33210d8d81ae67823ab59399/config/samples/gateway.yaml#L8-L11 for an example annotation patch for a specific workload cluster.
Remove the hardcode transform at https://github.com/Kuadrant/multicluster-gateway-controller/blob/c2357b58be65b3be33210d8d81ae67823ab59399/pkg/syncer/spec/controller.go#L266

PR #128, opened 2023-03-21
Gateway placement via Placement CR

david-martin:
Closes #52 

A new controller for placements has been added, along with an integration test.
The milestone 2 walkthrough has been moved to a subfolder, and a new milestone 3 folder & doc added as a starting point for m3 features.
To verify the changes, follow the steps in the milestone 3 walkthrough doc.

PR #132, opened 2023-03-22
GH-125 patch anotations for gatewayClassName

makslion:
## What
Change hardcoded className patch from syncer to a patch annotation applied by the gateway controller 

## Verification
Deploy gateway and ensure that downstream className gets updated to `istio`

### DIFF
--- a/pkg/controllers/gateway/gateway_controller.go
+++ b/pkg/controllers/gateway/gateway_controller.go
@@ -121,7 +121,7 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 
 	gateway := previous.DeepCopy()
 	acceptedStatus := metav1.ConditionTrue
-	programmedStatus, clusters, requeue, reconcileErr := r.reconcileGateway(ctx, *previous, gateway)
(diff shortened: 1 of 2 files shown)


pkg/controllers/gateway/gateway_controller.go
```diff
@@ -271,6 +272,12 @@ func (r *GatewayReconciler) reconcileGateway(ctx context.Context, previous gatew
 			}
 		}
 	}
+
+	//set jsonpatch annotation
+	metadata.AddAnnotation(gateway, "mctc-syncer-patch/kind-mctc-workload-1", "[\n"+
```
david-martin, line 277: We don't want this hardcoded to the `kind-mctc-workload-1` cluster.
Right now, the syncObjectToAllClusters func below uses a special annotation to sync the object to all clusters. 
I've reached out to Phil to see if there's an equivalent for the patching annotation.
If not, we can hold this PR, pending #128. The list of clusters can be fetched from the annotations added by the placement controller when that lands

david-martin (changes requested):
Holding for now

david-martin:
keeping on hold as the implementation will change depending on the outcome of #152 cc @maleck13

david-martin:
Closing as the underlying implementation of syncing has changed with #175

PR #136, opened 2023-03-24
Syncer fix

philbrookes:
two fixes:
- Default downstream namespace if none provided
- Status syncer no longer expects the syncer annotation on downstream resources

ping @mikenairn @david-martin


david-martin:
👀

mikenairn (approved):
Changes look fine to me.

david-martin (approved):
Verified status is now being synced to the upstream as expected

PR #141, opened 2023-03-29
#98: Add DNSPolicy API and bare controller

### LINKED ISSUE #98: Define API spec for DNSPolicy
**what**

Define an API that uses a target ref to target a gateway resource and specify a health check for the DNS Provider to implement in  order to decide whether a DNS record should be considered part of a healthy response.

This resource is how we express DNS policy. It is intended to be translated into the DNSRecord.

- expected status code
- number of re-tries
- path to endpoint
reference https://github.com/kcp-dev/kcp-glbc/blob/main/docs/dns/health-checks.md

sergioifg94:
## Description

> Closes #98 

Add CRD and empty controller for DNSPolicy resource, including spec for health checks.

## Verification steps

1. Run local setup
    ```sh
    make local-setup
    ```

1. Verify that the CRD is created in the control cluster
    ```sh
    kubectl get crd dnspolicies.kuadrant.io                                                                                                                                      
    NAME                      CREATED AT
    dnspolicies.kuadrant.io   2023-03-29T10:29:11Z
    ```

1. Start the controller
    ```sh
    (export $(cat ./controller-config.env | xargs) && export $(cat ./aws-credentials.env | xargs) && make build-controller install run-controller
    ```

1. Verify that the DNSPolicy controller is started
    ```sh
    2023-03-29T11:57:22+01:00       INFO    Starting EventSource    {"controller": "dnspolicy", "controllerGroup": "kuadrant.io", "controllerKind": "DNSPolicy", "source": "kind source: *v1alpha1.DNSPolicy"}                                                                                                                                                                
    2023-03-29T11:57:22+01:00       INFO    Starting Controller     {"controller": "dnspolicy", "controllerGroup": "kuadrant.io", "controllerKind": "DNSPolicy"} 
    ```


config/crd/bases/kuadrant.io_dnspolicies.yaml
```diff
@@ -0,0 +1,156 @@
...
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: DNSPolicySpec defines the desired state of DNSPolicy
+            properties:
+              dnsRecordRef:
```
maleck13, line 43: Two thoughts here:

1) I had thought we would be targeting a gateway with this API and the controller would then translate the check into DNSRecord changes
2) with the targeting, I think we should reuse the targetRef concept from policy attachment and gateway api 

```  
  targetRef: # This struct is defined as part of Gateway API
    group: 
    kind: 
    name: 
 ```
https://gateway-api.sigs.k8s.io/references/policy-attachment/

Comment (maleck13) [reply config/crd/bases/kuadrant.io_dnspolicies.yaml:43]: also I thik we only support targeting resources in the same ns

File: config/crd/bases/kuadrant.io_dnspolicies.yaml [hunk 2]
```diff
@@ -0,0 +1,156 @@
...
+                properties:
+                  name:
+                    type: string
+                  namespace:
+                    type: string
+                required:
+                - name
+                - namespace
+                type: object
+              healthCheck:
```
maleck13, line 53: good to add some comment here: # by default this health check will be applied to each unique DNS A Record for the listeners assigned to the target gateway

config/crd/bases/kuadrant.io_dnspolicies.yaml
sergioifg94: I see. I had misunderstood and thought we were moving away from directly referencing the gateway in the policy. So I guess the controller when reconciling gets the Gateway and then the associated DNSRecords from it?

maleck13: yes so each listener in a gateway will have a DNSRecord so controller can find that DNSRecord and apply the health check.

mikenairn: Should we try and keep this consistent with how the RLP does this and use `targeRef` with `group`, `kind`, `name` and as @maleck13 says it will be in teh same ns?

pkg/apis/v1alpha1/dnspolicy_types.go
```diff
@@ -0,0 +1,103 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	gatewayapiv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+)
+
+// DNSPolicySpec defines the desired state of DNSPolicy
+type DNSPolicySpec struct {
+
+	// +kubebuilder:validation:Required
+	// +required
+	TargetRef gatewayapiv1alpha2.PolicyTargetReference `json:"targetRef"`
```
sergioifg94, line 29: Updated to use `PolicyTargetReference` instead

pkg/apis/v1alpha1/dnspolicy_types.go
```diff
@@ -0,0 +1,99 @@
...
+//+kubebuilder:object:root=true
+
+// DNSPolicyList contains a list of DNSPolicy
+type DNSPolicyList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []DNSPolicy `json:"items"`
+}
+
+type ProviderHealthCheckSpec struct {
```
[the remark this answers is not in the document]
maleck13: Do we see that as something we would expose externally? I am thinking we would likely try to present a single API and take care of the messy translations internally?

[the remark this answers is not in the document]
sergioifg94: I'm thinking of the possibility of a different API, as in a separate set of health checks that don't translate to health checks in Route 53. I don't know exactly how that would look like but I thought having a separate section would be a good first step. If you think that's not going to happen I can simplify this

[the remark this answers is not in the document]
maleck13: I think for now and as this is a alpha api, I would be inclined to simply make the spec what is in ProviderHealthCheckSpec. We can make it more complicated in the future if the need shows itself. Lets just ensure we mark this an alpha api that way the expectation is it might change significantly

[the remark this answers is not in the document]
sergioifg94: Cool, I updated the PR with these changes

maleck13 (approved):

Issue #144, opened 2023-03-31
bug: Syncer errors when an upstream resource is deleting and has other finalizers


philbrookes:
## recreate:
- set up local setup
- deploy syncer
- create gateway to control plane
- add a custom finalizer
- see gateway in data plane
- delete gateway in control plane
- see errors in syncer

PR #145, opened 2023-04-03
no errors on delete, patch for status annotations

philbrookes:
closes: #144 and #146


R-Lawton:
errors in syncer logs are now gone

R-Lawton:
issue still happening with reconciler tho but fixed the errors we were seeing that we thought was causing the issue

R-Lawton (approved):

PR #147, opened 2023-04-04
Local setup updates

mikenairn:
Local setup changes

- Allow overriding metrics and health ports on the `make run-syncer` command so you can run more than one locally:

```
METRICS_PORT=8088 HEALTH_PORT=8089 make run-syncer
```

- Remove kind from names in cluster secrets and make the cluster name consistent accross the control and data plane.

```
kubectl get secret mctc-workload-1 -n argocd --context kind-mctc-control-plane -o jsonpath="{.data.name}" | base64 --decode && echo ""
mctc-workload-1
kubectl get secret control-plane-cluster -n mctc-system --context kind-mctc-workload-1 -o jsonpath="{.data.name}" | base64 --decode && echo ""
mctc-workload-1
```

- Add kuadrant CR to local-setup

### DIFF
--- a/hack/local-setup.sh
+++ b/hack/local-setup.sh
@@ -147,16 +147,23 @@ installGatewayAPI() {
 
 deployKuadrant(){
   clusterName=${1}
-  kubectl config use-context kind-${clusterName}
 
-  echo "Installing Kuadrant in ${clusterName}"
-  ${KUSTOMIZE_BIN} build config/kuadrant | kubectl apply -f -
   kubectl config use-context kind-mctc-control-plane
   echo "Installing Redis in kind-mctc-control-plane"
-  ${KUSTOMIZE_BIN} build ${REDIS_KUSTOMIZATION_DIR} | kubectl apply -f - 
+  ${KUSTOMIZE_BIN} build ${REDIS_KUSTOMIZATION_DIR} | kubectl apply -f -
+
   kubectl config use-context kind-${clusterName}
+  echo "Installing Kuadrant in ${clusterName}"
+  ${KUSTOMIZE_BIN} build config/kuadrant | kubectl apply -f -
   echo "Configuring Limitador in ${clusterName}"
   ${KUSTOMIZE_BIN} build ${LIMITADOR_KUSTOMIZATION_DIR} | kubectl apply -f -
+  echo "Waiting for Kuadrant deployments to be ready..."
+  kubectl -n kuadrant-system wait --timeout=300s --for=condition=Available deployments --all
+  echo "Adding Kuadrant CR in ${clusterName}"
+  ${KUSTOMIZE_BIN} build config/kuadrant/cr | kubectl apply -f -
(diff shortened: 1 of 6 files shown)


hack/local-setup.sh
```diff
@@ -147,16 +147,23 @@ installGatewayAPI() {
...
   echo "Configuring Limitador in ${clusterName}"
   ${KUSTOMIZE_BIN} build ${LIMITADOR_KUSTOMIZATION_DIR} | kubectl apply -f -
+  echo "Waiting for Kuadrant deployments to be ready..."
+  kubectl -n kuadrant-system wait --timeout=300s --for=condition=Available deployments --all
+  echo "Adding Kuadrant CR in ${clusterName}"
+  ${KUSTOMIZE_BIN} build config/kuadrant/cr | kubectl apply -f -
+  # This is adding the kuadrant CR before any gateways are created which is currently a known issue with kuadrant.
+  # After adding a gateway is will necessary to force a reconcile of the kuadrant CR, this can be done by running:
+  # kubectl annotate kuadrant/mctc -n kuadrant-system updateme=`date +%s` --overwrite
 }
```
mikenairn, line 167: @R-Lawton I added the creation of the kuadrant CR here. There is an issue with kuadrant in that it only reconciles the gateways when you create or update the kuadrant CR, which is the reason it works in the M2 walkthough as we do it in that order, but it shouldn't have to be done that way. Until there is a fix, we will just have to remember that when we create a gateway that we force a reconcile of the kuadrant cr on the workload clusters. @david-martin fyi

@maleck13 This got me thinking, should we be creating the kuadrant CR as part of HCG and syncing these out to the clusters? re. https://github.com/Kuadrant/architecture/pull/5 seems we could eventually have one per tenant, and would allow us in the short term to force an update of the kuadrant CR (update an annotation) every time a gateway is created in HCG.

hack/local-setup.sh
```diff
@@ -205,7 +212,7 @@ deployAgentSecret() {
 
   kubectl create namespace mctc-system || true
 
-  makeSecretForCluster $KIND_CLUSTER_CONTROL_PLANE $(kubectl config current-context) $localAccess |
+  makeSecretForCluster $KIND_CLUSTER_CONTROL_PLANE $clusterName $localAccess |
```
mikenairn, line 215: @philbrookes @sergioifg94 Could you check this please. Currently the secrets on the data and control plane differ in that they both have different cluster names in them. This removes the "kind-" prefix and makes it consistent in both scerets.

hack/local-setup.sh
david-martin: > Until there is a fix
Is there an upstream issue to track in the kuadrant operator?

philbrookes:
Hey Mike, sorry for the delay, I'll look at testing this today.

philbrookes:
Tested this locally and everything still looks good to me, approved.

david-martin (approved):
Approving based on review of changes.
I haven't tried out the changes, but I won't hold this up.

david-martin (approved):

PR #127, opened 2023-03-21
Aggregate data plane gateway statuses into control plane resource

david-martin:

PR #150, opened 2023-04-11
Status aggregation proposal

david-martin:
Based on offline conversation about #127 

That PR can be updated based on the outcome of this proposal

### DIFF
--- a/docs/status_aggregation.md
+++ b/docs/status_aggregation.md
@@ -1,0 +1,48 @@
+# Proposal: Aggregation of Status Conditions
+
+## Background
+
+[Status conditions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) are used to represent the current state of a resource and provide information about any problems or issues that might be affecting it. They are defined as an array of [Condition](https://pkg.go.dev/k8s.io/apimachinery@v0.26.3/pkg/apis/meta/v1#Condition) objects within the status section of a resource's YAML definition.
+
+## Problem Statement
+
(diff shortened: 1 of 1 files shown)


philbrookes:
This proposal is just to cover conditions, I assume, and not the other items in the status? Would you think then, that the status annotation would still represent the entire status object (for easier parsing back to an object), or would the status annotation not contain the conditions then?

I'm asking as I'm wondering if the duplication of the message is necessary - I see that their is value for human consumption, but as the information there would already be present in the annotation then we could remove the duplication into the status block where it could potentially be extremely long (e.g. with many clusters all reporting different errors)?

david-martin:
> This proposal is just to cover conditions, I assume

Yes, exactly. Other items tend to be resource specific, while this proposal is generic.



> the status annotation would still represent the entire status object

Yes, that sounds reasonable.



> if the duplication of the message is necessary

Good point.
It comes down to the user of the status block in the control plane (which may be a physical user, or some other controller) and what they want from it.
If they are relying on certain conditions being true/false/other, the proposal allows for that without the message field.

> value for human consumption
The value of messages is most useful when there's an error. 
If the message wasn't readily available in the status block, the user would need a process to pull it out of the status annotation. 
If they are using a CLI or console to view the object status in the control plane, the processed data can be transient rather than storing it on the resource in control plane a 2nd time in a different format.
I don't see any harm is leaving the message piece out initially and let the use case bubble up more naturally.

philbrookes (approved):
Cool, this sounds like a great way of compressing all of the status conditions into a much more readable format.

maleck13 (approved):

PR #162, opened 2023-04-14
Gh 117 Cert and DNRecord deletion

R-Lawton:
1. Setup local environment
https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/docs/milestone2_walkthrough.md 
1. Create a gateway
```
kubectl apply -n mctc-tenant -f ./gateway.yaml
```
2. Ensure DNS record and certs are created  
 ```
kubectl get dnsrecords -n multi-cluster-traffic-controller-system
kubectl get certs -n mctc-tenant 
kubectl get secrets -n mctc-tenant 
```
3. Delete the gateway 
 ```
kubectl delete gateway <gateway-name> -n mctc-tenant
```
4. Check that the DNS record and certs are removed

### DIFF
--- a/pkg/controllers/gateway/gateway_controller.go
+++ b/pkg/controllers/gateway/gateway_controller.go
@@ -19,11 +19,12 @@ package gateway
 import (
 	"context"
 	"fmt"
+	certman "github.com/jetstack/cert-manager/pkg/apis/certmanager/v1"
+	corev1 "k8s.io/api/core/v1"
 	"reflect"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"strings"
 	"time"
-
-	corev1 "k8s.io/api/core/v1"
 
 	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/_internal/conditions"
 	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/_internal/metadata"
@@ -47,6 +48,7 @@ import (
 const (
 	ClusterSyncerAnnotation               = "clustersync.kuadrant.io"
 	GatewayClusterLabelSelectorAnnotation = "kuadrant.io/gateway-cluster-label-selector"
+	GatewayFinalizer                      = "kuadrant.io/gateway"
 )
 
 type HostService interface {
@@ -103,7 +105,11 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		// TODO: Do we need to remove dns records and/or endpoints?
 		//       Will ownerRefs be sufficient
 		log.Info("Gateway is deleting", "gateway", previous.Name, "namespace", previous.Namespace)
-		return ctrl.Result{}, nil
+		// cleanup dns records
+		// cleunup certificates
+		err = r.cleanupCertificates(ctx, previous)
+		// delete gateway itself if it has no finalizers?
+		return ctrl.Result{}, err
 	}
 
 	// Check if the class name is one of ours
@@ -464,4 +470,61 @@ func (r *GatewayReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		})).
 		Owns(&kuadrantapi.RateLimitPolicy{}).
 		Complete(r)
+}
+
+func (r *GatewayReconciler) cleanupCertificates(ctx context.Context, gateway *gatewayv1beta1.Gateway) error {
+	var hostsToRemove []string
+	var allGateways gatewayv1beta1.GatewayList
+
+	// get names of hosts for traffic object being deleted
+	for _, listener := range gateway.Spec.Listeners {
+		if string(*listener.Hostname) != "*." {
+			hostsToRemove = append(hostsToRemove, string(*listener.Hostname))
+		}
+	}
+	// list all the traffic objects
+	err := r.Client.List(ctx, &allGateways, client.InNamespace(gateway.Namespace))
+	if err != nil {
+		return fmt.Errorf("error listing all Gateways: %s", err)
+	}
+	// for each traffic object check if host is being used
+	for _, candidateGateway := range allGateways.Items {
+		// ignore the gateway being deleted
+		if candidateGateway.Name != gateway.Name && candidateGateway.DeletionTimestamp == nil {
+			// remove host from "delete" list if it is being used
+			for _, listener := range candidateGateway.Spec.Listeners {
+				for i, v := range hostsToRemove {
+					if v == string(*listener.Hostname) {
+						hostsToRemove = append(hostsToRemove[:i], hostsToRemove[i+1:]...)
+					}
+				}
+			}
+		}
+	}
+	// if there are hosts left - only deleting gateway using them
+	for _, host := range hostsToRemove {
+		secret := &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      host,
+				Namespace: gateway.Namespace,
+			},
+		}
+		cert := &certman.Certificate{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      host,
+				Namespace: gateway.Namespace,
+			},
+		}
+		// only secret or cert cr as well?
+		err = r.Client.Delete(ctx, secret)
+		if err != nil && !k8serrors.IsNotFound(err) {
+			return fmt.Errorf("error deleting cert secret: %s", err)
+		}
+		err = r.Client.Delete(ctx, cert)
+		if err != nil && !k8serrors.IsNotFound(err) {
+			return fmt.Errorf("error deleting certificate: %s", err)
+		}
+	}
+	// assumptions: all traffic objects using the same host are under the same namespace
+	return nil
 }
--- a/pkg/controllers/gateway/gateway_controller.go
+++ b/pkg/controllers/gateway/gateway_controller.go
@@ -513,7 +513,7 @@ func (r *GatewayReconciler) cleanupCertificates(ctx context.Context, gateway *ga
 
 	// get names of hosts for traffic object being deleted
 	for _, listener := range gateway.Spec.Listeners {
-		if string(*listener.Hostname) != "*." {
+		if !strings.Contains(string(*listener.Hostname), "*.") {
 			hostsToRemove = append(hostsToRemove, string(*listener.Hostname))
 		}
 	}
--- a/pkg/controllers/gateway/gateway_controller.go
+++ b/pkg/controllers/gateway/gateway_controller.go
@@ -100,16 +100,34 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		}
 		return ctrl.Result{}, nil
 	}
+	if !controllerutil.ContainsFinalizer(previous, GatewayFinalizer) {
+		controllerutil.AddFinalizer(previous, GatewayFinalizer)
+		err = r.Update(ctx, previous)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
 
 	if previous.GetDeletionTimestamp() != nil && !previous.GetDeletionTimestamp().IsZero() {
-		// TODO: Do we need to remove dns records and/or endpoints?
-		//       Will ownerRefs be sufficient
-		log.Info("Gateway is deleting", "gateway", previous.Name, "namespace", previous.Namespace)
-		// cleanup dns records
-		// cleunup certificates
+		log.Info("Deleting gateway", "gateway", previous.Name, "namespace", previous.Namespace)
+		//delete dnsRecord
+		err := r.DNSDeletion(ctx, *previous)
+		if err != nil {
+			log.Error(err, "Error deleting DNS record")
+		}
+
+		// Cleanup certificates
 		err = r.cleanupCertificates(ctx, previous)
-		// delete gateway itself if it has no finalizers?
-		return ctrl.Result{}, err
+		if err != nil {
+			log.Error(err, "Error deleting certs")
+		}
+		controllerutil.RemoveFinalizer(previous, GatewayFinalizer)
+		err = r.Update(ctx, previous)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, nil
+
 	}
 
 	// Check if the class name is one of ours
@@ -453,23 +471,40 @@ func selectClusters(gateway gatewayv1beta1.Gateway) []string {
 	return []string{}
 }
 
-// SetupWithManager sets up the controller with the Manager.
-func (r *GatewayReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	return ctrl.NewControllerManagedBy(mgr).
-		For(&gatewayv1beta1.Gateway{}).
-		Watches(&source.Kind{
-			Type: &corev1.Secret{},
-		}, &ClusterEventHandler{client: r.Client}).
-		WithEventFilter(predicate.NewPredicateFuncs(func(object client.Object) bool {
-			gateway, ok := object.(*gatewayv1beta1.Gateway)
-			if !ok {
-				return true
-			}
+func (r *GatewayReconciler) DNSDeletion(ctx context.Context, gateway gatewayv1beta1.Gateway) error {
+	trafficAccessor := traffic.NewGateway(&gateway)
+	allHosts := trafficAccessor.GetHosts()
 
-			return slice.ContainsString(getSupportedClasses(), string(gateway.Spec.GatewayClassName))
-		})).
-		Owns(&kuadrantapi.RateLimitPolicy{}).
-		Complete(r)
+	for _, host := range allHosts {
+		managedZone, subDomain, err := r.Host.GetManagedZoneForHost(ctx, host, trafficAccessor)
+		if err != nil {
+			return err
+		}
+		if managedZone == nil {
+			log.Log.Info("Managed zone not found cannot get DNS record")
+			return nil
+		}
+
+		log.Log.Info("getting DNS record associated with deleting gateway")
+		dnsRecord, err := r.Host.GetDNSRecord(ctx, subDomain, managedZone)
+		if err != nil && !k8serrors.IsNotFound(err) {
+			return err
+		}
+		if dnsRecord == nil {
+			log.Log.Info("dns record not found")
+			continue
+		}
+
+		log.Log.Info("deleting DNS record ", "DNSRecord:", dnsRecord.Name)
+		err = r.Delete(ctx, dnsRecord)
+		log.Log.Info("DNS record deleted ", "DNSRecord:", dnsRecord.Name)
+
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
 
 func (r *GatewayReconciler) cleanupCertificates(ctx context.Context, gateway *gatewayv1beta1.Gateway) error {
@@ -527,4 +562,23 @@ func (r *GatewayReconciler) cleanupCertificates(ctx context.Context, gateway *ga
 	}
 	// assumptions: all traffic objects using the same host are under the same namespace
 	return nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *GatewayReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&gatewayv1beta1.Gateway{}).
+		Watches(&source.Kind{
+			Type: &corev1.Secret{},
+		}, &ClusterEventHandler{client: r.Client}).
+		WithEventFilter(predicate.NewPredicateFuncs(func(object client.Object) bool {
+			gateway, ok := object.(*gatewayv1beta1.Gateway)
+			if !ok {
+				return true
+			}
+
+			return slice.ContainsString(getSupportedClasses(), string(gateway.Spec.GatewayClassName))
+		})).
+		Owns(&kuadrantapi.RateLimitPolicy{}).
+		Complete(r)
 }
--- a/pkg/controllers/gateway/gateway_controller.go
+++ b/pkg/controllers/gateway/gateway_controller.go
@@ -19,7 +19,6 @@ package gateway
 import (
 	"context"
 	"fmt"
-	certman "github.com/jetstack/cert-manager/pkg/apis/certmanager/v1"
 	corev1 "k8s.io/api/core/v1"
 	"reflect"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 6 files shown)


pkg/controllers/gateway/gateway_controller.go
```diff
@@ -351,6 +375,99 @@ func selectClusters(gateway gatewayv1beta1.Gateway) []string {
...
+			return nil
+		}
+
+		log.Log.Info("getting DNS record associated with deleting gateway")
+		dnsRecord, err := r.Host.GetDNSRecord(ctx, subDomain, managedZone)
+		if err != nil && !k8serrors.IsNotFound(err) {
+			return err
+		}
+		if dnsRecord == nil {
+			log.Log.Info("dns record not found")
```
maleck13, line 398: perhaps more useful if we log the host with it when not found?

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -351,6 +375,99 @@ func selectClusters(gateway gatewayv1beta1.Gateway) []string {
...
+			continue
+		}
+
+		log.Log.Info("deleting DNS record ", "DNSRecord:", dnsRecord.Name)
```
maleck13, line 402: not sure this log really adds anything?

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -351,6 +375,99 @@ func selectClusters(gateway gatewayv1beta1.Gateway) []string {
...
+
+	return nil
+}
+
+func (r *GatewayReconciler) cleanupCertificates(ctx context.Context, gateway *gatewayv1beta1.Gateway) error {
+	var hostsToRemove []string
+	var allGateways gatewayv1beta1.GatewayList
+
+	// get names of hosts for traffic object being deleted
+	for _, listener := range gateway.Spec.Listeners {
```
maleck13, line 419: I think we could check the protocol of the listener here 

```listener.Protocol``` should be HTTPS

File: pkg/controllers/gateway/gateway_controller.go [hunk 4]
```diff
@@ -351,6 +375,99 @@ func selectClusters(gateway gatewayv1beta1.Gateway) []string {
...
+		if string(*listener.Hostname) != "*." {
```
maleck13, line 420: not sure I get this check? There would be no cert with *. as a host

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -351,6 +375,99 @@ func selectClusters(gateway gatewayv1beta1.Gateway) []string {
...
+			hostsToRemove = append(hostsToRemove, string(*listener.Hostname))
+		}
+	}
+	// list all the traffic objects
+	err := r.Client.List(ctx, &allGateways, client.InNamespace(gateway.Namespace))
```
maleck13, line 425: do we still need to do this?

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -94,12 +96,34 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 	if previous.GetDeletionTimestamp() != nil && !previous.GetDeletionTimestamp().IsZero() {
-		// TODO: Do we need to remove dns records and/or endpoints?
-		//       Will ownerRefs be sufficient
-		log.Info("Gateway is deleting", "gateway", previous.Name, "namespace", previous.Namespace)
+		log.Info("Deleting gateway", "gateway", previous.Name, "namespace", previous.Namespace)
+		//delete dnsRecord
+		err := r.DNSDeletion(ctx, *previous)
+		if err != nil {
+			log.Error(err, "Error deleting DNS record")
+		}
```
mikenairn, line 113: What happens if you create 2 Gateways with the same host and delete one of them (The second one in particular)? Is it it still adding endpoints from both Gateways and then deleting the DNSRecord when either one of the Gateways is deleted?

pkg/controllers/gateway/gateway_controller.go
R-Lawton: Sure

pkg/controllers/gateway/gateway_controller.go
R-Lawton: Cool ill remove that

pkg/controllers/gateway/gateway_controller.go
R-Lawton: Yes exactly, with docs to come specifying that we do not support multiple gateways having the same hostname. A follow up ticket is to add some check to stop people from being able to implement it not just a warning in a doc

pkg/controllers/gateway/gateway_controller.go
makslion: Would there be a case when based on the protocol we would decide to keep the hostname? Right now it tries to get rid of all of them regardless the protocol

pkg/controllers/gateway/gateway_controller.go
makslion: CRD allows the `*.` notation as a wildcard. Just double-checked regex in the validation section and updated the `if` statement to be functional. In case there will be wildcards in the future it would ensure we are not trying to delete corresponding "wildcard" certificate

pkg/controllers/gateway/gateway_controller.go
makslion: The idea is to check that none of the other Gateways is using this certificate before deleting it. The assumption I'm making here is that all the gateways that might reuse cert will reside in the same namespace (as it should represent a tenant)

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -94,12 +96,34 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		}
 		return ctrl.Result{}, nil
 	}
+	if !controllerutil.ContainsFinalizer(previous, GatewayFinalizer) {
+		controllerutil.AddFinalizer(previous, GatewayFinalizer)
+		err = r.Update(ctx, previous)
```
maleck13, line 101: are we updating here to ensure we always add the finalizer?

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -351,6 +375,96 @@ func selectClusters(gateway gatewayv1beta1.Gateway) []string {
...
+			return err
+		}
+		if managedZone == nil {
+			log.Log.Info("Managed zone not found cannot get DNS record")
+			return nil
+		}
+
+		log.Log.Info("getting DNS record associated with deleting gateway")
+		dnsRecord, err := r.Host.GetDNSRecord(ctx, subDomain, managedZone)
+		if err != nil && !k8serrors.IsNotFound(err) {
```
maleck13, line 394: would there be value in adding a ```r.Host.DeleteDNSForHost()``` that takes does this piece of logic (IE get the managed zone? and deletes the DNSRecord? rather than having the gateway controller have to understand and do that?

maleck13: I feel like line 392 - 405 could become part of the reusable DNS service. At some point we would like to add support an "ingress" gateway so it will also need to do these types of things

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -351,6 +375,96 @@ func selectClusters(gateway gatewayv1beta1.Gateway) []string {
...
+		log.Log.Info("DNS record deleted ", "DNSRecord:", dnsRecord.Name)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (r *GatewayReconciler) cleanupCertificates(ctx context.Context, gateway *gatewayv1beta1.Gateway) error {
```
maleck13, line 411: same for this I think it could becomes part of  ```CertificateService``` WDYT?

pkg/controllers/gateway/gateway_controller.go
maleck13: what we have so far are the following simplistic rules

```
How would work with wildcard hosts assigned to a gateway
A more specific host can be assigned to another Gateway e.g. i can use myapp.example.com in another Gateway even though *.example.com is specified in another gateway. This is expected behaviour when you are more specific with a hostname. Each more specific host would receive its own certificate. 
2 hostnames of the same value cannot be used in 2 different Gateways. If a hostname is required in more than 1 Gateway, use a single Gateway instead of 2. If there’s a use case for using a hostname in 2 different Gateways, we need to understand that better as we may be missing a feature
For the first Gateway implementation, each host will get its own certificate for that host. If a wildcard host is specified in a listener, a wildcard certificate (and dns record) will be used

```

maleck13: So with a rule that a host is unique to a gateway and each host gets its own certificate, I think we can now just remove the certificates associated with that gateway

pkg/controllers/gateway/gateway_controller.go
R-Lawton: yes

maleck13 (commented):
Couple of comments

I think we should bring DNSDeletion and Certificate Deletion into the interfaces and services we provide for other actions of this type so that they can be reused for other potential gateway types

I think we can make an assumption at this point that a host is unique to a gateway and each host will get it own cerificate. So when a gateway is deleted we can remove all associated certifcates and DNSRecords for any hosts in that gateway.

pkg/controllers/gateway/gateway_controller.go
R-Lawton: Ok yes if we are planning on doing this again that makes sense

pkg/controllers/gateway/gateway_controller.go
philbrookes: shouldn't this if statement also check that the resource is not deleting (i.e. don't re-add the finalizer when we want it removed)?

pkg/controllers/gateway/gateway_controller.go
makslion: Pushed changes. It make my life much easier :laughing:

pkg/controllers/gateway/gateway_controller.go
makslion: added booth deletion functions to interfaces. Hope I got the idea about "ingress gateway" correct :thinking:

pkg/controllers/gateway/gateway_controller.go
makslion: added check for the deletion timestamp - should prevent the addition of the finalizer if we are deleting the gateway. And requeued (is it even a word? :thinking: ) it

makslion:
Rebased. 
Summary (mostly for myself because I'm getting lost in all the replies): 
* Moved booth deletion calls into interfaces
* Cert gets deleted by the `ownerReference` - no need to make a specific `.Delete()` call for this. 
* Added check for the deletion timestamp before adding finalizer and requeueing it after we put the finalizer back in 
* Greatly simplified logic for the cert deletion - the assumption is that Gateway listener to cert is a 1:1 relationship. If we are deleting the gateway we are always deleting all the certs and they (certs) will match the hostnames on listeners.

Questions: 
* If we are deleting the DNSRecord - it is owned by the ManagedZone CR - should we look into deleting that one as well? 
* In case we would update Gateway by removing one of the listeners we, most likely, should also clean up related DNSRecord and Certificate - correct? Would it mean getting (again) the list of all the Gateways and comparing it to the list of all the Certs/DNSRecords to find the "extra" ones to be cleaned up on each reconcile loop? :thinking:

philbrookes:
When I went through the verification, I did not create an application which I believe means no DNS Records were created in Route53.

I could see that the DNS Records, cert and secret were created, but when I deleted the gateway, the DNS Record was not cleaned up. Looking at the errors in the controller, I suspect this is because no DNS Records were ever created:
```
2023-05-02T12:39:58+02:00	ERROR	Reconciler error	{"controller": "dnsrecord", "controllerGroup": "kuadrant.io", "controllerKind": "DNSRecord", "DNSRecord": {"name":"myapp.pb.hcpapps.net","namespace":"multi-cluster-traffic-controller-system"}, "namespace": "multi-cluster-traffic-controller-system", "name": "myapp.pb.hcpapps.net", "reconcileID": "061c7d18-012b-4a34-a76a-900918b21b22", "error": "failed to update record in route53 hosted zone /hostedzone/Z0490617QKSNPXXQRPII: no endpoints"}
```

I think this is a scenario that the controller should be able to cope with, though.

mikenairn:
> > * If we are deleting the DNSRecord - it is owned by the ManagedZone CR - should we look into deleting that one as well?
> 

No. A managed zone can own many DNSRecords, it should not be deleted just because you are removing a record.

maleck13:
> > > * If we are deleting the DNSRecord - it is owned by the ManagedZone CR - should we look into deleting that one as well?
> 
> No. A managed zone can own many DNSRecords, it should not be deleted just because you are removing a record.

Yes agree, MZ is the top level resource in the DNS "tree" records can be in many gateways. Creating and Deleting a MZ should be a deliberate step just like creating and deleting a gateway

philbrookes:
> > > > * If we are deleting the DNSRecord - it is owned by the ManagedZone CR - should we look into deleting that one as well?
> > 
> > 
> > No. A managed zone can own many DNSRecords, it should not be deleted just because you are removing a record.
> 
> Yes agree, MZ is the top level resource in the DNS "tree" records can be in many gateways. Creating and Deleting a MZ should be a deliberate step just like creating and deleting a gateway

And for the reverse scenario? It makes sense to me that if a MZ is deleted, we should ensure all related DNS Records are removed, correct?

mikenairn:
> > > > > * If we are deleting the DNSRecord - it is owned by the ManagedZone CR - should we look into deleting that one as well?
> > > 
> > > 
> > > No. A managed zone can own many DNSRecords, it should not be deleted just because you are removing a record.
> > 
> > 
> > Yes agree, MZ is the top level resource in the DNS "tree" records can be in many gateways. Creating and Deleting a MZ should be a deliberate step just like creating and deleting a gateway
> 
> And for the reverse scenario? It makes sense to me that if a MZ is deleted, we should ensure all related DNS Records are removed, correct?

Yes, and that should already be happening due to the owner ref.

makslion:
The issue #178 was created to cover the "update gateway" case while getting rid of one of the listeners. And with the discussion about ManagedZone CR would there be anything I could improve on this PR? :thinking:

maleck13:
@makslion is this ready for another review?

makslion:
@maleck13 yes

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -93,16 +96,38 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	if err != nil {
 		if err := client.IgnoreNotFound(err); err != nil {
 			log.Error(err, "Unable to fetch Gateway")
-			return ctrl.Result{}, err
+			return ctrl.Result{Requeue: true, RequeueAfter: time.Second * 10}, err
```
maleck13, line 99: from reading the controller runtime code, if an error is returned then it will re-queue so no value in returning the result with a requeue

```
result, err := c.Reconcile(ctx, req)
	switch {
	case err != nil:
		c.Queue.AddRateLimited(req)
		ctrlmetrics.ReconcileErrors.WithLabelValues(c.Name).Inc()
		ctrlmetrics.ReconcileTotal.WithLabelValues(c.Name, labelError).Inc()
		log.Error(err, "Reconciler error")
	case result.RequeueAfter > 0:
		// The result.RequeueAfter request will be lost, if it is returned
		// along with a non-nil error. But this is intended as
		// We need to drive to stable reconcile loops before queuing due
		// to result.RequestAfter
		c.Queue.Forget(obj)
		c.Queue.AddAfter(req, result.RequeueAfter)
		ctrlmetrics.ReconcileTotal.WithLabelValues(c.Name, labelRequeueAfter).Inc()
	case result.Requeue:
		c.Queue.AddRateLimited(req)
		ctrlmetrics.ReconcileTotal.WithLabelValues(c.Name, labelRequeue).Inc()
	default:
		// Finally, if no error occurs we Forget this item so it does not
		// get queued again until another change happens.
		c.Queue.Forget(obj)
		ctrlmetrics.ReconcileTotal.WithLabelValues(c.Name, labelSuccess).Inc()
	}
```

maleck13: So I think the above can be adjusted to just return 
``` ctrl.Result{}, err```

Comment (makslion) [reply pkg/controllers/gateway/gateway_controller.go]: updated

File: pkg/controllers/gateway/gateway_controller.go [hunk 11]
```diff
@@ -19,12 +19,12 @@ package gateway
...
 	"reflect"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"strings"
 	"time"
 
-	corev1 "k8s.io/api/core/v1"
-
 	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/_internal/conditions"
 	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/_internal/metadata"
 	"github.com/Kuadrant/multi-cluster-traffic-controller/pkg/_internal/slice"
```
mikenairn, line 30: Consider grouping your imports, stdlib, ks8, current project etc..

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -97,12 +100,34 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		}
 		return ctrl.Result{}, nil
 	}
+	if !controllerutil.ContainsFinalizer(previous, GatewayFinalizer) && previous.GetDeletionTimestamp().IsZero() {
+		controllerutil.AddFinalizer(previous, GatewayFinalizer)
+		err = r.Update(ctx, previous)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
```
mikenairn, line 109: This should probably all be moved to after the deletion steps https://github.com/Kuadrant/multicluster-gateway-controller/pull/162/files#diff-6afa7d73841a5c0ebdb78eaf859ce5a400203a037f871440940b5347c3a97376R111. I would have considered the first thing any reconciler should do is check if we are deleting it first.

pkg/dns/service.go
```diff
@@ -3,6 +3,9 @@ package dns
 import (
 	"context"
 	"fmt"
+	v12 "k8s.io/api/networking/v1"
+	"reflect"
+	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
```
mikenairn, line 8: import grouping

pkg/dns/service.go
```diff
@@ -431,6 +434,48 @@ func (s *Service) PatchTargets(ctx context.Context, targets, hosts []string, clu
...
+			}
+		}
+
+		return nil
+	}
+	// TODO ingress case
+	_, ok = owner.(*v12.Ingress)
+	if ok {
+		return nil
+	}
```
mikenairn, line 475: If you just pass the traffic.Interface into this method you won't need to worry about this.

pkg/tls/service.go
```diff
@@ -53,6 +59,33 @@ func (s *Service) GetCertificateSecret(ctx context.Context, host string, namespa
...
+		}
+		return nil
+	}
+	// TODO ingress case
+	_, ok = owner.(*v12.Ingress)
+	if ok {
+		return nil
+	}
+	return fmt.Errorf("uknkown object type for a certificate deletion: %s", reflect.TypeOf(owner))
+}
```
mikenairn, line 87: Same here, use  traffic.Interface instead of interface{}, and make the reconciler(gateway, ingress) pass it in.

mikenairn:
Left some comments. I would say though that i think we need to setup a relationship between the DNSRecord and the Gateway sooner rather than later since this as it is can cause DNSRecords to be deleted that are being used by other Gateway resources. I think that is what is being proposed here https://github.com/Kuadrant/multicluster-gateway-controller/issues/178, but i almost think that should be done before we start adding deletion logic.

Admittedly though nobody is using this yet, so probably not a big deal to do it this way round.

pkg/dns/service.go
makslion: Not sure that Gateway implements `traffic.interface` :thinking:

pkg/tls/service.go
makslion: gateway type does not implement `traffic.interface` if I'm not mistaken

pkg/controllers/gateway/gateway_controller.go
makslion: sorted

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -47,18 +47,21 @@ import (
...
+	GatewayFinalizer                      = "kuadrant.io/gateway"
 )
 
 type HostService interface {
 	CreateDNSRecord(ctx context.Context, subDomain string, managedZone *v1alpha1.ManagedZone) (*v1alpha1.DNSRecord, error)
 	GetDNSRecord(ctx context.Context, subDomain string, managedZone *v1alpha1.ManagedZone) (*v1alpha1.DNSRecord, error)
 	GetManagedZoneForHost(ctx context.Context, domain string, t traffic.Interface) (*v1alpha1.ManagedZone, string, error)
 	AddEndpoints(ctx context.Context, t traffic.Interface, dnsRecord *v1alpha1.DNSRecord) error
+	DNSDeletion(ctx context.Context, owner interface{}) error
 }
```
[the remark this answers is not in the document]
makslion: renamed

pkg/dns/service.go
makslion: sorted

makslion:
> Left some comments. I would say though that i think we need to setup a relationship between the DNSRecord and the Gateway sooner rather than later since this as it is can cause DNSRecords to be deleted that are being used by other Gateway resources. I think that is what is being proposed here #178, but i almost think that should be done before we start adding deletion logic.
> 
> Admittedly though nobody is using this yet, so probably not a big deal to do it this way round.

Addressed comments. 
Not sure could use the gateway as the `traffic.interface` object, but yes, that would make the code much more elegant. Haven't sound any of the interfaces that Gateway implements to do the work :thinking: 
 And I had a strong impression that DNSRecord CR can be related to one and only one DNSRecord, unless I misunderstood [this](https://github.com/Kuadrant/multicluster-gateway-controller/pull/162#discussion_r1177580436) comment

maleck13:
```upstreamAccessor := traffic.NewGateway(upstreamGateway)``` will give you the implementation of the traffic interface

pkg/dns/service.go
mikenairn: You are already getting here https://github.com/Kuadrant/multicluster-gateway-controller/pull/162/files#diff-547de635669379196f2ffdf2e286b12a8736b8ba08e18cd156a238f2c3fb2578R440. I'm suggesting you pass that into the function from the controller that already knows what type of resource you are dealing with.

mikenairn:
> And I had a strong impression that DNSRecord CR can be related to one and only one DNSRecord, unless I misunderstood [this](https://github.com/Kuadrant/multicluster-gateway-controller/pull/162#discussion_r1177580436) comment

A DNSRecord is currently created for every host in a gateway that has a managed zone in this instance of MGC for that domain. This creates records in the dns provider, one for each ip address pulled from the traffic resource it's being used by (Gateway). This is an issue if more than one gateway tries to use the same host, as any additional gateway reconciles will find the existing dns record and start adding their endpoints to it causing all sorts of weird issues. This PR also now deletes DNS records, which currently could be getting updated by more than one gateway, leaving the existing gateway without a DNS record.

A DNSRecord represents one host, and a ost shoud only be used by one gateway, so essentially we need to be able say that a DNSRecord belongs to this Gateway and any other gateways trying to use that host should get an error of some description in their status. This could be done via owner refs, but you have to consider that a DNSRecord can exist in a different ns, so i think @R-Lawton had been considering using annotations for this.

pkg/dns/service.go
makslion: implemented

makslion:
How about something like this @mikenairn ? :thinking: 
If I did things correctly we now will not try to mess around with DNS records created for a different Gateway plus this lets us list all the records created for a specific gateway. 
And I finally found the `.NewGateway()` function and the deletion logic is much simpler right now

philbrookes (approved):

PR #163, opened 2023-04-14
Control plane RateLimitPolicy

### LINKED ISSUE #116: Multi-Cluster rate limit
**What**

Being able to define a rate limit policy within the multi-cluster control plane based on a multi-cluster topology that are informed by the multi-cluster topology and are enforced by the data plane.

**Status**
When defining a multi-cluster gateway and defining a RLP targeting that gateway. Understanding the status of the rate limiting on a given cluster is important from a admin using the control plane point of view. When I have created a RLP targeting a mc gateway, I also want to know that it has been synced and applied to the underlying clusters. 
Ideally I think we would add a aggregated status to the RLP but also a  condition to the gateway status. 

On the gateway it should be an all or nothing IMO. IE is my gateway protected or not

```
 - lastTransitionTime: "2023-04-14T14:46:18Z"
    message: Rate Limit Policy  $name applied to all gateways
    observedGeneration: 1
    reason: Applied
    status: "True"
    type: Applied

```

On the RLP it should probably provide a condition for each target cluster?

```
 - lastTransitionTime: "2023-04-14T14:46:18Z"
    message: Rate Limit Policy  $name applied to gateway on $cluster
    observedGeneration: 1
    reason: Applied
    status: "True"
    type: Applied

```

Only if all RLP conditions are True can the gateway be True 
https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/docs/status_aggregation.md

**Why**

- Being able to declare policy in the same place you create your multi-cluster gateway have that policy be distributed to and applied to multiple instances of a gateway and enforced in that gateways data plane
- Having the attributes of your cluster topology automatically available as well known descriptors allows for simpler and richer options for providing rate limits based on your infrastructure

**Use Cases**

*Infra Rate Limiting*

As a gateway administrator, I want to define a limit that restricts the total number of requests that each instance of a uniform gateway will safely handle within a set period of time before beginning to reject requests so that I can ensure my gateway will remain up and running and importantly my cluster resources are protected so that I can access the cluster to restrict the gateway or certain users/IPs further. As instances of this gateway is running on multiple clusters, I want to define this limit in a single place that allows it to target my gateway definition

*Global Rate Limiting*

As a gateway administrator, I want to define a default limit per identifier of my choice to restrict access to endpoints exposed via my multi-cluster gateway to a sane default and ensure no unprotected endpoint is exposed that may allow consuming the entire limit for the gateway and provide fair access by default to those accessing endpoints attached to the gateway. I want to define this in the same place I define and manage my multi-cluster gateway

- [x] Install Rate Limiting CRD into the control plane 
- [x] Add new Rate Limiting Controller to the Gateway Traffic Controller
- [x] Add Transforms to inject cluster specific attributes
- [ ] Add a multi-cluster gateway and RLP status

**Demo Outline**

- Define a gateway and place it onto two data plane clusters with Kuadrant installed
- Deploy HTTPRoute to the gateway
- Observe no rate limiting happening
- Define a RLP at the control plane targeting the gateway 
- Observe that the attributes of the clusters are added to each RLP that is synced (control plane added these)
- Observe that requests are limited based on the header
- Define a new limit based with a variable based on the cluster id (which has been injected as a descriptor)
- Observe each user limited regardless of the cluster they hit
- Define a more restrictive limit for /upload that targets the HTTPRoute attached to the gateway
- Observe the more restrictive limit now takes effect

The above assumes the following
- Kuadrant installed on each DP cluster
- Some form of shared storage between the DP clusters

mikenairn:
closes #116 

Adds a RateLimitPolicy reconciler that will reconcile kuadrant RateLimitPolicies to apply mctc sync and patch annotations based on the RLPs target owner Gateway. If the target Gateway is using the wildcard `all` sync annotation, a list of available clusters is queried, and individual sync annotations added for each. A cluster specific patch annotation is added for each cluster the RLP is being synced to which adds an RLP configuration action using the cluster name as a key.
This will automatically be made available for use in the limits section of the RLP allowing gateway admins to add cluster specific limits.

* Add kuadrant RateLimitPolicy CRD to control plane cluster.
* Add RateLimitPolicy controller and kuadrant-operator APIs as a dependency.
* Update sigs.k8s.io/gateway-api to v0.6.2.
* Add ratelimitpolicies.v1beta1.kuadrant.io to default syncer synced resources.
* Use gateway-api-v0.6.2 version of kuadrant
* Set gateway as owner of RLP.
* Add sync annotations based on target gateway sync annotations.

Includes a commit from https://github.com/Kuadrant/multicluster-gateway-controller/pull/124 (f9ffbe106386c7dd347bdf471aa75a45780282de), we should try and merge that first, or pull these out into their own PR. 

Requires https://github.com/Kuadrant/kuadrant-operator/pull/153. We can merge this using the `gateway-api-0.6.2` kuadrant operator branch, but we should probably try and get that kuadrant operator change merged first and revert to main.

**Verification**

Follow this doc https://github.com/Kuadrant/multicluster-gateway-controller/blob/e1b74f4040760fa5893eb15495b1b38f90a489fb/docs/ratelimitpolicy/walkthrough/rlp_walkthrough.md

- [x] tests
- [x] docs
~~- [ ] gateway status~~ Deferring to after ACM POC work so we are clear how we are doing this
~~- [ ] aggregated RLP status~~ Deferring to after ACM POC work so we are clear how we are doing this


pkg/controllers/ratelimitpolicy/ratelimitpolicy_controller.go
```diff
@@ -0,0 +1,257 @@
...
+
+func (r *RateLimitPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	_ = log.FromContext(ctx)
+
+	previous := &kuadrantapi.RateLimitPolicy{}
+	err := r.Client.Get(ctx, client.ObjectKey{Namespace: req.Namespace, Name: req.Name}, previous)
+	if err != nil {
+		if err := client.IgnoreNotFound(err); err == nil {
+			return ctrl.Result{}, nil
+		} else {
```
maleck13, line 64: prob can remove the else?

pkg/controllers/ratelimitpolicy/ratelimitpolicy_controller.go
```diff
@@ -0,0 +1,257 @@
...
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	targetGateway, err := r.getTargetGateway(ctx, rateLimitPolicy)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
```
maleck13, line 102: wonder could we move this to line 81 and just add

```
err = controllerutil.SetControllerReference(targetGateway, rateLimitPolicy, r.Scheme)
	if err != nil {
		return err
	}

```
to line 87?

maleck13: is the ```setTargetOwner``` worth its own function?

maleck13:
@mikenairn i didn't see anything around the control plane level gateway status in the PR?
the final item on #116 is to add multi-cluster gateway status. However I also see there is not a lot of details there. I will add some more detail. I don't think it blocks this PR but would be good to cover it in a follow up

mikenairn:
> @mikenairn i didn't see anything around the control plane level gateway status in the PR? the final item on #116 is to add multi-cluster gateway status. However I also see there is not a lot of details there. I will add some more detail. I don't think it blocks this PR but would be good to cover it in a follow up. Added some more details to that issue.
> 
> @mikenairn Ignore me. I see you called it out at the end ! I need to read crossed out stuff :) holding off for now is fine

Yeah, i did look into this for a while, but since we deferred the gateway aggregated status, and the RLP in the control plane isn't going to have a status until we do that, i decided to leave it out for now. Will add a follow up task to add it.

pkg/controllers/ratelimitpolicy/ratelimitpolicy_controller.go
mikenairn: Yeah, we might be able to get rid of it tbh, i just start out that way so i can easily see the steps it has to take, and we need to test.

pkg/controllers/ratelimitpolicy/ratelimitpolicy_controller.go
```diff
@@ -0,0 +1,257 @@
...
+		annotations[syncer.MCTC_SYNC_ANNOTATION_PREFIX+cluster.Name] = "true"
+	}
+
+	obj.SetAnnotations(annotations)
+}
+
+// setPatchAnnotationsForClusters adds patch annotations to the given RateLimitPolicy for each cluster.
+// Adds a patch annotation to the RLP that injects an Action with a generic key (key: cluster name, value: 1) which
+// can then be used in limit conditions.
+// ToDo add attributes from an annotation on the cluster secrets
```
mikenairn, line 234: @maleck13 fyi, the reason I'm referring to the secrets everywhere  and not relying on just the name is so we can do this. It would be fairly straightforward to add, but did you have an idea of what you wanted this to look like? Was the idea to add a single annotation with a bob of json ( i.e. {"provider": "aws"}), or multiple annotations with a single key/value that we merge together?

maleck13: I had thought of it as multiple annotations. ```kuadrant.io/attribute-cloud: aws kuadrant.io/attribute-geo: EU-West``` so basically a convention

maleck13:
working through the walkthough

pkg/controllers/ratelimitpolicy/ratelimitpolicy_controller.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *RateLimitPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		// Uncomment the following line adding a pointer to an instance of the controlled resource as an argument
+		For(&kuadrantapi.RateLimitPolicy{}).
```
maleck13, line 126: should we add a watch for the cluster secrets?

mikenairn: Secrets are already being watched by the Gateway Reconciler, i just updated that to include RateLimitPolicies as a thing to update as required https://github.com/Kuadrant/multicluster-gateway-controller/pull/163/commits/bafcc3ab7399975217a561e885338f05e05ab58d

mikenairn:
> working through the walkthough
> 
> One minor change as mentioned in chat:
> 
> currently we set the descriptor key for the cluster to be
> 
> ```
> configurations:
>       - actions:
>         - generic_key:
>             descriptor_key: mctcworkload1
>             descriptor_value: "1"
> ```
> 
> I think it would make more sense and be easier to communicate via docs if it was under a well known key example:
> 
> ```
> - configurations:
>       - actions:
>         - generic_key:
>             descriptor_key: kudarant_gateway_cluster
>             descriptor_value: "mctcworkload1"
> ```

Updated

mikenairn:
/hold

I need top update this to pull the cluster name from the secret data rather than using the secret name

maleck13:
so now if a cluster changes (IE an attribute annotation is added) that triggers a gateway reconcile that enques any RLP targeting that gateway? Just trying to understand the flow.

mikenairn:
> so now if a cluster changes (IE an attribute annotation is added) that triggers a gateway reconcile that enques any RLP targeting that gateway? Just trying to understand the flow.

Not exactly, it just triggers the reconciliation of a Gateway and an RLP. The cluster event handler was already there for Gateways to get updates when an associated secret changes, this just adds the RLPs to that. If a secret changes, it checks alls the RLPs and triggers a reconcile of any that are being synced to that cluster.

pkg/controllers/ratelimitpolicy/ratelimitpolicy_controller.go
```diff
@@ -0,0 +1,206 @@
...
+
+	if !controllerutil.ContainsFinalizer(rateLimitPolicy, RateLimitPolicyFinalizer) {
+		log.Log.Info("RateLimitPolicyFinalizer Reconcile: Added finalizer and set gateway owner", "rateLimitPolicy", rateLimitPolicy.Name)
+
+		controllerutil.AddFinalizer(rateLimitPolicy, RateLimitPolicyFinalizer)
+
+		err = r.Update(ctx, rateLimitPolicy)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
```
maleck13, line 101: should we add a return after this if as I think the update will trigger a new reconcile and currently this reconcile would continue?

maleck13:
I haven't re-tested but generally looking fine. 
@mikenairn ok to put a demo together for this functionality?

maleck13 (approved):

maleck13:
went through initial verification. several code reviews

/lgtm

Issue #105, opened 2023-03-09
AWS DNSHealthCheck


maleck13:
**Use Case**

As a gateway admin managing a set of gateways across multiple clusters, I want to define a common health check that will govern whether a DNS A record is considered healthy so that with a single policy, I can ensure the DNS response for services exposing endpoints and using listeners on my gateway are only returning endpoints that are from healthy service endpoints allowing us to automatically and rapidly mitigate an unexpected outage. I also want to define a strategy for if all the endpoints become available. I want to define a minimum number of records per hostnames so that my health check doesn't remove all records . 

**What**
This is an API that targets a gateway and encapsulates an end point check. We should base it on what we provided via annotations in the GLBC https://github.com/kcp-dev/kcp-glbc/blob/main/docs/dns/health-checks.md

**Constraint**
At this point the endpoint has to be publicly reachable. In the future we may look at other forms of check that do not require this. Example something running at the gateway level that performs the check and reports back via a status condition on the listeners status block

- [x] #98 
- [x] #106 
- [x] #107

david-martin:
Closing as done.
@sergioifg94 to create a new issue for provider agnostic health checks

PR #169, opened 2023-04-20
#105: Implement DNS Health Check reconciliation

sergioifg94:
## Description

Add health check spec to DNSPolicy and implement reconciliation of the health checks on Route 53

## Verification steps

In order to verify, run the controller locally alongside the example application to ensure endpoints are added. Then, create the example DNSPolicy and verify that the status is updated and the health check is created correctly

1. Run the local setup

    ```sh
    ```

1. Run the controller

    ```sh
    ```

1. Create the gateway class and the tenant namespace

    ```sh
    kubectl apply -f config/samples/gatewayclass.yaml
    ```

1. In another terminal, target the workload cluster, and run the syncer

    ```sh
    make kind-load-syncer
    make deploy-syncer

    kubectl wait --for=condition=Available deployment sync-agent -n mctc-system
    ```

1. In the terminal that targets the control plane, create the gateway

    ```sh
    kubectl apply -n mctc-tenant -f gateway.yaml
    ```

1. In the terminal that targets the workload cluster, deploy the application

    ```sh
    kubectl apply -f application.yaml
    ```

1. In the terminal that targets the control plane, create the DNSPolicy

    ```sh
    kubectl apply -n mctc-tenant -f config/samples/dnspolicy.yaml
    ```

1. The controller will reconcile the DNSPolicy and update the status with the `Created` condition in the health check section

    ```sh
    kubectl get dnspolicy example-dnspolicy -n mctc-tenant -o yaml
    ```

1. In the health check section of Route 53, in the AWS console, the new health check will be visible

    image

1. Verify in the AWS console that the DNS record for the app references the newly created health check


pkg/controllers/dnspolicy/health.go
```diff
@@ -0,0 +1,217 @@
...
+}
+
+func (r *DNSPolicyReconciler) getDNSRecord(ctx context.Context, policy *v1alpha1.DNSPolicy) (*v1alpha1.DNSRecord, bool, error) {
+	gateway, err := r.getGateway(ctx, policy)
+	if err != nil {
+		return nil, false, err
+	}
+
+	dnsRecordList := &v1alpha1.DNSRecordList{}
+	if err := r.Client.List(ctx, dnsRecordList); err != nil {
```
maleck13, line 152: does this list all records? I think we could def limit it to the same namespace as the gateway?

maleck13: actually not seeing where dnsRecordList is used ?

sergioifg94: Oops, that's actually a bit that I forgot to remove after I changed the implementation to get the DNSRecord

pkg/controllers/dnspolicy/health.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+	trafficAccessor := traffic.NewGateway(gateway)
+	allHosts := trafficAccessor.GetHosts()
+
+	if len(allHosts) == 0 {
+		return nil, false, nil
+	}
+
+	host := allHosts[0]
```
maleck13, line 163: why only the first one?

sergioifg94: My understanding is that it's a 1:1 relation between Gateways and DNSRecords, so there should be only one associated DNSRecord to the gateway. If I'm wrong then I'll update this to create a health check per DNSRecord

maleck13: no it is   a 1:1 releationship between a `hostname` and a dnsrecord and it is a 1:1 relationship between a specific subdomain and a gateway but a gateway can have many hostnames and so many DNSRecords

sergioifg94: I see. That makes more sense tbh. I'll make the changes

pkg/controllers/dnspolicy/health.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	if len(dnsRecords) == 0 {
+		log.Info("No DNSRecord found matching policy", "policy", policy)
+		return nil
+	}
+
+	// Get the configuration for the health checks. If no configuration is
+	// set, ensure that the health checks are deleted
+	config := getHealthChecksConfig(policy)
+	log.V(3).Info("No health check for DNSPolicy, ensuring deletion")
```
maleck13, line 40: this log looks wrong?

maleck13:
@sergioifg94 given this a first pass now. I haven't done verification yet. Generally looking in ok shape. I know we are iterating quickly but I think some basic unit tests here would be appropriate WDYT?

@philbrookes could you take a look through and verify

philbrookes:
Verified, but ran into an issue where the dnspolicy reconcile isn't triggered by the related dns record update / add events.

@sergioifg94 is going to update to address this, and I will re-verify then.

pkg/controllers/dnspolicy/dnsrecord_eventhandler.go
```diff
@@ -0,0 +1,78 @@
...
+
+func (eh *DNSRecordEventHandler) enqueueForObject(obj v1.Object, q workqueue.RateLimitingInterface) {
+	dnsRecord, ok := obj.(*v1alpha1.DNSRecord)
+	if !ok {
+		return
+	}
+
+	dnsPolicyList := &v1alpha1.DNSPolicyList{}
+	if err := eh.client.List(context.TODO(), dnsPolicyList); err != nil {
+		log.Log.Error(err, "unexpected error listing DNSPolicies when enqueinng DNSRecord", "dnsRecord", dnsRecord)
```
philbrookes, line 53: nit: `enqueinng` > `enqueuing`

pkg/controllers/dnspolicy/dnsrecord_eventhandler.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func (eh *DNSRecordEventHandler) enqueueForObject(obj v1.Object, q workqueue.RateLimitingInterface) {
+	dnsRecord, ok := obj.(*v1alpha1.DNSRecord)
+	if !ok {
+		return
+	}
+
+	dnsPolicyList := &v1alpha1.DNSPolicyList{}
+	if err := eh.client.List(context.TODO(), dnsPolicyList); err != nil {
```
philbrookes, line 52: is the intention to change this context eventually? If not, I think it's better to use `context.Background`

philbrookes: functionally they are the same, but `context.Background` suggests the use was intentional where `context.TODO` suggests that it should be changed to something else eventually.

pkg/dns/aws/health.go
```diff
@@ -0,0 +1,252 @@
...
+var _ dns.HealthCheckReconciler = &Route53HealthCheckReconciler{}
+
+func NewRoute53HealthCheckReconciler(client route53iface.Route53API) *Route53HealthCheckReconciler {
+	return &Route53HealthCheckReconciler{
+		client: client,
+	}
+}
+
+func (r *Route53HealthCheckReconciler) Reconcile(ctx context.Context, spec dns.HealthCheckSpec, endpoint *v1alpha1.Endpoint) (dns.HealthCheckResult, error) {
+	healthCheck, exists, err := r.findHealthCheck(ctx, endpoint)
```
philbrookes, line 41: Is this going to the AWS API on every reconcile? Could this lead to us being rate limited?

pkg/dns/aws/health.go
```diff
@@ -0,0 +1,252 @@
...
+	return dns.NewHealthCheckResult(dns.HealthCheckDeleted, ""), nil
+}
+
+func (c *Route53HealthCheckReconciler) findHealthCheck(ctx context.Context, endpoint *v1alpha1.Endpoint) (*route53.HealthCheck, bool, error) {
+	id, hasId := getHealthCheckId(endpoint)
+	if !hasId {
+		return nil, false, nil
+	}
+
+	response, err := c.client.GetHealthCheckWithContext(ctx, &route53.GetHealthCheckInput{
```
philbrookes, line 92: Could we cache this?

philbrookes: maybe cache it with the generation of the policy that caused it? So it invalidates when the policy is changed?

philbrookes (commented):
Verified locally, mostly looks fine, just a concern over how much this looks to be accessing the Route53 API.

sergioifg94:
@philbrookes I just pushed a couple of commits addressing your suggestions

pkg/dns/health.go
```diff
@@ -60,3 +61,62 @@ func (*FakeHealthCheckReconciler) Delete(ctx context.Context, _ *v1alpha1.Endpoi
 }
 
 var _ HealthCheckReconciler = &FakeHealthCheckReconciler{}
+
+type CachedHealthCheckReconciler struct {
+	reconciler HealthCheckReconciler
+	provider   Provider
+
+	cache map[string]HealthCheckSpec
```
maleck13, line 69: do we need to worry about concurrent access with this cache?

sergioifg94: It would probably be wiser to make it thread-safe but as the health check reconciler is only being used by the DNSPolicy controller it shouldn't be a problem

maleck13: Yes but I would imagine each object event is handled in its own routine to avoid blocking. So multiple go routines will access this shared map. Granted they may not clash but probably safer to protect access to it anyway

maleck13: actually I am wrong by default there is only 1 worker. However this can be tweaked with config

maleck13: so I think a threadsafe map makes sense as the default

sergioifg94: Thanks @maleck13 I just pushed a commit with the change using sync.Map as the cache

pkg/controllers/dnspolicy/dnsrecord_eventhandler.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (eh *DNSRecordEventHandler) enqueueForObject(obj v1.Object, q workqueue.RateLimitingInterface) {
+	ctx := context.Background()
+
+	dnsRecord, ok := obj.(*v1alpha1.DNSRecord)
+	if !ok {
+		return
+	}
+
+	dnsPolicyList := &v1alpha1.DNSPolicyList{}
```
maleck13, line 53: This looks like it will list all policies in the cluster?  Should list only policies in the same namespace as the DNSRecord I believe

sergioifg94: Should we restrict them to be only in the same namespace? At the moment I'm creating the DNSPolicy in the tenant namespace, whereas the DNSRecord is created in the system namespace, so that wouldn't work

maleck13: I think with the change to leverage OCM we should start with the expectation that MZ and DNSRecord will exist in the same namespace. So lets change this as part of a follow up. At some stage if we have the usecase we may look to introduce a MZ being shared / bound  to other namespces. For example one potential option would be a `subdomain` resource that references and zone resource. Similar to how a HTTPRoute references a Gateway cc @mikenairn do this make sense to you also?  
It may also be worth considering whether MZ should be a cluster level resource.

philbrookes (approved):
/lgtm

PR #184, opened 2023-05-08
in cluster addon manager

R-Lawton:


config/ocm/cluster-role-binding.yaml
```diff
@@ -9,4 +9,4 @@ roleRef:
 subjects:
   - kind: ServiceAccount
     name: kuadrant-addon-sa
-    namespace: open-cluster-management
\ No newline at end of file
+    namespace: ocm-addon-manager
```
maleck13, line 12: should we use this namespace or something more descriptive? ```kuadrant-addon-manager```

maleck13 (commented):
Question on the namespace.

Issue #114, opened 2023-03-10
Quality: bring up our coverage of the gateway reconciler


maleck13:
**What**

We have moved fast with supporting gateways in the control plane in order to get a steel thread and clearer understanding. It is important we take the time to go back and pay down some of the debt.

- [x]  Add Unit Coverage
- [x]  Add Integration test (after unit) to cover the deletion (covered by #261 )

PR #175, opened 2023-04-26
OCM hub integration

maleck13:
Should be in an operational state 
Still to do

- tidy up and go through code comments etc linting etc
- rebase
- improve walkthrough

#verification 

Go through the walkthrough doc
https://github.com/Kuadrant/multicluster-gateway-controller/pull/175/files#diff-d9a0f59f1e1d060dd48a5280234d1fa3fbc36aec23bdb679af3d2884eb84e0b9

Key Points Proven out

- Status integration from spoke and hub
- Placement integration 
- Work API integration 

Remaining Work

- [ ] Get the e2e tests passing again
- [ ] Add unit tests for placement and dns service
- [ ] Update to use Raw status response

PR #186, opened 2023-05-09
GH-114 increase coverage in gateway package

makslion:
## What
Coverage change for the gateway package to 90.1%
Extended unit test coverage and created test files for booth controllers. 
No work on the integration right now. 

## Verification
Verification is straightforward - tests are passing on the PR and coe looks neat and covers all the cases it should - PR is verified.

### DIFF
--- a/pkg/controllers/gateway/cluster_eventhandler_test.go
+++ b/pkg/controllers/gateway/cluster_eventhandler_test.go
@@ -16,6 +16,7 @@ import (
 	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 
 	"github.com/Kuadrant/multicluster-gateway-controller/pkg/_internal/clusterSecret"
+	"github.com/Kuadrant/multicluster-gateway-controller/test/util"
 )
 
 func TestClusterEventHandler(t *testing.T) {
@@ -75,7 +76,7 @@ func TestClusterEventHandler(t *testing.T) {
 					Namespace: "test-ns",
 				},
 				Data: map[string][]byte{
-					"name": []byte("test_cluster_one"),
+					"name": []byte(testutil.Cluster),
 					"config": []byte(
 						`
 				{
@@ -120,6 +121,69 @@ func TestClusterEventHandler(t *testing.T) {
 			},
 			enqueuedGateways: make([]gatewayv1beta1.Gateway, 0),
 		},
+		{
+			name: "Not enqueued. Error parsing cluster config",
+			gateways: []gatewayv1beta1.Gateway{
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "test-gateway",
+						Namespace: "test-ns",
+						Annotations: map[string]string{
+							GatewayClusterLabelSelectorAnnotation: "type=test",
+						},
+					},
+				},
+			},
+			secret: corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						clusterSecret.CLUSTER_SECRET_LABEL: clusterSecret.CLUSTER_SECRET_LABEL_VALUE,
+					},
+					Name:      "cluster",
+					Namespace: "test-ns",
+				},
+				Data: map[string][]byte{
+					"name": []byte(testutil.Cluster),
+					"config": []byte(
+						`
+				{
+					"tlsClientConfig":
+					  {					  
+				}
+				`),
+				},
+			},
+			enqueuedGateways: make([]gatewayv1beta1.Gateway, 0),
+		},
+		{
+			name:     "Not enqueued. Error listing gateways",
+			gateways: nil,
+			secret: corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						clusterSecret.CLUSTER_SECRET_LABEL: clusterSecret.CLUSTER_SECRET_LABEL_VALUE,
+					},
+					Name:      "cluster",
+					Namespace: "test-ns",
+				},
+				Data: map[string][]byte{
+					"name": []byte(testutil.Cluster),
+					"config": []byte(
+						`
+				{
+					"tlsClientConfig":
+					  {
+					    "insecure": false,
+					    "caData": "test",
+					    "certData": "test",
+					    "keyData": "test"
+					  }
+				}
+				`),
+				},
+			},
+			enqueuedGateways: make([]gatewayv1beta1.Gateway, 0),
+		},
 	}
 
 	scheme := runtime.NewScheme()
@@ -129,13 +193,18 @@ func TestClusterEventHandler(t *testing.T) {
 
 	for _, testCase := range cases {
 		t.Run(testCase.name, func(t *testing.T) {
+
+			if testCase.gateways == nil {
+				// reset scheme to simulate client.client error
+				scheme = runtime.NewScheme()
+			}
 			client := fake.NewClientBuilder().WithScheme(scheme).WithLists(
 				&gatewayv1beta1.GatewayList{
 					Items: testCase.gateways,
 				},
 			).Build()
 
-			testQ := &testQueue{t: t}
+			testQ := &TestQueue{t: t}
 			clusterEventHandler := &ClusterEventHandler{
 				client: client,
 			}
@@ -146,12 +215,12 @@ func TestClusterEventHandler(t *testing.T) {
 	}
 }
 
-type testQueue struct {
+type TestQueue struct {
 	t                *testing.T
 	enqueuedRequests map[types.NamespacedName]bool
 }
 
-func (q *testQueue) Add(item interface{}) {
+func (q *TestQueue) Add(item interface{}) {
 	req, ok := item.(ctrl.Request)
 	if !ok {
 		q.t.Fatalf("expected enqueued item to be of type ctrl.Request, got %v", item)
@@ -164,14 +233,14 @@ func (q *testQueue) Add(item interface{}) {
 	q.enqueuedRequests[req.NamespacedName] = true
 }
 
-func (q *testQueue) MustHaveEnqueued(gateways []gatewayv1beta1.Gateway) {
+func (q *TestQueue) MustHaveEnqueued(gateways []gatewayv1beta1.Gateway) {
 	enqueuedCopy := map[types.NamespacedName]bool{}
 	for obj := range q.enqueuedRequests {
 		enqueuedCopy[obj] = true
 	}
 
 	for _, gateway := range gateways {
-		var nsn types.NamespacedName = types.NamespacedName{
+		var nsn = types.NamespacedName{
 			Namespace: gateway.Namespace,
 			Name:      gateway.Name,
 		}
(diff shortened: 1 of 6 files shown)


makslion:
/hold 
This PR now will wait for #162 and #175 and they will change a lot about the `Gateway` package and tests would need to be done over again

jasonmadigan:
At a skim, this looks OK to me - but it would be good to get another once-over from someone who's familiar with our tests generally as well @mikenairn @david-martin @philbrookes

jasonmadigan (approved):

pkg/controllers/gateway/test_helper.go
```diff
@@ -0,0 +1,212 @@
...
+		{
+			Type:  &t,
+			Value: "1.1.1.1",
+		},
+	}, nil
+}
+
+func newTestGatewayPlacer() *TestGatewayPlacer {
+	return &TestGatewayPlacer{}
+}
```
mikenairn, line 212: This file is being included as part of the pkg and not a test file, so it's skewing your coverage report. You should probably move it to a `test` package, or fgure out how you can include test helpers for unit tests.

makslion: well, in fact, it bumps coverage up right now :laughing:  but I see the point. 
Why don't we take it even a bit further - I moved all that is not the test cases into the `test` package and once integration tests are done they'd end up in the same `test` package?

test/mockInterfaces.go
```diff
@@ -0,0 +1,302 @@
...
+		if host == FailEndpointsHostname {
+			return fmt.Errorf(FailEndpointsHostname)
+		}
+	}
+	return nil
+}
+
+func NewTestHostService(client client.Client) *TestHostService {
+	return &TestHostService{controlClient: client}
+}
```
mikenairn, line 155: Why is this making calls via the client. You are using this in unit tests right? So it shouldn't be making calls to get k8s resources via that API.

There might be a need to mock this interface, but there shouldn't be this much code in it and the values that are going to be returned for each function should be explicit. Your tests could easy fail because of bugs in your test code.

test/mockInterfaces.go
```diff
@@ -0,0 +1,302 @@
...
+
+func (p *TestGatewayPlacer) GetAddresses(_ context.Context, _ *v1beta1.Gateway, _ string) ([]v1beta1.GatewayAddress, error) {
+	t := v1beta1.IPAddressType
+	return []v1beta1.GatewayAddress{
+		{
+			Type:  &t,
+			Value: "1.1.1.1",
+		},
+	}, nil
+}
```
mikenairn, line 199: This ones better, but i think it needs to be more explicit where it's being used what values are going to actually be returned if you are trying to make this into a reusable interface that any test could use where they want to mimic the placement functionality but not actually need to use it.

I had to create a similar one here https://github.com/Kuadrant/multicluster-gateway-controller/pull/237/files#diff-525abdb709ea5f8ab96976a27ec21c0e9545dd47dd82b0e2abff7a475f9672e9R40

test/helper.go
```diff
@@ -0,0 +1,277 @@
...
+}
+
+func GetValidTestRequest() ctrl.Request {
+	return ctrl.Request{
+		NamespacedName: types.NamespacedName{
+			Name:      TestDummyCRName,
+			Namespace: TestNamespace,
+		},
+	}
+}
```
mikenairn, line 65: Instead of Get, Build might be a better name for these since you are building a new instance of a resource. Get suggests you are getting it via the API.

I'd also suggest if you want these to be reusable outside of your tests that you add params to a function to let you change appropriate resource values. 

```
func BuildValidTestRequest() ctrl.Request {
	return BuildTestRequest(TestDummyCRName, TestNamespace)
}

func BuildTestRequest(name, ns string) ctrl.Request {
	return ctrl.Request{
		NamespacedName: types.NamespacedName{
			Name:      name,
			Namespace: ns,
		},
	}
}
```

test/helper.go
```diff
@@ -0,0 +1,277 @@
...
+	if programmedStatus == v1.ConditionTrue {
+		message = "gateway placed on clusters %v %s"
+	}
+	if programmedStatus == v1.ConditionFalse {
+		message = "gateway failed to be placed on all clusters %v error %s"
+	}
+	if programmedStatus == v1.ConditionUnknown {
+		message = "current state of the gateway is unknown error %s"
+	}
+	return conditions.SetCondition(gatewayStatus.Conditions, generation, string(v1beta1.GatewayConditionProgrammed), programmedStatus, string(v1beta1.GatewayReasonProgrammed), fmt.Sprintf(message, placed, errorMsg))
```
mikenairn, line 266: Why is this here? Is this not code that belongs in the pkg?

test/helper.go
```diff
@@ -0,0 +1,277 @@
...
+		if want != gatewayClassIsAccepted(class) {
+			t.Fatalf("controller ignored or not accepted gateway class")
+		}
+	}
+}
+
+func gatewayClassIsAccepted(gatewayClass *v1beta1.GatewayClass) bool {
+	acceptedCondition := conditions.GetConditionByType(gatewayClass.Status.Conditions, string(v1beta1.GatewayConditionAccepted))
+	return (acceptedCondition != nil && acceptedCondition.Status == v1.ConditionTrue)
+}
```
mikenairn, line 249: Same here, what is this in the test package for?

test/mockInterfaces.go
mikenairn: It's being created with a fake client in the unit tests, so can ignore this comment

test/helper.go
makslion: it is gone now - reusing one from the gateway package

test/helper.go
makslion: reusing one from the gateway package

test/helper.go
makslion: renamed to `build...`

test/mockInterfaces.go
```diff
@@ -0,0 +1,302 @@
...
+func (*TestQueue) Forget(item interface{}) {
+	panic("unimplemented")
+}
+
+// NumRequeues implements workqueue.RateLimitingInterface
+func (*TestQueue) NumRequeues(item interface{}) int {
+	panic("unimplemented")
+}
+
+var _ workqueue.RateLimitingInterface = &TestQueue{}
```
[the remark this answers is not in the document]
makslion: resides now in the `eventhandler_test.go`

makslion:
resolved the merge conflict and moved things around a bit. The goal is to have the `test` package hold interface implementations and functions that could be reused in other packages. The package-specific helpers now reside at the bottom of the appropriate unit test. Added build flags. 
The conversation remains about what exactly should be covered by unit and integration but this should be as close as we could get right now

pkg/controllers/gateway/cluster_eventhandler_test.go
```diff
@@ -13,24 +13,25 @@ import (
 	"k8s.io/client-go/util/workqueue"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
-	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+	v1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
```
mikenairn, line 16: Probably better to leave this to avoid potential confusion with our own local apis

pkg/controllers/gateway/gateway_controller_test.go
```diff
@@ -0,0 +1,977 @@
...
+				return
+			}
+			if !verifyTLSSecretTestResultsAsExpected(got, tt.want, tt.args.upstreamGateway) {
+				t.Errorf("reconcileTLS() \ngot: \n%v \nwant: \n%v", got, tt.want)
+			}
+		})
+	}
+}
+
+func TestGatewayReconciler_reconcileDNSEndpoints(t *testing.T) {
```
mikenairn, line 666: You need to rebase main, this no longer exists in the gateway controller.

test/helper.go
```diff
@@ -0,0 +1,120 @@
+package test
```
mikenairn, line 1: You'll probably want to rebase and make sure this still makes sense here after the e2e tests PR got merged. Also, i think you will need build tags on this

pkg/controllers/gateway/gateway_controller_test.go
makslion: And a few more tests cases not don't make much sense. Removed

pkg/controllers/gateway/cluster_eventhandler_test.go
makslion: renamed

test/helper.go
makslion: moved files around. Removed duplicate helper function. Added build tags.

pkg/controllers/gateway/params_test.go
```diff
@@ -55,6 +55,97 @@ func TestGetParams(t *testing.T) {
...
+		},
+		{
+			name: "Misconfigured ConfigMap",
+			gatewayClass: &gatewayv1beta1.GatewayClass{
+				Spec: gatewayv1beta1.GatewayClassSpec{
+					ParametersRef: &gatewayv1beta1.ParametersReference{
+						Group:     "",
+						Kind:      "ConfigMap",
+						Name:      "test-params",
+						Namespace: addr(gatewayv1beta1.Namespace("test-ns")),
```
mikenairn, line 89: I think you can get rid of the addr function and use testutil.Pointer everywhere?

test/util/helper.go
```diff
@@ -0,0 +1,118 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/gateway-api/apis/v1beta1"
+
+	"github.com/Kuadrant/multicluster-gateway-controller/pkg/apis/v1alpha1"
+)
+
+const (
+	Domain                 = "thecat.com"
+	ValidTestHostname      = "boop." + Domain
+	FailFetchDNSsubdinain  = "failfetch"
```
mikenairn, line 24: FailFetchDNSSubdomain

pkg/controllers/gateway/params_test.go
makslion: missed one! I knew those sneaky calls would get me. Done

pkg/controllers/gateway/gateway_controller_test.go
```diff
@@ -0,0 +1,821 @@
...
+					},
+				),
+				Scheme: testutil.GetValidTestScheme(),
+			},
+			args: args{
+				req: testutil.BuildValidTestRequest(testutil.DummyCRName, testutil.Namespace),
+			},
+			verify: func(res ctrl.Result, err error, t *testing.T) {
+				if !reflect.DeepEqual(res, ctrl.Result{}) &&
+					!strings.Contains(err.Error(), "gatewayckasses") &&
```
mikenairn, line 162: gatewayckasses?

pkg/controllers/gateway/gateway_controller_test.go
```diff
@@ -0,0 +1,821 @@
...
+					!strings.Contains(err.Error(), "not found") {
+					t.Errorf("expected to fail finding gateway classm but got err: %s ", err)
```
mikenairn, line 164: classm?

pkg/controllers/gateway/gateway_controller_test.go
makslion: the funny part as it was passing with the typo. Fixed

pkg/controllers/gateway/gateway_controller_test.go
makslion: fixed

PR #199, opened 2023-05-17
fix permissions issues

philbrookes:
fix permissions issues. Also removed the call to test for docker build as tests are not passing on this branch yet.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -6,6 +6,38 @@ metadata:
   name: manager-role
 rules:
 - apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - cert-manager.io
+  resources:
+  - certificates
+  verbs:
+  - create
+  - delete
+  - get
+  - list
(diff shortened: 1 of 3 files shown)


config/rbac/role.yaml
```diff
@@ -5,6 +5,42 @@ metadata:
...
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
```
R-Lawton, line 18: Do we want this cluster role to have full permissions on everything for secrets? my main thinking is the potential for  updating or patching existing secrets that it didn't create

R-Lawton: same for certs

maleck13: I think cert-manager needs that but the controller probably just needs read access as it copies the secret into a manifestwork

maleck13: the controller creates the certificate resources

philbrookes: oh good point, I just went super-permissive to get everything working and didn't look back. I'll update this now.

PR #215, opened 2023-05-24
Rename module to multicluster-gateway-controller

mikenairn:
Rename module multi-cluster-traffic-controller -> multicluster-gateway-controller

### DIFF
--- a/docs/milestone2/milestone2_application.yaml
+++ b/docs/milestone2/milestone2_application.yaml
@@ -1,13 +1,13 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: mctc-demo
+  name: mgc-demo
 ---
 apiVersion: v1
 kind: Service
 metadata:
-  name: mctc-demo
-  namespace: mctc-demo
+  name: mgc-demo
+  namespace: mgc-demo
 spec:
   ports:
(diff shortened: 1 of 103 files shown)


docs/milestone2/milestone2_application.yaml
```diff
@@ -1,39 +1,39 @@
 apiVersion: v1
```
david-martin, line 1: These milestone 2 files can be removed on main and live on on the milestone 2 tag.
The steps won't work on main regardless of the mctc/mgc change. (relies on syncer pre-ocm changes)

I'll PR that separate after merge of this if it's simpler

david-martin (commented):
/approve

docs/milestone2/milestone2_application.yaml
mikenairn: There seems to be quite a few things that need to be removed from this repo, syncer, RLP controller, outdated docs etc ... I'd rather just merge this, and do other follow up PRs for the rest if possible.

mikenairn: Follow up to remove RLP and milestone 2 docs https://github.com/Kuadrant/multicluster-gateway-controller/pull/223

PR #216, opened 2023-05-24
fix reference to gateway to ensure deletion of dns and certs

maleck13:
to verify run through the walkthrough and then delete the gateway. 

```
kubectl delete gateway prod-web -n multi-cluster-gateways

```
the dns and tls resources should also be gone

### DIFF
--- a/pkg/controllers/gateway/gateway_controller.go
+++ b/pkg/controllers/gateway/gateway_controller.go
@@ -120,33 +120,27 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	}
 	upstreamGateway := previous.DeepCopy()
 	log.V(3).Info("reconciling gateway", "classname", upstreamGateway.Spec.GatewayClassName)
-	if !controllerutil.ContainsFinalizer(upstreamGateway, GatewayFinalizer) && !isDeleting(upstreamGateway) {
+	if isDeleting(upstreamGateway) {
+		log.Info("gateway being deleted ", "gateway", upstreamGateway.Name, "namespace", upstreamGateway.Namespace)
+		if _, _, _, err := r.reconcileDownstreamFromUpstreamGateway(ctx, upstreamGateway, nil); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to reconcile downstream gateway after upstream gateway deleted: %s ", err)
+		}
+		fmt.Println("upstream post reconcile ", upstreamGateway.UID, upstreamGateway.ResourceVersion)
+		controllerutil.RemoveFinalizer(upstreamGateway, GatewayFinalizer)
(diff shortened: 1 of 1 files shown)


pkg/controllers/gateway/gateway_controller.go
```diff
@@ -120,35 +120,29 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		fmt.Println("upstream post reconcile ", upstreamGateway.UID, upstreamGateway.ResourceVersion)
+		controllerutil.RemoveFinalizer(upstreamGateway, GatewayFinalizer)
 		if err := r.Update(ctx, upstreamGateway); err != nil {
 			return ctrl.Result{}, fmt.Errorf("failed to remove finalizer from gateway : %s", err)
 		}
+		log.Info("gateway being deleted finalizer removed ", "gateway", upstreamGateway.Name, "namespace", upstreamGateway.Namespace)
 		return ctrl.Result{}, nil
 	}
 
+	if !controllerutil.ContainsFinalizer(upstreamGateway, GatewayFinalizer) && !isDeleting(upstreamGateway) {
```
mikenairn, line 137: isDeleting is redundant now, but not an issue either

docs/ocm-control-plane-walkthrough.md
```diff
@@ -223,7 +223,7 @@ spec:
     name: prod-web
     namespace: kuadrant-multi-cluster-gateways
   hostnames:
-  - "sub.replace.this"  
+  - "api.cb.hcpapps.net"
```
mikenairn, line 226: revert?

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -207,13 +201,14 @@ func (r *GatewayReconciler) reconcileDownstreamFromUpstreamGateway(ctx context.C
 		downstream.Labels = map[string]string{}
 	}
 	downstream.Labels[ManagedLabel] = "true"
-	accessor := traffic.NewGateway(downstream)
-	if isDeleting(downstream) {
+	accessor := traffic.NewGateway(upstreamGateway)
+	if isDeleting(upstreamGateway) {
```
mikenairn, line 205: How does it get here? Do we not return earlier if this is deleting?

maleck13: it calls the reconcile in the deleting block

maleck13: https://github.com/Kuadrant/multicluster-gateway-controller/pull/216/files#diff-6afa7d73841a5c0ebdb78eaf859ce5a400203a037f871440940b5347c3a97376R125

mikenairn: ignore, i read this as being further down the Reconcile function

PR #217, opened 2023-05-24
Fix local-setup on linux

mikenairn:
Make sure the counter expression never has a value of 0.

see https://unix.stackexchange.com/questions/146773/why-bash-increment-n-0n-return-error/146777

Alternatively just setting the initial value of the counters to 1 would also likely work


mikenairn:
@philbrookes lgtm and approve need to be on different lines:

/lgtm
/approve

PR #222, opened 2023-05-26
Enable RawFeedbackJsonString feature gate in OCM

### LINKED ISSUE #213: Move to using raw json response from status with OCM
Once this merges, we should try out the change and move our implementation to support it. This will allow us to get more than one address back and also awkward translations between the work status and a gateway status

https://github.com/open-cluster-management-io/work/pull/202

We want to know if this works as expected and also when it will be available in a release of OCM

david-martin:
closes #213 

Enable RawFeedbackJsonString feature gate in OCM

Pull back the full array of Gateway addresses as json, then unmarshall them into GatewayAddress objects.

This change requires an updated clusteradm cli (> 0.5.1) and an updated version of ocm (>= 0.11.0)

Note there is an issue with the 0.11.0 ocm placement image on m1 macs.
See https://kubernetes.slack.com/archives/C01GE7YSUUF/p1685016272443249 for discussion.
The workaround in the commit is to patch the clustermanager to use an amd64 build.

TODO:

- [x] Fix clusteradm version to 0.6.0 when released
- [x] Fix ocm version to 0.11.0 when released


hack/local-setup.sh
```diff
@@ -254,7 +256,7 @@ deployOCMSpoke(){
   kubectl config use-context kind-${KIND_CLUSTER_CONTROL_PLANE}
   join=$(${CLUSTERADM_BIN} get token --context kind-${KIND_CLUSTER_CONTROL_PLANE} |  grep -o  'clusteradm.*--cluster-name')
   kubectl config use-context kind-${clusterName}
-  ${BIN_DIR}/${join} kind-${clusterName} --force-internal-endpoint-lookup --context kind-${clusterName} | grep clusteradm
+  ${BIN_DIR}/${join} kind-${clusterName} --bundle-version='0.11.0' --feature-gates=RawFeedbackJsonString=true --force-internal-endpoint-lookup --context kind-${clusterName} | grep clusteradm
```
maleck13, line 259: wondering if these flags can be set for existing clusters? If not we might need to be able to fall back to the old way

maleck13 (commented):
change looks fine. But wondering if we want to hold off still?

hack/local-setup.sh
david-martin: The Klusterlet resource can be modified.
Here's an example of one where the gate is enabled

```yaml
apiVersion: operator.open-cluster-management.io/v1
kind: Klusterlet
metadata:
  name: klusterlet
spec:
  clusterName: kind-mgc-control-plane
  namespace: open-cluster-management-agent
  registrationConfiguration:
    featureGates:
    - feature: AddonManagement
      mode: Enable
  workConfiguration:
    featureGates:
    - feature: RawFeedbackJsonString
      mode: Enable
```

david-martin:
/unhold

Rebased and conflicts resolved.
@maleck13 I'd like to try land this again now.
It would be a good base for pending status aggregation changes in #275

maleck13:
Code changes look fine, just trying it out locally also

pkg/placement/placement_test.go
```diff
@@ -30,6 +31,18 @@ func init() {
 
 func TestGetAddresses(t *testing.T) {
 	testAddress := "127.0.0.1"
+	ipAddressType := v1beta1.IPAddressType
+	addresses := []v1beta1.GatewayAddress{
```
maleck13, line 35: one question here, is it worth testing with more than 1 IP? The main issue we had was around multiple addresses.

david-martin: Yes.
I'll push an update.

david-martin: Pushed.
I intend to do additional testing that hits this in the form on integration tests as part of implementing #275

maleck13:
status looks good locally:
```
statusFeedback:
        values:
        - fieldValue:
            jsonRaw: '[{"type":"IPAddress","value":"172.32.200.0"}]'
            type: JsonRaw
```

maleck13:
dns record also looks good 👍

maleck13 (approved):

PR #224, opened 2023-05-26
Pin clusteradm to 0.5.1 release and ocm bundle to 0.10.0

roivaz:
* pin the version of clusteradm to 0.5.1
* pin the version of the ocm bunlde to 0.10.0

### DIFF
--- a/hack/local-setup.sh
+++ b/hack/local-setup.sh
@@ -224,13 +224,13 @@ deployOCMHub(){
   clusterName=${1}
   echo "installing the hub cluster in kind-(${clusterName}) "
 
-  ${CLUSTERADM_BIN} init --wait --context kind-${clusterName}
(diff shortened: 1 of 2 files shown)


david-martin:
Tried out locally.
Working, with 1 exception that I've reached out on slack about (both internal slack and ocm https://kubernetes.slack.com/archives/C01GE7YSUUF/p1685016272443249)
Happy to approve the change, but won't add an lgtm until someone else with an m1 mac has tried it to rule out the issue I'm seeing

/approve

roivaz:
@david-martin pushed changes to pin clusteradm to 0.5.1 and the ocm bundle to 0.10.0. Seems to be the only combination that works currently. Can you give it a try?

R-Lawton:
im not getting any issues but only one cluster is coming up when ives asked for multiple. Are any of ye seeing that?

PR #226, opened 2023-05-26
Gh 107

philbrookes:
e2e test for DNS Health checks. Uses the fake dns provider, so the e2e test is confirmed by the presence of the healthcheck condition on the dnspolicy.

### DIFF
--- a/config/ocm/crd/appliedmanifestwork.crd.yaml
+++ b/config/ocm/crd/appliedmanifestwork.crd.yaml
@@ -1,0 +1,123 @@
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: appliedmanifestworks.work.open-cluster-management.io
+spec:
+  group: work.open-cluster-management.io
+  names:
+    kind: AppliedManifestWork
+    listKind: AppliedManifestWorkList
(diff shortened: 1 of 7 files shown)


sergioifg94:
Another alternative to test more of the health check reconciler without actually hitting AWS could be using the [mock AWS API](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/dns/aws/health_test.go#L280) that I wrote for the unit tests, instead of the fake reconciler. Although I guess that the fact that it's being used in the unit tests means it would be a bit redundant. What do you think @philbrookes?

philbrookes:
> Another alternative to test more of the health check reconciler without actually hitting AWS could be using the [mock AWS API](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/dns/aws/health_test.go#L280) that I wrote for the unit tests, instead of the fake reconciler. Although I guess that the fact that it's being used in the unit tests means it would be a bit redundant. What do you think @philbrookes?

Do you think that would be a better test than what I have here?

sergioifg94:
@philbrookes Thinking more about it I think it's good the way it is. The unit tests cover the health check reconciler and the e2e tests cover the DNSPolicy reconciler

PR #227, opened 2023-05-29
Multiple dns provider support proposal

mikenairn:
Add proposal doc for multiple dns provider support. 
Add template for proposal docs (Copied from kcp-glbc).


docs/proposals/multiple-dns-provider-support.md
```diff
@@ -0,0 +1,86 @@
...
+## Job Stories
+
+- As a developer, I want to use MGC with a domain hosted in one of the major cloud DNS providers (Google Cloud DNS, Azure DNS or AWS Route53) 
+- As a developer, I want to use  multiple domains with a single instance of MGC, each hosted on different cloud providers
+
+## Goals
+
+- Add ManagedZone and DNSRecord support for [Google Cloud DNS](https://cloud.google.com/dns/docs/)
+- Add ManagedZone and DNSRecord support for [Azure DNS](https://azure.microsoft.com/en-us/services/dns)
+- Add DNSRecord support for [CoreDNS](https://coredns.io/) (Default for development environment)
```
david-martin, line 19: What's the reason for choosing this as the default for development environment?

david-martin: Also, why not ManagedZone support here?

docs/proposals/multiple-dns-provider-support.md
```diff
@@ -0,0 +1,86 @@
...
+
+
+## Job Stories
+
+- As a developer, I want to use MGC with a domain hosted in one of the major cloud DNS providers (Google Cloud DNS, Azure DNS or AWS Route53) 
+- As a developer, I want to use  multiple domains with a single instance of MGC, each hosted on different cloud providers
+
+## Goals
+
+- Add ManagedZone and DNSRecord support for [Google Cloud DNS](https://cloud.google.com/dns/docs/)
```
david-martin, line 17: Of the 3 providers that don't currently have implementations in mgc, what priority/order would you put on them?

docs/proposals/multiple-dns-provider-support.md
```diff
@@ -0,0 +1,86 @@
...
+Add DNSProvider as an API for MGC which contains all the required config for that particular provider including the credentials. This can be thought of in a similar way to a cert manager Issuer.
+Update ManagedZone to add a reference to a DNSProvider. This will be a required field on the ManagedZone and a DNSProvider must exist before a ManagedZone can be created.
+Update all controllers load the DNSProvider directly from the ManagedZone during reconciliation loops and remove the single controller wide instance. 
+Add new provider implementations for [google](assets/multiple-dns-provider-support/google/google.md), [azure](assets/multiple-dns-provider-support/azure/azure.md) and coredns.
+    * All providers constructors should accept a single struct containing all required config for that particular provider.
+    * Providers must be configured from credentials passed in the config and not rely on environment variables.
+
+## Testing
+
+{outline any testing considerations. Does this need some form of load/performance test. Are there any considerations when thinking about an e2e test}
```
david-martin, line 57: Is there content missing here?

docs/proposals/multiple-dns-provider-support.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Other Solutions investigated
+
+Investigation was carried out into the suitability of [External DNS] (https://github.com/kubernetes-sigs/external-dns) as the sole means of managing dns resources.
+Unfortunately, while external dns does offer support for basic dns record management with a wide range of providers, there were too many features missing making it unsuitable at this time for integration.
```
david-martin, line 62: Was there consideration given to adding the missing features upstream?
What does external-dns roadmap look like and does what we want align?

docs/proposals/template.md
```diff
@@ -0,0 +1,39 @@
+# Proposal Template
```
david-martin, line 1: Any particular reason for this template?

Asking as I had a [proposal](https://github.com/Kuadrant/multicluster-gateway-controller/blob/milestone-3/docs/milestone2/status_aggregation.md) on main that wasn't in the right directory and is now gone from main.
I'm going to PR it back and wondering what the intention behind these headings is.

mikenairn: Copied from kcp-glbc https://github.com/kcp-dev/kcp-glbc/tree/main/docs/proposals

docs/proposals/multiple-dns-provider-support.md
mikenairn: Yes, and we could have forked external-dns and made things work, but decided initially to copy code into our repo and adapt so we could implement faster.

External DNS is quite focused on one part of DNS management, basic record creation, and we likely want to do significantly more in each provider than external dns is likely going to want to add. 

No idea about roadmap of external-dns, but features getting back to main are limited and not much is really added back regularly from what i can gather.

docs/proposals/multiple-dns-provider-support.md
mikenairn: guess so, although i have nothing much to add here other than we should add tests

docs/proposals/multiple-dns-provider-support.md
mikenairn: :man_shrugging:

docs/proposals/multiple-dns-provider-support.md
mikenairn: Using CoreDNS as an option is really an investigation task, will update.

Currently in order to actually do anything useful with MGC you have to have a Route53 account and your own domain registered with it, or at least a zone delegated to Route53. CoreDNS would give the option to run a CoreDNS instance locally, on the control plane cluster, and interact with it like we would do an external provider.  This could then be used locally, assuming you can add an additional nameserver on your host, to demo features. 

External DNS have a CoreDNS provider which just interacts with an etcd instance, and CoreDNS is configured with zone backed by etcd.

It might be possible to add some kind of ManagedZone support , updating CoreDNS config, but doesn't seem worth while and a single zone for development purposes would work fine.

docs/proposals/template.md
mikenairn: I don't particularly care if we keep this template or not, but i do think it would be good to keep all all proposal docs in the same place.

docs/proposals/assets/multiple-dns-provider-support/google/google.md
```diff
@@ -0,0 +1,107 @@
...
+* Weight must be an integer between 0 - 10000
+* There are no continent options for region, only finer grained regions such as us-east1, europe-west -1 etc...
+* There appears to be no way to set a default region, google just routes requests to the nearest supported region.
+* The current approach used in AWS Route53 for geo routing will work in the same way on Google DNS. A single CNAME record with geo routing policy specifying multiple geo specific A record entries as targets.
+* Geo and weighted routing can be combined, as with AWS Route53, allowing traffic with a region to be routed using weightings.
+* The modelling of the records is slightly different from AWS, but the current DNSRecord spec could still work. The google implemention of AddRecords will have to process the list of endpoints in order to group related endpoints in order to build up the required API request.
+In this case there would not be a 1:1 mapping between an endpoint in a DNSREcord and the dns provider, but the DNSRecord contents would be kept consistent accross all providers and compatability with external-dns would be maintained.
```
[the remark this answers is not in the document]
mikenairn: > Consistency would be more important than external-dns compatibility do you think? Are there changes to the DNSRecord resource you'd recommend that would make consistency easier?

Yes , external-dns compatibility isn't of high importance and shouldn't stop us modifying the spec if it make sense to us. It was just an observation, as being able to commit back to external-dns was something i had hoped we would be able to do, and not changing the Endpoints spec would likely make that easier and more likely to be accepted. 

We could make changes to the spec to allow a more provider specific set of endpoints to be defined, but whether they will ultimately be better or not I'm not sure. It might actually be better to keep it generic and have the implementation modify the endpoint list as required to make the necessary requests, and document what will be created in each provider. This especially the case with Azure that requires additional resources to be created.

I think ultimately we can decide this as we implement it and we get a better idea of what makes sense.

docs/proposals/assets/multiple-dns-provider-support/azure/azure.md
```diff
@@ -0,0 +1,56 @@
...
+## Azure
+
+https://portal.azure.com/
+
+Azure supports Weighted and Geolocation routing policies, but requires records to alias to a Traffic Manager resource that must also be created in the users acocunt https://learn.microsoft.com/en-us/azure/traffic-manager/traffic-manager-routing-methods
+
+Notes:
+
+* A Traffic Manager Profile is created per record set and is created with a routing method (Weighted or Geographic) https://portal.azure.com/#view/Microsoft_Azure_Network/LoadBalancingHubMenuBlade/~/TrafficManagers
+* Only a singe IP can be added to a DNSRecord set. A traffic manager profile must be created and aliased from a DNSRecord set for anything that involves more than a single target.
```
maleck13, line 11: record set is a strange name for something that only has 1 record 🤷

maleck13: pricing seems similar if maybe a bit cheaper for azure (was just curious) https://azure.microsoft.com/en-us/pricing/details/traffic-manager/

docs/proposals/assets/multiple-dns-provider-support/google/google.md
```diff
@@ -0,0 +1,107 @@
...
+
+https://console.cloud.google.com/net-services/dns/zones
+
+Google supports Weighted(Weighted Round Robin) and Geolocation routing policies https://cloud.google.com/dns/docs/zones/manage-routing-policies. Both of these can be configured directly on records in Google Cloud DNS and no secondary Traffic Management resource is required.
+
+Notes:
+
+* Record sets are modelled as a single endpoint with routing policy embedded. This is a different approach to Route53 where each individual A/CNAME would have its own record entry. 
+* Weight must be an integer between 0 - 10000
+* There are no continent options for region, only finer grained regions such as us-east1, europe-west-1 etc...
```
maleck13, line 11: This probably presents a problem for supported attributes on clusters? I wonder if we should we come a level and support something like `kuadrant.io/attribute-geo-region=us-east1` ? not part of this proposal but might be a more user friendly approach than continent?

docs/proposals/multiple-dns-provider-support.md
maleck13: Yeah not sure we can answer that without use feedback. I personally would prob go for google but I have no real reasoning behind that other than it looks to be a simpler integration

docs/proposals/multiple-dns-provider-support.md
maleck13: Yeah for local setup or upstream community use not requiring a cloud provider would be a pretty compelling option to get developers to try it out

docs/proposals/multiple-dns-provider-support.md
```diff
@@ -0,0 +1,82 @@
...
+
+### External DNS as a separate controller
+
+Run external dns, as intended, as a separate controller alongside mgc, and pass all responsibility for reconciling DNSRecord resources to it. All DNSRecord reconciliation is removed from MGC.
+
+Issues:
+
+* A single instance of external dns will only work with a single provider and a single set of credentials. As it is, in order to support more than a single provider, more than one external dns instance would need to be created, one for each provider/account pair.
+* Geo and Weighted routing policies are not implemented for any provider other than AWS Route53.
+* Only supports basic dns record management (A,CNAME, NS records etc ..), with no support for managed zones of health checks.
```
maleck13, line 68: This is the crux of it really. We could run an instance per provider but then would still need to code our own solution for Geo etc

docs/proposals/assets/multiple-dns-provider-support/google/google.md
david-martin: Maybe you're saying the same thing here, perhaps this can be pushed back to the user to decide.
For example, they label their ManagedClusters with a geo label.
The value of that label is what get's used for setting up geo with a provider.

docs/proposals/template.md
david-martin: Agreed.

docs/proposals/assets/multiple-dns-provider-support/azure/azure.md
mikenairn: > record set is a strange name for something that only has 1 record shrug

Yeah, not really sure why this is the case tbh, but you have to use the Traffic Manager profiles to do weighted anyway which we want as a minimum.

image

> a bit cheaper for azure (was just curious) https://azure.microsoft.com/en-us/pricing/details/traffic-manager/

I couldn't really work out what the real cost was going to be, even using their calculator, depends on queries made. I was going to keep an eye on the cost of the example geo one i setup and see what the actual cost is. Azure has quite a good screen to break down the costs of each resource.

docs/proposals/assets/multiple-dns-provider-support/google/google.md
maleck13: yeah exactly. It's a bit of a shame they have to know specifics like that but not sure I see any other way

mikenairn: I think we should just define our own and map that to a provider specific code. I guess we could optionally allow that to be overridden by additional provider specific attributes if they know what they are doing.

docs/proposals/multiple-dns-provider-support.md
mikenairn: Yeah, id agree google will be the easier of the two (not counting CoreDNS) to integrate with and probably the one to do first.

docs/proposals/assets/multiple-dns-provider-support/google/google.md
```diff
@@ -0,0 +1,107 @@
...
+* There appears to be no way to set a default region, google just routes requests to the nearest supported region.
```
maleck13, line 12: hmm so this prob affects our proposed API for DNSPolicy? no point asking for a default if only one provider supports it?

mikenairn: I think Azure might have an option:

image

I'm assuming that sets the NA region as the default, but EU traffic should go to the other one. Need to verify that.

maleck13 (approved):
looks good to me. My vote would be to work on google DNS integration first

mikenairn:
/hold

@david-martin Are you wanting to do something with that proposal template?

david-martin:
@mikenairn The CI Test Suite is failing.
This should fix it https://github.com/Kuadrant/multicluster-gateway-controller/pull/230

PR #231, opened 2023-06-02
Readd status aggregation proposal

david-martin:
Accidentally removed from main.


david-martin:
@mikenairn Just re-adding this back in the same format as before.
I'll hold off on suggestions for changing the structure of the template

PR #232, opened 2023-06-02
fixed a bug where a copy instead of an OG element was getting updated in `SetProviderSpecific`

### LINKED ISSUE #192: bug: A record weighting is incorrect
setting up two clusters we end up with the following DNS record

```
spec:
    endpoints:
    - dnsName: api.cb.hcpapps.net
      providerSpecific:
      - name: aws/weight
        value: "120"
      recordTTL: 60
      recordType: A
      setIdentifier: 172.32.200.0
      targets:
      - 172.32.200.0
    - dnsName: api.cb.hcpapps.net
      providerSpecific:
      - name: aws/weight
        value: "60"
      recordTTL: 60
      recordType: A
      setIdentifier: 172.32.201.0
      targets:
      - 172.32.201.0
    managedZone:
      name: mctc-dev-mz
```

The weight applied here by default should be the same to allow a round robin type load balancing.

jasonmadigan:
Fixes #192 

The `pair` variable in the loop here is a copy, rather than the original element.


mikenairn:
Looks fine. Wondered why this was not an issue before, but it's because we never used this function `SetProviderSpecific`. Not sure where it came from. Can we add tests for it?

PR #235, opened 2023-06-06
Enhance lookup logic for managed zones

philbrookes:
- verification: run through https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/docs/ocm-control-plane-walkthrough.md
- change the host in the gateway and the httproute to be a further nested subdomain (e.g. managedzone of pb.hcpapps.net, host and httproute of: test.sub.pb.hcpapps.net)


config/rbac/role.yaml
```diff
@@ -5,6 +5,14 @@ metadata:
...
   name: manager-role
 rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - list
+  - watch
```
mikenairn, line 15: This was removed when the RLP work was removed. Looks to have been added back since https://github.com/Kuadrant/multicluster-gateway-controller/commit/bf3c6e112a4172d696adceb49675bcfc1e6075f9#diff-6afa7d73841a5c0ebdb78eaf859ce5a400203a037f871440940b5347c3a97376R96. Seeing the same locally. We should probably have a test that runs the generate commands and checks nothing changed.

mikenairn: Added issue for a CI job to catch these https://github.com/Kuadrant/multicluster-gateway-controller/issues/236

pkg/dns/service.go
```diff
@@ -224,28 +224,38 @@ func (s *Service) SetEndpoints(ctx context.Context, addresses []gatewayv1beta1.G
...
 	}
+	return FindMatchingManagedZone(host, host, managedZones.Items)
+}
+
+func FindMatchingManagedZone(originalHost, host string, zones []v1alpha1.ManagedZone) (*v1alpha1.ManagedZone, string, error) {
+	if len(zones) == 0 {
+		return nil, "", fmt.Errorf("no managed zone found for host : %s", host)
+	}
+	host = strings.ToLower(host)
+	hostParts := strings.SplitN(host, ".", 2)
```
mikenairn, line 242: Do we need to consider tldrs that are more than 1 i.e. co.uk. It maybe doesn't matter here really, but we should probably prevent a MZ from being created with a domain that isn't at least a root domain i.e <domain name>.<top level domain>

philbrookes: good catch, fixed this now.

pkg/dns/service.go
```diff
@@ -224,28 +225,46 @@ func (s *Service) SetEndpoints(ctx context.Context, addresses []gatewayv1beta1.G
...
 
-	var managedZones v1alpha1.ManagedZoneList
+	//get the TLD from this host
+	tld, icann := publicsuffix.PublicSuffix(host)
 
-	if err := s.controlClient.List(ctx, &managedZones, client.InNamespace(t.GetNamespace()), client.MatchingFields{"spec.domainName": parentDomain}); err != nil {
-		log.FromContext(ctx).Error(err, "unable to list managed zones in traffic resource NS")
-		return nil, "", err
+	//The host is just the TLD, or the detected TLD is not an ICANN TLD
+	if host == tld || !icann {
```
mikenairn, line 252: Whats the significance of it being icann or not? Would that affect private networks using some internal dns names?

philbrookes: > Whats the significance of it being icann or not? Would that affect private networks using some internal dns names?

yeah, ICANN is a boolean that is false if the TLD is not managed by ICANN (e.g. `.phil`)

philbrookes: > > Whats the significance of it being icann or not? Would that affect private networks using some internal dns names?
> 
> yeah, ICANN is a boolean that is false if the TLD is not managed by ICANN (e.g. `.phil`)

I'm not against allowing non-ICANN domains, I just assumed it was safer to default to not allowing them.

mikenairn: Yeah I'm not sure tbh. I guess we can just go with it, but we should probably prevent the creation of ManagedZones that are using a domain that isn't icann managed in that case. Maybe a separate issue that needs discussion though. One thing i had thought was if we to implement a CoreDNS dns provider for local dev we could use a `foo.local` domain or something which a assume this wouldn't allow.

pkg/dns/service.go
```diff
@@ -224,28 +224,38 @@ func (s *Service) SetEndpoints(ctx context.Context, addresses []gatewayv1beta1.G
...
-	if len(hostParts) < 2 {
-		return nil, "", fmt.Errorf("unable to parse host : %s on traffic resource : %s", host, t.GetName())
-	}
-	subDomain := hostParts[0]
-	parentDomain := hostParts[1]
-
 	var managedZones v1alpha1.ManagedZoneList
-
-	if err := s.controlClient.List(ctx, &managedZones, client.InNamespace(t.GetNamespace()), client.MatchingFields{"spec.domainName": parentDomain}); err != nil {
+	if err := s.controlClient.List(ctx, &managedZones, client.InNamespace(t.GetNamespace())); err != nil {
```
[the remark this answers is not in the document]
mikenairn: Can be removed here also https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/controllers/gateway/suite_test.go#L139-L148

pkg/dns/service.go
philbrookes: ok, I'll allow non-ICANN domains for now, it's trivial to turn it back on if required.

pkg/dns/service.go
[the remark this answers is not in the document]
philbrookes: oh thanks, done.

PR #237, opened 2023-06-07
DNSPolicy manages all DNS resources for a target Gateway

### LINKED ISSUE #221: Move all DNS related responsibilities to DNSPolicy reconciler
**What**

Move all DNS logic into DNSPolicy reconciliation.

Currently the gateway controller reconciliation will create and update DNSRecords. The DNSPolicy targets a Gateway and it's reconciliation will get existing DNSRecords and update them to add heath check annotations. This is two things manipulating the same resource. Considering that we want to expand the DNSPolicy to include additional DNS routing which is going to directly influence how the DNSRecord.Endpoints are configured it would make more sense that all DNS operations were carried out in a single reconciliation.

* Update DNSPolicy reconciler to create/update/delete DNSRecords and pull the target information from the targeted Gateway. The DNSPolicy controller is the only thing that updates the DNSRecord spec based on the target Gateway status and DNSPolicy spec.
* Create a default DNSPolicy if none already exists targeting the current Gateway during Gateway reconciliation.
* Set target Gateway as the owner of any created DNSRecords. This allows the DNSPolicy to be deleted without losing connectivity. Gateway deletion will continue to remove owned DNSRecords.
* Trigger DNSPolicy reconciliation on changes to it's target Gateway.

mikenairn:
Closes #221

Moves all DNS related resource management (DNSRecord, DNS Health checks etc ...), out of the Gateway reconciliation and into the DNSPolicy reconciliation. With this change it is now required for a DNSPolicy to be created that targets the Gateway before any DNSRecord resources will be created. To avoid unnecessary extra steps for a user, a default DNSPolicy will be created for each Gateway.

DNSPolicy is updated to use the same pattern for policy attachment and lifecycle as other kuadrant policies [RateLimitPolicy](https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/ratelimitpolicy_controller.go) and [AuthPolicy](https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/authpolicy_controller.go), moving us towards more consistency across the two streams of work.

Kuadrant polices maintain an annotation on the target resource (Always Gateway in our case), and this can be used to build up a list of Gateways that need resources created/updated(Gateway created/updated) for them, or resources removed (Target ref changed, Gateway deleted etc..). It should cover the following:

* DNSPolicy is created. Triggers policy reconcile, and creates DNSRecords for Gateway listeners with appropriate ManagedZones available (Same logic as always) with the target Gateway set as the owner. Creates HealthChecks based on the policy config, these belong to the policy.
* Gateway targeted by a policy is deleted. Triggers policy reconcile, and deletes resources created and owned by the policy (HealthChecks only, DNSRecords belong to target). 
* DNSPolicy is deleted. Triggers policy reconcile, and deletes resources created and owned by the policy (HealthChecks only, DNSRecords belong to target). Deletes the dnspolicy annotation from the target gateway allowing it be targeted by another policy.
* DNSPolicy target ref changes. Triggers policy reconcile, follows the same steps for DNSPolicy deletion for the old target, and follows the DNSPolicy create steps for the new target.

Move SetEndpoints to  provider (Prep for multiple dns provider implementations).

ToDo

- [x] Create default DNSPolicy on initial Gateway discovery
- [x] Update policy status
- [x] Add/ Update all tests


pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -20,37 +20,51 @@ import (
...
+}
+
 // DNSPolicyReconciler reconciles a DNSPolicy object
 type DNSPolicyReconciler struct {
-	Client client.Client
-
+	reconcilers.TargetRefReconciler
 	DNSProvider dns.Provider
 	HostService gateway.HostService
+	Placement   gateway.GatewayPlacer
```
guicassolato, line 54: I see the use in the Gateway reconciler. Is adding to the DNSPolicy reconciler preparation for a next step maybe?

pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -61,26 +75,130 @@ func (r *DNSPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		return err
+	}
+
+	// set direct back ref - i.e. claim the target network object as taken asap
+	if err := r.ReconcileTargetBackReference(ctx, client.ObjectKeyFromObject(dnsPolicy), targetNetworkObject, DNSPoliciesBackRefAnnotation); err != nil {
+		return err
+	}
+
+	// set annotation of policies affecting the gateway - should be the last step, only when all the reconciliation steps succeed
+	return r.ReconcileGatewayPolicyReferences(ctx, dnsPolicy, gatewayDiffObj)
```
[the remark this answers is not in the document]
guicassolato: Does anybody know if the annotations are included in the resource version and therefore covered by Kubernetes concurrency control when modifying objects?

pkg/controllers/dnspolicy/dnspolicy_controller.go
mikenairn: It's used in the DNSPolicy to determine what what clusters a Gateway is actually placed on (OCM thing for multi cluster support). When we create a DNSRecord and add endpoints to it, we only use gateway addresses for gateways actually placed on workload clusters.

It would be nice though if this reconciler could work in a multi cluster (OCM) and singe cluster (non OCM) fashion. The whole DNSPolicy, much like RateLimitPolicy and AuthPolicy, could be effuse in it's own right outside of an OCM environment.

pkg/controllers/dnspolicy/dnspolicy_controller.go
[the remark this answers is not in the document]
mikenairn: So ... from what i understand here, we maintain a reference, currently as an annotation on the target resource (always Gateway in our case) with the id of the current policy. We need to do this in order to cleanup resources when a target ref is changed, and `ReconcileGatewayPolicyReferences` is ensuring that the policy annotations are removed from any old Gateway targets.

If there was some issue with updating annotations on resources concurrently the kuadrant operator would have that same issue right?

Has there been thoughts/discussions around changing how this works in kuadrant?

pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -61,26 +75,130 @@ func (r *DNSPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
-		Watches(&source.Kind{
-			Type: &v1alpha1.DNSRecord{},
-		}, &DNSRecordEventHandler{
-			client:      r.Client,
-			hostService: r.HostService,
-		}).
+		Watches(
+			&source.Kind{Type: &gatewayapiv1beta1.Gateway{}},
+			handler.EnqueueRequestsFromMapFunc(gatewayEventMapper.MapToDNSPolicy),
+		).
```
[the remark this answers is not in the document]
sergioifg94: Agreed, that makes sense

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
```diff
@@ -22,91 +26,104 @@ type healthChecksConfig struct {
...
+		if err != nil {
+			return allResults, err
 		}
 
 		allResults = append(allResults, results...)
 
-		if err = r.updateDNSRecord(ctx, dnsRecordOriginal, dnsRecord); err != nil {
-			return err
+		if !equality.Semantic.DeepEqual(dnsRecordOriginal, dnsRecord) {
+			err = r.Client().Update(ctx, dnsRecord)
```
sergioifg94, line 79: When the reconciliation of the health checks was done in a different controller than the reconciliation of the DNS records it made sense that as a result of reconciling the health checks we would update the affected DNS records, but now that this is done in the same reconciliation loop I wonder if we should refactor it so that the `reconcileDNSRecords` method behaviour is affected by the reconciliation of the health checks.
The current approach still works but it feels weird that in the same reconciliation loop we are creating/updating the DNSRecords and then immediately (potentially) updating them again. Wdyt?

mikenairn: Agreed, i went back and forth with this a few times. I wanted to have the dns records reconcile, return the records and then reconcile the health checks, and the last thing that happened was a single update to the record.

I went back on that for some reason, i think it was to keep the changes in this PR simpler initially, but i might go back and see if we can do it that way.

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -133,6 +134,10 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	}
 
 	if !controllerutil.ContainsFinalizer(upstreamGateway, GatewayFinalizer) {
+		// The first time we see the gateway we want to make sure a DNSPolicy exists for it
```
eguzki, line 137: wouldn't be easier and more robust to check for DNSPolicy back references in the gateway and when there is none, then create a default DNS policy?

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -180,9 +185,31 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	return ctrl.Result{}, reconcileErr
 }
 
+func (r *GatewayReconciler) reconcileDefaultDNSPolicy(ctx context.Context, upstreamGateway *gatewayv1beta1.Gateway) error {
```
eguzki, line 188: nitpick: This method is not doing a reconciliation. It is (conditionally) creating an object. For me reconciliation is basically:
```
desiredObject = GetDesiredObjectFromCRandClusterInfo()
existingObject = readFromCluster()
if existingObject == nil {
   createObject(desiredObject)
} else {
   if desired != existing {
      existing = desired
       updateObject(existing)
  }
}
```

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -180,9 +185,31 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	log := crlog.FromContext(ctx)
+
+	defaultDNSPolicy := policy.NewDefaultDNSPolicy(upstreamGateway)
+
+	var dnsPolicies v1alpha1.DNSPolicyList
+	err := r.List(ctx, &dnsPolicies, client.InNamespace(upstreamGateway.GetNamespace()), client.MatchingFields{policy.POLICY_TARGET_REF_KEY: policy.GetTargetRefValueFromPolicy(&defaultDNSPolicy)})
```
eguzki, line 194: If I am not wrong, this operation is doing: "read all the dns policies that are targeting this gateway". 
I think that you do not need to do this if you rely on the annotations in the gateway which are back references to the DNS policies. In other words, Just check gateway annotation for a back references.

eguzki (commented):
The PR is looking good.

pkg/apis/v1alpha1/dnspolicy_types.go
```diff
@@ -64,6 +67,35 @@ type DNSPolicy struct {
...
+func (p *DNSPolicy) Validate() error {
+	if p.Spec.TargetRef.Group != ("gateway.networking.k8s.io") {
+		return fmt.Errorf("invalid targetRef.Group %s. The only supported group is gateway.networking.k8s.io", p.Spec.TargetRef.Group)
+	}
+
+	if p.Spec.TargetRef.Kind != ("Gateway") {
+		return fmt.Errorf("invalid targetRef.Kind %s. The only supported kind is Gateway", p.Spec.TargetRef.Kind)
+	}
+
+	if p.Spec.TargetRef.Namespace != nil && string(*p.Spec.TargetRef.Namespace) != p.Namespace {
```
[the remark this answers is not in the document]
mikenairn: Yes, at the moment we want to keep these as Direct Policy Attachement, and one must exist in teh same NS as the Gateway.

pkg/controllers/gateway/gateway_controller.go
mikenairn: Yes, this is listing the polices in the current namespace that are targeting this gateway. Whether or not we can use the annotations instead relates to this question https://github.com/Kuadrant/multicluster-gateway-controller/pull/237#discussion_r1226412200 I'm not really seeing how we can reliably, but maybe I'm wrong.

pkg/controllers/dnspolicy/dnspolicy_controller.go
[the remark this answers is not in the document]
mikenairn: Added requeue on status update failure due to conflict

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
mikenairn: As discussed, going to leave this as it is for now. We will revisit how we are doing this when we finalize what we are doing with health checks.

pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -61,26 +75,130 @@ func (r *DNSPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+
+	// add finalizer to the dnsPolicy
+	if !controllerutil.ContainsFinalizer(dnsPolicy, DNSPolicyFinalizer) {
+		if err := r.AddFinalizer(ctx, dnsPolicy, DNSPolicyFinalizer); client.IgnoreNotFound(err) != nil {
+			return ctrl.Result{Requeue: true}, err
+		}
+	}
+
+	err = r.reconcileResources(ctx, dnsPolicy, targetNetworkObject)
+	if err != nil {
```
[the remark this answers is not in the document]
mikenairn: Added ready condition for policy and not update status regardless of error during reconciliation. Further enhancements required here to deal with health check status, and making sure we always process the status (Missing target for instance still skips the status update) but this will cover most issues that can happen during reconciliation.

pkg/controllers/gateway/gateway_controller.go
mikenairn: renamed

pkg/controllers/gateway/gateway_controller.go
eguzki: The gateway can be targeted by a single policy. If you implement based on back references as I suggest, deleting the default policy will trigger a new policy to be created. Users would have to update the default policy instead of deleting existing default one and create a new one. 

If you want otherwise, then LGTM.

pkg/controllers/gateway/gateway_controller.go
eguzki: I think that this operation is only needed for an edge case: First you create some DNS policy targeting a gateway that does not exist. Then you create the gateway. In that scenario, the gateway would not have any finalizer annotation and `ensureDefaultDNSPolicy` method would be called. Then, this operation would detect that policy and prevent the creation of the default dns policy. 

For the normal use case, when the gateway has been created and it does not have the kuadrant's finalizer, the default DNS policy needs to be created. Then the user needs to delete it and create his own one. 

I do not expect this edge situation to happen often, but yeah, who knows. LGTM

sergioifg94:
I'm verifying the PR and I see that the default DNSPolicy gets created automatically, but when the HTTPRoute is created afterwards, the DNSPolicy is not reconciled so the DNSRecord is not created. I had to manually edit the DNSPolicy to trigger a reconcile loop and then they did get created. Maybe we need an extra event handler to make sure it gets reconciled automatically?

pkg/controllers/gateway/gateway_controller.go
mikenairn: Yes, so when i discussed this with @maleck13  we didn't, for reasons i can't remember, want them to have to update the default one we were creating. It really depends what we actually want to happen here, have us make sure a policy always exist and users just update that one, or create one initially and allow users to delete that and create their own.

I'll leave it as is for now, we can always change it later.

mikenairn:
> I'm verifying the PR and I see that the default DNSPolicy gets created automatically, but when the HTTPRoute is created afterwards, the DNSPolicy is not reconciled so the DNSRecord is not created. I had to manually edit the DNSPolicy to trigger a reconcile loop and then they did get created. Maybe we need an extra event handler to make sure it gets reconciled automatically?

Hmm ok, what steps are you doing so i can make sure we have a test for it?

mikenairn:
/hold

The DNSPolicy reconcile isn't getting triggered on route attachment. Seems like we relied on manifest work and placement events in the gateway controller to do this. Ideally the dnspolicy controller would not need to concern itself with this, and instead rely on appropriate updates of the gateway to trigger a reconcile.

mikenairn:
/hold cancel

The issue with route attachment was caused by a bug in the policy reconciler, it was updating a single annotation, instead of having one for both the direct ref and all back refs https://github.com/Kuadrant/multicluster-gateway-controller/pull/237/commits/ebfea0a2951ac5831319545613e181ac26d01d0b. This was causing it to try and parse a string as json, which fails, and doesn't queue the policy on gateway change.

pkg/controllers/gateway/gateway_controller.go
maleck13: The reason for not forcing the use of the existing policy was to allow users to create a new policy resource and store it in git and use something like argo to sync state. 
The user flows here are:
1) Update the existing policy 
2) create a new policy and then delete the existing policy

pkg/_internal/policy/policy.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func NewDefaultDNSPolicy(gateway *gatewayv1beta1.Gateway) v1alpha1.DNSPolicy {
+	gatewayTypedNamespace := gatewayv1beta1.Namespace(gateway.Namespace)
+	return v1alpha1.DNSPolicy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      gateway.Name,
+			Namespace: gateway.Namespace,
+		},
+		Spec: v1alpha1.DNSPolicySpec{
```
maleck13, line 34: When we add support for GEO and Custom Weighting, we might want a default of something like 

```
loadbalancing:
   strategy: "RoundRobin"
```

pkg/controllers/dnspolicy/dnspolicy_controller.go
maleck13: We have an issue to aggregate the status of spoke gateways back to the hub gateway. The addresses will be part of that so we can prob disconnect this from the placement logic when that is in place

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/Kuadrant/multicluster-gateway-controller/pkg/apis/v1alpha1"
+	"github.com/Kuadrant/multicluster-gateway-controller/pkg/traffic"
+)
+
+func (r *DNSPolicyReconciler) reconcileDNSRecords(ctx context.Context, dnsPolicy *v1alpha1.DNSPolicy, gwDiffObj *reconcilers.GatewayDiff) error {
+	log := crlog.FromContext(ctx)
+
+	for _, gw := range gwDiffObj.GatewaysWithInvalidPolicyRef {
+		log.V(1).Info("reconcileDNSRecords: gateway with invalid policy ref", "key", gw.Key())
+		//ToDo Since gateways own DNSRecords I don't think there is anything to do here?
```
maleck13, line 23: should we remove the policy ref or is that responsibility somewhere else?

maleck13: I assume if there is no policy then a new default is created?

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	// Reconcile DNSRecords for each gateway directly referred by the policy (existing and new)
+	for _, gw := range append(gwDiffObj.GatewaysWithValidPolicyRef, gwDiffObj.GatewaysMissingPolicyRef...) {
```
maleck13, line 27: why do we reconcile dns records when ```GatewaysMissingPolicyRef``` not quite getting that? is that the current policy being reconciled used to target another gateway and so we will apply this policy to that gateway and the new target? Would we not expect that now there is no valid policy a new default for that gateway would be created?

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
```diff
@@ -22,91 +26,108 @@ type healthChecksConfig struct {
...
 	}
-	if len(dnsRecords) == 0 {
-		log.Info("No DNSRecord found matching policy", "policy", policy)
-		return nil
+
+	healthCheckConfig := getHealthChecksConfig(dnsPolicy)
+	allResults := []dns.HealthCheckResult{}
+
+	// Reconcile Health checks for each gateway directly referred by the policy (existing and new)
+	for _, gw := range append(gwDiffObj.GatewaysWithValidPolicyRef, gwDiffObj.GatewaysMissingPolicyRef...) {
```
maleck13, line 45: so again not sure I am following this, a policy can only target one resource, so if the target changes is not the case that we should delete the associated health checks? I am probably misunderstanding what is happening here

pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -61,26 +80,158 @@ func (r *DNSPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 
+func (r *DNSPolicyReconciler) reconcileResources(ctx context.Context, dnsPolicy *v1alpha1.DNSPolicy, targetNetworkObject client.Object) error {
+	// validate
+	err := dnsPolicy.Validate()
+	if err != nil {
+		return err
+	}
+
+	// reconcile based on gateway diffs
+	gatewayDiffObj, err := r.ComputeGatewayDiffs(ctx, dnsPolicy, targetNetworkObject, &DNSPolicyRefsConfig{})
```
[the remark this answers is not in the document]
eguzki: I can answer that. The purpose is to manage the resources attached to the gateway directly like Istio WASMPlugins, AuthorizationPolicies or in this particular case DNSRecords and Healthchecks. 

According to the "Golang Doc" from https://github.com/Kuadrant/kuadrant-operator/blob/41281635dfe5f3890457dc52b6fb60c31d834f53/pkg/reconcilers/targetref_reconciler.go#L174

```
// Returns:                                                                      
// * list of gateways to which the policy applies for the first time             
// * list of gateways to which the policy no longer applies                      
// * list of gateways to which the policy still applies                         
```

As the target references from the DNSPolicy can change, a dns policy can be targeting gateway A at t0 and later, gateway B at t1. The controller needs to clean up resources (DNS Records and healthchecks) from gateway A and create new resources (DNS Records and healthchecks) for the gateway B. How does the controller know which gateways should have the DNSPolicy configuration on those DNS Records and healthchecks and which gateways should not have? We designed an annotation system where the gateways have (back) references to the DNSPolicies targeting them. That annotation system allows the controller to know that a DNSPolicy D *was* targeting gateway A and now it is targeting a different gateway B, so the controller knows what to do to on each gateway.

Arguably the `ComputeGatewayDiffs` could be refactored to return only the "list of gateways to which the policy no longer applies" and "list of gateways to which the policy applies"

pkg/_internal/policy/policy.go
mikenairn: Yes, we will add a default here when we have more than one strategy that we support

pkg/dns/service.go
```diff
@@ -60,12 +59,13 @@ func (s *Service) resolveIPS(ctx context.Context, addresses []gatewayv1beta1.Gat
...
 
 func (s *Service) GetManagedHosts(ctx context.Context, traffic traffic.Interface) ([]v1alpha1.ManagedHost, error) {
-	managed := []v1alpha1.ManagedHost{}
+	var managed []v1alpha1.ManagedHost
 	for _, host := range traffic.GetHosts() {
-		managedZone, subDomain, err := s.GetManagedZoneForHost(ctx, host, traffic)
-		if err != nil {
-			return nil, err
-		}
+		//ToDo Should just continue on "no managed zone found for host" error
```
maleck13, line 64: is this todo still relevant?

maleck13 (commented):
changes look good, just some clarification on what we want to do with the diffs

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
mikenairn: > should we remove the policy ref or is that responsibility somewhere else?

Policy refs are removed elsewhere at the end of the policy reconcile. This one is focused on dnsrecord resources that are created. If we wanted records to be removed when a policy no longer applies, we would do it here.

> I assume if there is no policy then a new default is created?

The dnspolicy reconciler will not create policies. We currently have the gateway controller creating a default one, but if it is removed it would need to be manually created again re. https://github.com/Kuadrant/multicluster-gateway-controller/pull/237#discussion_r1224293911

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
mikenairn: This might explain it better https://github.com/Kuadrant/multicluster-gateway-controller/pull/237#discussion_r1232097666. GatewaysWithValidPolicyRef and GatewaysMissingPolicyRef both mean the policy applies. For GatewaysMissingPolicyRef, the policy ref will be added later on in the reconcile.

pkg/dns/service.go
mikenairn: Fixing this at the moment, relates to the failing tests

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -226,6 +252,10 @@ func (r *GatewayReconciler) reconcileDownstreamFromUpstreamGateway(ctx context.C
 	if err != nil {
 		return false, metav1.ConditionFalse, clusters, fmt.Errorf("failed to get managed hosts : %s", err)
 	}
+	if len(managedHosts) == 0 {
+		return false, metav1.ConditionFalse, clusters, fmt.Errorf("no managed hosts found")
+	}
```
mikenairn, line 257: The "Programmed" status is currently being set to false only when we don't have a valid ManagedZone for any hosts and does not check if DNS is actually setup. This is true on main and this PR.

My understanding based on https://github.com/Kuadrant/multicluster-gateway-controller/pull/251#discussion_r1229393328, and external conversations is that we would expect that to remain false until at least on managed host is found for a listener and a DNSRecord created and ready. Since that's not happening on main, I'm not fixing it in this PR but will add a follow on issue for it.

mikenairn: @philbrookes @maleck13 fyi

pkg/dns/service.go
```diff
@@ -60,10 +63,10 @@ func (s *Service) resolveIPS(ctx context.Context, addresses []gatewayv1beta1.Gat
 }
 
 func (s *Service) GetManagedHosts(ctx context.Context, traffic traffic.Interface) ([]v1alpha1.ManagedHost, error) {
-	managed := []v1alpha1.ManagedHost{}
+	var managed []v1alpha1.ManagedHost
 	for _, host := range traffic.GetHosts() {
 		managedZone, subDomain, err := s.GetManagedZoneForHost(ctx, host, traffic)
-		if err != nil {
+		if err != nil && !errors.Is(err, ErrNoManagedZoneForHost) {
```
mikenairn, line 69: We need to ignore these "no managed zone for host" errors or it will only ever check the first one

maleck13 (approved):

PR #238, opened 2023-06-07
make verify-manifests

### LINKED ISSUE #236: Add verify-manifests make target and CI job
**What**

Add a verify-manifests make target than executes `make manifests` and returns a non zero exit code if anything chnages in the config directory. Run this is part of our PR CI jobs.

Take inspiration from https://github.com/Kuadrant/kuadrant-operator/blob/main/make/verify.mk#L7 https://github.com/Kuadrant/kuadrant-operator/blob/main/.github/workflows/test.yaml#L79

**Why**

Catch issue like https://github.com/Kuadrant/multicluster-gateway-controller/pull/235#discussion_r1221034170

jasonmadigan:
- adds a new `verify-manifests` target to the makefile
- added this to our GH Action CI

fixes #236

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -33,6 +33,20 @@ jobs:
           only-new-issues: true
           args: --timeout=5m
           skip-go-installation: true
+  verify-manifests:
+    name: Verify manifests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Go 1.19.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.19.x
+        id: go
+      - name: Check out code
+        uses: actions/checkout@v2
+      - name: Run make verify-manifests
+        run: |
+          make verify-manifests
   test_suite:
     name: Test Suite
     runs-on: ubuntu-20.04
--- a/hack/make/check.make
+++ b/hack/make/check.make
@@ -3,3 +3,8 @@
 .PHONY: verify-code
 verify-code: vet lint ## Verify code formatting
 	@diff -u <(echo -n) <(gofmt -d `find . -type f -name '*.go' -not -path "./vendor/*"`)
+
+.PHONY: verify-manifests
+verify-manifests: manifests ## Verify manifests update.
+	git diff --exit-code ./config
+	[ -z "$$(git ls-files --other --exclude-standard --directory --no-empty-directory ./config)" ]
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -88,11 +88,11 @@ ifndef ignore-not-found
 endif
 
 .PHONY: install
-install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
+install: verify-manifests manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
```
jasonmadigan, line 91: perhaps not needed, given the action will run - but it may be better to warn about this in multiple targets?

jasonmadigan:
/hold

Until latest `config/rbac/role.yaml` checked in

.github/workflows/ci.yaml
```diff
@@ -33,6 +33,20 @@ jobs:
...
           args: --timeout=5m
           skip-go-installation: true
+  verify-manifests:
+    name: Verify manifests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Go 1.18.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.18.x
```
mikenairn, line 43: Probbaly want to use v1.19

.github/workflows/ci.yaml
```diff
@@ -33,6 +33,20 @@ jobs:
           only-new-issues: true
           args: --timeout=5m
           skip-go-installation: true
+  verify-manifests:
+    name: Verify manifests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Go 1.18.x
```
mikenairn, line 40: 1.19

Makefile
mikenairn: Yeah, this would fail locally until you committed the changes in which would be annoying

mikenairn:
/lgtm
/approve

I'm not sure if the failing test will stop it merging with the current setup, hopefully it does, but the hold will prevent it merging anyway.

mikenairn:
We cold just commit this change as part of this PR https://github.com/Kuadrant/multicluster-gateway-controller/pull/235/files#diff-6afa7d73841a5c0ebdb78eaf859ce5a400203a037f871440940b5347c3a97376L95.

jasonmadigan:
Might be worth adding that just to test the CI then passes.

PR #239, opened 2023-06-07
OCM gateway controller suite tests

R-Lawton:
### Verification
1. Run `make test`
Tests should pass


david-martin:
👀

pkg/controllers/gateway/suite_test.go
```diff
@@ -50,11 +55,15 @@ import (
...
-	cancel    context.CancelFunc
+	cfg              *rest.Config
+	k8sClient        client.Client
+	testEnv          *envtest.Environment
+	ctx              context.Context
+	cancel           context.CancelFunc
+	nsSpoke1Name     = "test-spoke-cluster-1"
+	nsSpoke2Name     = "test-spoke-cluster-2"
+	defaultNS        = "default"
+	gatewayFinalizer = "kuadrant.io/gateway"
```
david-martin, line 66: Can move these to const block

pkg/controllers/gateway/suite_test.go
```diff
@@ -35,10 +35,15 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
-	clusterv1beta2 "open-cluster-management.io/api/cluster/v1beta1"
```
david-martin, line 38: I'm guessing the naming of the var here was an oversight (beta2 vs beta1), hence your change to fix it as ocmclusterv1beta1?

pkg/controllers/gateway/suite_test.go
```diff
@@ -262,234 +275,349 @@ var _ = Describe("GatewayClassController", func() {
...
+			}
+			Expect(k8sClient.Status().Update(ctx, placementDecision)).To(BeNil())
+			//Before: Stub Gateway for tests but not creating it here
+			hostname := gatewayv1beta1.Hostname("test.example.com")
+			gateway = &gatewayv1beta1.Gateway{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-gw-1",
+					Namespace: defaultNS,
+					Labels: map[string]string{
+						"cluster.open-cluster-management.io/placement": "true",
```
david-martin, line 357: The test works as is with the value of 'true' for this label,
but I'm wondering if a more meaning value would be better in the test?
Something like 'GatewayControllerTest'? (can be anything)
The value of 'true' might give the impression that it can only be true or false, which is not the case i.e. it's an arbitrary value that must match between the Gateway and the PlacementDecision

david-martin (changes requested):
Good to have this core test in place.
I've left a couple of small suggestions, nothing major.

pkg/controllers/gateway/suite_test.go
R-Lawton: Yeah exactly make it a bit more readable to

pkg/controllers/gateway/suite_test.go
```diff
@@ -35,10 +35,15 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
-	clusterv1beta2 "open-cluster-management.io/api/cluster/v1beta1"
-	workv1 "open-cluster-management.io/api/work/v1"
+	"encoding/json"
```
mikenairn, line 38: nit: import grouping `"encoding/json"` should be in the first group

config/ocm/crd/manifestwork.yaml
```diff
@@ -0,0 +1,558 @@
...
+    served: true
+    storage: true
+    subresources:
+      status: {}
+status:
+  acceptedNames:
+    kind: ""
+    plural: ""
+  conditions: []
+  storedVersions: []
```
mikenairn, line 558: What's the reason for the duplicate crds ? https://github.com/Kuadrant/multicluster-gateway-controller/blob/15f25d153ac2f5826552e4c78ef4b1b3b130aecf/config/ocm/crd/manifestwork.crd.yaml, or are they different somehow?

R-Lawton: Thats a mistake, I've removed that and others that were duplicates i think were created from rebasing the ocm into main

mikenairn:
Looks fine to me. 
 
I think we might want to consider the structure of our tests a bit, they don't need to all be bundled into that one test_suite file and i can imagine we might want to add more specs in there as we go. For the gateway controller we could at the very least split them into a gateway_controller_spec.go and  gatewayclass_controller_spec.go, ginkgo will deal with loading them all into the test suite for that pkg.

Might also be worth considering a controllers_test pkg that actually integrates all the controllers. The way we have this setup currently wants you to write test suites per controller which is kind of counter what an integration test should be doing.

david-martin:
closes #201

PR #240, opened 2023-06-08
Provider agnostic DNS health checks proposal

sergioifg94:
[Proposal doc](https://github.com/sergioifg94/multi-cluster-traffic-controller/blob/proposal-health-checks/docs/proposals/provider-agnostic-dns-health-checks.md)

### DIFF
--- a/docs/proposals/provider-agnostic-dns-health-checks.md
+++ b/docs/proposals/provider-agnostic-dns-health-checks.md
@@ -1,0 +1,77 @@
+# Provider agnostic DNS Health checks
+
+## Introduction
+
+The MGC has the ability of extending the DNS configuration of the gateway with the DNSPolicy resource. This resource allows users to configure health checks. As a result of configuring health checks, the controller creates the health checks in Route53, attaching them to the related DNS records. This has the benefit of automatically disabling an endpoint if it becomes unhealthy, and enabling it again when it becomes healthy again.
+This feature has a few shortfalls:
+
+1. It’s tightly coupled with Route53. If other DNS providers are supported they must either provide a similar feature, or health checks will not be supported
+2. Lacks the ability to reach endpoints in private networks
+
+This document describes a proposal to extend the current health check implementation to overcome these shortfalls.
+
+### Goals
+
+* Ability to configure health checks in the DNSPolicy associated to a Gateway
+* Ability for the health checks to reach endpoints in private networks
+* DNS records are disabled when the associated health check fails
+* Current status of the defined health checks is visible to the end user
+* ~~Transparently keep support for other health check providers like Route53~~
+
+## Proposal
+
+#### `DNSPolicy` resource
+
+The presence of the `healthCheck` field in the DNSPolicy will affect the reconciliation
+of the endpoints when creating/updating DNSRecords:
+
+* For every endpoint that is generated, a health check is created based on the configuration in the DNSPolicy
+* Relevant `providerSpecific` fields are added to the endpoint:
+  * A reference to the equivalent `DNSHealthCheckProbe` in the downstream cluster
+  will be set: `health-probe/name`, `health-probe/namespace`
+  * The health status of the `DNSHealthCheckProbe` will be propagated
+
+A `failureThreshold` field will be added to the health spec, allowing users
+to configure how many consecutive health check failures are observed before
+taking the unhealthy endpoint down
+
+#### `DNSHealthCheckProbe` resource
+
+The DNSHealthCheckProbe resource configures a health probe in the controller to perform the health checks against a local endpoint.
+
+```yaml
+apiVersion: kuadrant.io/v1alpha1
+kind: DNSHealthCheckProbe
+metadata:
+  name: example-probe
+spec:
+  port: "..."
+  host: “...”
+  ipAddress: "..."
+  path: "..."
+  protocol: "..."
+  interval: "..."
+status:
+  healthy: true
+  consecutiveFailures: 0
+  reason: ""
+  lastCheck: "..."
+```
+
+The reconciliation of this resource results in the configuration of a health probe,
+which targets the endpoint and updates the status. The status is propagated to the providerSpecific status of the equivalent endpoint in the DNSRecord
+
+### Changes to current controllers
+
+In order to support this new feature, the following changes in the behaviour of the controllers are proposed.
+
+#### DNSPolicy controller
+
+Currently the reconciliation loop of this controller is in charge of creating health checks in the configured DNS provider (Route53 currently) based on the spec of the DNSPolicy, separately from the reconciliation of the DNSRecords. The proposed change is to reconcile health checks as the DNSRecords are created, using the health check
+configuration to alter the behaviour of the DNSRecord reconciliation.
+
+Instead of Route53 health checks, the controller will create `DNSHealthCheckProbe` resources
+
+#### DNSRecord controller
+
+Currently the reconciliation loop of this controller updates the DNS records using the aws/health-check-id provider specific field, that is updated as a result of the DNSPolicy reconciliation. The proposed change is to extend the functionality of acting on provider specific fields, to disable/enable the DNS record if the health-probe/healthy field is Unhealthy. This can be achieved by either deleting/re-creating the DNS record, or setting the weight to 0 in order to disable it


docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,101 @@
...
+The reconciliation logic of this resource broadly follows:
+
+1. Select the health check reconciler implementation based on the value of the provider field
+1. Delegate the reconciliation on the health check reconciler, passing the configuration and the provider specific values
+1. Update the status and the providerSpecific status values from the reconcile result
+1. The provider specific status values are propagated to the related DNSRecord
+
+### Data plane
+
+#### `DNSHealthCheckProbe` resource
```
sergioifg94, line 70: Something I haven't decided yet is whether to make the agent a controller that watches the DNSHealthCheckProbe resource, or a separate system that is managed by the DNSHealthCheckProbe controller

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,116 @@
...
+    type: DNSPolicy
+    name: example-policy
+spec:  
+  config:
+    port: "..."
+    path: "..."
+    protocol: "..."
+    host: "..."
+    ipAddress: "..."
+  provider: "aws | agent | ..."
```
mikenairn, line 44: Why aws here? I don't think a HealthCheck should really be tied to a provider, or needs to be, in the same way a DNSRecord isn't.
It looks to me like we have 2 types of health check, provider or agent, so do we not just want a type in the spec that is one of "provider" or "agent"?

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,116 @@
...
+  providerSpecific:
+    "agent/managed-cluster": "..."
+status:
+  providerSpecific:
+    "agent/health-status": "Healthy"
+    "agent/last-check": "..."
+    "aws/health-check-id": "..."
+```
```
mikenairn, line 52: Mentioned this on the original doc, but do we want to keep the spec of this consistent with DNSRecord to which it is tied quite tightly?  

My understanding of what you are proposing is that we create a DNSHealthCheck resource for every endpoint in a DNSRecord, but we could also create a DNSHealthCheck that contains a list of endpoints the same as the DNSRecord.

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,116 @@
...
+  ipAddress: "..."
+  path: "..."
+  protocol: "..."
+  interval: "..."
+status:
+  status: Healthy
+  lastCheck: "..."
+```
+
+The reconciliation of this resource results in the deployment of the health check agent, which performs health checks following the configuration in the spec, and updates the status. The status is propagated upstream to the providerSpecific status of the equivalent DNSHealthCheck
```
mikenairn, line 104: Would the agent not already have been placed on the workload cluster prior to this? Do you mean that the control plane will ensure it's deployed on cluster when we first know we want to add an agent health probe?

docs/proposals/provider-agnostic-dns-health-checks.md
mikenairn: I would expect it should be a controller reconciling DNSHealthCheckProbes, but might be missing something?

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,116 @@
...
+The DNSHealthCheck resource represents the configuration of a health check. The reconciliation of this resource results in the creation of health checks in the selected DNS provider, or, if configured to use the agent, the creation of a health check probe in the downstream cluster
+
+```yaml
+apiVersion: kuadrant.io/v1alpha1
+kind: DNSHealthCheck
+metadata:
+  name: example-dnshealthcheck
+  ownerReference:
+    type: DNSPolicy
+    name: example-policy
```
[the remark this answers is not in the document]
sergioifg94: That makes me think that maybe a simpler solution would be to not have a separate CRD for the health checks, but add it as a section to the endpoint in the DNSRecord. The configuration would be set automatically from the reconciliation of the DNSPolicy, as it already happens. What do you think?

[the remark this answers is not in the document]
philbrookes: Does this mean that each endpoint would then need it's own status block to report heath? This might lead to a very verbose DNSRecord CR?

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,86 @@
...
+
+Add a field in the health check section of the DNSPolicy to select the health
+check type to use
+
+```yaml
+healthCheck:
+  # ...
+  port: "..."
+  # ...
+  type: "Provider | Agent"
```
david-martin, line 33: Just a thought here. 
The provider vs. the agent are implementations of health checks.
From a high level functionality point of view they are doing the same thing.
However, there is a subtle difference that makes me wonder if the type of health check should be based on the location of the health check.
For example, `internal` vs `external`. Or `private` vs `public`.
If you're using mgc with route53 you can define both types of health checks.
If not, only internal/private is allowed

maleck13: how much does the end user care about this detail I wonder? I want a health check and I want it to remove unhealthy endpoints / alert me when things are going wrong. Not sure how much I care where that health check is?

maleck13: I am wondering basically if we have an agent based approach, is there any advantage to using the cloud providers health checks?
They cost more and work in less scenerios

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+The presence of the `healthCheck` field in the DNSPolicy will affect the reconciliation
+of the endpoints when creating/updating DNSRecords:
+
+* For every endpoint that is generated, a health check is created based on the configuration in the DNSPolicy
+* Relevant `providerSpecific` fields are added to the endpoint depending on the
```
maleck13, line 40: I think it is worth a debate whether we want to use provider specific health checks or not. If we are happy with the agent based approach, it give us a lot more flexibility

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,86 @@
...
+  # ...
+  port: "..."
+  # ...
+  type: "Provider | Agent"
+```
+
+The presence of the `healthCheck` field in the DNSPolicy will affect the reconciliation
+of the endpoints when creating/updating DNSRecords:
+
+* For every endpoint that is generated, a health check is created based on the configuration in the DNSPolicy
```
maleck13, line 39: I think we should only create the health check if there is a DNSRecord present for the listener (IE the gateway listener has a HTTPRoute attached. I don't think we should just create one as it will naturally fail if there are no routes attached

docs/proposals/provider-agnostic-dns-health-checks.md
maleck13: I think a controller that reconciles the resource and pushes the checks into a queue and sets up a worker to process that check every n seconds. If the check changes the queue is flushed and the worker recreated to perform the check based on the new time WDYT

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,86 @@
...
+  host: “...”
+  ipAddress: "..."
+  path: "..."
+  protocol: "..."
+  interval: "..."
+status:
+  status: Healthy
+  lastCheck: "..."
+```
+
```
maleck13, line 72: With using an agent on the spoke cluster, what is our policy if that cluster becomes unreachable or cannot reach the hub for some reason?
This would manifest itself as the DNSHealthCheckProbe not being updated and perhaps the managed cluster status changing. 
My default here would be to do nothing with DNS in this case. That said it could be a valid failure (IE the cluster has failed for some reason or become unreachable ) one option would be have secondary check from the hub that is triggered if the health check is not updated within a reasonable period and the managed cluster is marked unavailable

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,86 @@
...
+The reconciliation of this resource results in the configuration of the health check agent, which targets the endpoint and updates the status. The status is propagated upstream to the providerSpecific status of the equivalent endpoint in the DNSRecord
+
+### Changes to current controllers
+
+In order to support this new feature, the following changes in the behaviour of the controllers are proposed.
+
+#### DNSPolicy controller
+
+Currently the reconciliation loop of this controller is in charge of creating health checks in the configured DNS provider (Route53 currently) based on the spec of the DNSPolicy, separately from the reconciliation of the DNSRecords. The proposed change is to reconcile health checks as the DNSRecords are created, using the health check
+configuration to alter the behaviour of the DNSRecord reconciliation.
```
maleck13, line 82: I am not getting if the DNSHealthCheckProbe needs to be synced to the spokes or if it stays in the control plane and is read by the agents on the spokes?

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,86 @@
...
+
+#### DNSRecord controller
+
+Currently the reconciliation loop of this controller updates the DNS records using the aws/health-check-id provider specific field, that is updated as a result of the DNSPolicy reconciliation. The proposed change is to extend the functionality of acting on provider specific fields, to disable/enable the DNS record if the agent/health-status field is Unhealthy. This can be achieved by either deleting/re-creating the DNS record, or setting the weight to 0 in order to disable it
```
maleck13, line 86: I like the idea of setting the weight to 0 is that works, that way we are not actually removing the record just stopping it from being returned.

docs/proposals/provider-agnostic-dns-health-checks.md
sergioifg94: Yes, I think the best thing is to move the logic that creates the health checks to create them as the DNSRecords are being created, that will avoid a situation like that

docs/proposals/provider-agnostic-dns-health-checks.md
sergioifg94: One idea that I've considered is to install the agent as an add-on in the spoke clusters, which I think has some sort of health check feature as well. I'm not familiar enough with it to be sure

docs/proposals/provider-agnostic-dns-health-checks.md
sergioifg94: It would be syncted to the spokes, and reconciled by the agent that runs in the spokes, then the status would be synced back to the control plane

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,86 @@
...
+
+```yaml
+apiVersion: kuadrant.io/v1alpha1
+kind: DNSHealthCheckProbe
+metadata:
+  name: example-probe
+spec:
+  port: "..."
+  host: “...”
+  ipAddress: "..."
```
maleck13, line 64: could this possibly just be set to Address?

sergioifg94: > and have one property for host and address?

Do you mean that the requests are made to `http(s)://<address>:<port><path>` instead of `http(s)://<ipAddress>:<port><path>` with the `Host: <host>` header

docs/proposals/provider-agnostic-dns-health-checks.md
maleck13: also I think we should only create health checks for the "final record" not sure there is value in creating it for every end point

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,77 @@
...
+The presence of the `healthCheck` field in the DNSPolicy will affect the reconciliation
+of the endpoints when creating/updating DNSRecords:
+
+* For every endpoint that is generated, a health check is created based on the configuration in the DNSPolicy
+* Relevant `providerSpecific` fields are added to the endpoint:
+  * A reference to the equivalent `DNSHealthCheckProbe` in the downstream cluster
+  will be set: `health-probe/name`, `health-probe/namespace`
+  * The health status of the `DNSHealthCheckProbe` will be propagated
+
+A `failureThreshold` field will be added to the health spec, allowing users
```
maleck13, line 34: @philbrookes I wonder should we have a successThreshold? IE when do we consider an endpoint healthy again? one 200 response is likely not enough to assume it is back to ready??

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,77 @@
...
+#### `DNSPolicy` resource
+
+The presence of the `healthCheck` field in the DNSPolicy will affect the reconciliation
+of the endpoints when creating/updating DNSRecords:
+
+* For every endpoint that is generated, a health check is created based on the configuration in the DNSPolicy
+* Relevant `providerSpecific` fields are added to the endpoint:
+  * A reference to the equivalent `DNSHealthCheckProbe` in the downstream cluster
+  will be set: `health-probe/name`, `health-probe/namespace`
+  * The health status of the `DNSHealthCheckProbe` will be propagated
```
mikenairn, line 32: If you setup a relationship of a `DNSRecord` owns `DNSHealthCheckProbes` that the dnspolicy is creating, do you need to set this `providerSpecific` field? 

In order to keep the DNSPolicy reconciler clean it would be better if we did not have to add these values. Doing this, you either have to do what we currently do and pull all the records again after having just created/updated them, or combine the creation/deletion of `DNSHealthCheckProbes` resources into the `dnspolicy_dnsrecord` reconcile.

Would it be a better solution to just lookup healthchecks owned by the current dnsrecord during the dnsrecord reconciliation and build up a list of hosts/ips and their current health status and pass that into the [MultiClusterGatewayTarget](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go#L111), which will ultimately use that when deciding the weight for the record with that host/ip https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/dns/service.go#L234

philbrookes:
Moved the proposal to my own PR for easier modification: https://github.com/Kuadrant/multicluster-gateway-controller/pull/319

PR #243, opened 2023-06-09
e2e tests

roivaz:
This PR adds an e2e test suite that can be run against the local-setup environment. At the moment it contains just some basic smoke tests that focus on checking stuff around the gateway resource and ensure the happy path scenario works properly (dns resolution and a 200 ok request to a service behind a gateway). The tests just check the system as a black box, without checking any of the "inner workings".
I created a small package with utils for the e2e tests in general. It was originally intended to make the tests run both in the HCG unstable cluster and with the local-setup, so I added some logic that configures the test environment and that makes sure the tests cleane up after themselves as much as possible. This is less relevant now that they will only run in local-setup, but still it's nice as avoids from having to recreate the environment constantly to clean it up when things fail and you need to debug.

## How to test

* To run the tests in local-setup with a single cluster that has both hub and spoke

```bash
export TEST_HUB_NAMESPACE=multi-cluster-gateways
export TEST_MANAGED_ZONE=<your-zone>
export TEST_HUB_KUBE_CONTEXT=kind-mgc-control-plane
export OCM_SINGLE=1
```
```bash
make local-setup
```
```bash
make test-e2e
```

* To run the tests in local-setup with a hub cluster and a dedicated workload cluster

```bash
export TEST_HUB_NAMESPACE=multi-cluster-gateways
export TEST_MANAGED_ZONE=<your-zone>
export TEST_HUB_KUBE_CONTEXT=kind-mgc-control-plane
export TEST_SPOKE_KUBE_CONTEXT_PREFIX=kind-mgc-workload
export MGC_WORKLOAD_CLUSTERS_COUNT=1

```
```bash
make local-setup
```
```bash
make test-e2e
```


pkg/controllers/gateway/gateway_controller.go
```diff
@@ -91,7 +91,7 @@ type GatewayPlacer interface {
 // +kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gateways/finalizers,verbs=update
 // +kubebuilder:rbac:groups=work.open-cluster-management.io,resources=manifestworks,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=placementdecisions,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch
+// +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;delete
```
roivaz, line 94: I catch a bug here, the permission to cleanup secrets was missing.

sergioifg94:
Not sure if it's related with the changes in this PR but when verifying the tests with separate hub and workload clusters the local-setup was failing consistently in [here](https://github.com/Kuadrant/multicluster-gateway-controller/blob/b7e6b3c0ed3f95eab066c38bf14f2863a57515c1/hack/local-setup.sh#L61). I doubled the timeout and it worked. Other than that everything works as expected

test/smoke/gateway_single_spoke_test.go
```diff
@@ -0,0 +1,267 @@
+package smoke
```
sergioifg94, line 1:
```suggestion
//go:build e2e

package smoke
```

I think it won't make a difference practically but just to follow the convention introduced in #244 I'd suggest to add the tag to separate these from the integration tests

sergioifg94 (commented):
Tests work as expected, just the one suggestion

test/smoke/gateway_single_spoke_test.go
mikenairn, line 1: Why "smoke" and not "e2e"? What's the intention of these tests, to execute and e2e suite with different scenarios in a test cluster only, or run smoke tests against any cluster in a non destructive way to check the cluster is working?

roivaz: The original intent was to run them against any cluster, when the plan was having a unstable and stable clusters running the service. When that changed I reworked the tests to work against the local-setup and never changed the name. But now that you mention it, it probably doesn't make sense anymore and we might just call them e2e.

mikenairn: We should probably just rename the directory and package to `e2e` i.e. `test/e2e` ?

maleck13:
@mikenairn @sergioifg94 ok to give this another review? Would like to get it closed out

maleck13:
changes look reasonable to me and have been tried by @sergioifg94. @roivaz lets create a follow up to start executing these during a PR

maleck13 (approved):

PR #244, opened 2023-06-09
Add test build tags (unit and integration)

mikenairn:
Add unit and integration buid tags to all test files and adds 2 new make targets to run the unit (make test-unit) and integration (make test-integration) tests. Running the integration tests will run it with the ginkgo verbose flag which fails when you try and use it on normal unit test files.


sergioifg94:
Verified both `make test-unit` and `make test-integration`
/lgtm

PR #251, opened 2023-06-14
adding managed zone integration tests

philbrookes:
tests:
- managed zone can be created and deleted
- managed zone will enable a gateway using that host to be programmed
- managed zone will not enable a gateway using a different host
- managed zone will not enable a gateway in a different namespace, using it's host

### DIFF
--- a/pkg/controllers/managedzone/suite_test.go
+++ b/pkg/controllers/managedzone/suite_test.go
@@ -1,0 +1,185 @@
+/*
+Copyright 2023 The MultiCluster Traffic Controller Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package managedzone
+
+import (
+	"context"
+	"path/filepath"
+	"testing"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+
+	certman "github.com/jetstack/cert-manager/pkg/apis/certmanager/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	clusterv1beta2 "open-cluster-management.io/api/cluster/v1beta1"
+	workv1 "open-cluster-management.io/api/work/v1"
+	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+
+	"github.com/Kuadrant/multicluster-gateway-controller/pkg/apis/v1alpha1"
+	"github.com/Kuadrant/multicluster-gateway-controller/pkg/dns"
+	//+kubebuilder:scaffold:imports
+)
+
+// These tests use Ginkgo (BDD-style Go testing framework). Refer to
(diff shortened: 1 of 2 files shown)


R-Lawton:
👀

pkg/controllers/managedzone/suite_test.go
```diff
@@ -0,0 +1,443 @@
...
+	dns := dns.NewService(k8sManager.GetClient(), dns.NewSafeHostResolver(dns.NewDefaultHostResolver()), dnsProvider)
+	plc := placement.NewOCMPlacer(k8sManager.GetClient())
+	err = (&gateway.GatewayReconciler{
+		Client:       k8sManager.GetClient(),
+		Scheme:       k8sManager.GetScheme(),
+		Certificates: certificates,
+		Host:         dns,
+		Placement:    plc,
+	}).SetupWithManager(k8sManager, ctx)
+	Expect(err).ToNot(HaveOccurred())
```
mikenairn, line 146: Do we even need the Gateway and Gateway Class controllers? ManagedZones are essentially a completely separate API with no dependency on gateways.

I do think an integration test should start them all together, but with these pkg level suite_tests i've been conflicted, but that's probably because i don;t think we should have them.

philbrookes: if I didn't start this, I couldn't prove that gateways in different namespaces or different hosts didn't get programmed, the status was permanently empty.

philbrookes: as you say, we should aim to put these all in one package eventually which would resolve this issue then, when all controllers are running.

david-martin: We solved this kind of problem in the gateway controller integration test with mocking.
e.g. to get the gateway controller reconciler to moved past a certain point, cert manager had to be running and do it's thing. So instead of starting cert manager (since we're not testing that piece) we created a secret that looks like what cert-manager would create.

In this case, you should update the Gateway status directly to get the managedzone controller to continue past whatever point it gets stuck

pkg/controllers/managedzone/suite_test.go
```diff
@@ -0,0 +1,443 @@
...
+				}
+
+				gwIsProgrammed := false
+				for _, condition := range createdGW.Status.Conditions {
+					if condition.Type == "Programmed" && condition.Status == metav1.ConditionTrue {
+						gwIsProgrammed = true
+					}
+				}
+
+				return mzIsReady && gwIsProgrammed
```
mikenairn, line 291: I'm not sure why, but it seems odd to me that the existence of a manage zone has any influence on whether a gateway is programmed or not. Is this only true if we have at least one host that has a corresponding managed zone that we can create a record in?

pkg/controllers/managedzone/suite_test.go
mikenairn: > if I didn't start this, I couldn't prove that gateways in different namespaces or different hosts didn't get programmed, as the gateway's status was permanently empty.

True, but that test would make more sense to me in the gateway_controller test suite since it's that controller doing that

pkg/controllers/managedzone/suite_test.go
philbrookes: yes, exactly.

pkg/controllers/managedzone/suite_test.go
philbrookes: This is interesting. I don't think mocking would resolve as I am actually testing what happened in the gateway reconciler as a result of the managed zone, it's not so much that it got stuck. 

Maybe Mike is right here, and this should be moved to the gateway controller test suite... Though I do then wonder if the managed zone controller might need to be started in that test.

I suppose all of these questions are really answered by our plan to amalgamate all of these tests into one package.

philbrookes: I will move these tests to the gateway test suite for now.

philbrookes: done this now. It does make a lot more sense this way.

pkg/controllers/managedzone/suite_test.go
```diff
@@ -0,0 +1,192 @@
...
+		Scheme:      k8sManager.GetScheme(),
+		DNSProvider: dnsProvider,
+	}).SetupWithManager(k8sManager)
+	Expect(err).ToNot(HaveOccurred())
+
+	err = (&gateway.GatewayClassReconciler{
+		Client: k8sManager.GetClient(),
+		Scheme: k8sManager.GetScheme(),
+	}).SetupWithManager(k8sManager)
+	Expect(err).ToNot(HaveOccurred())
```
mikenairn, line 130: Is this needed

mikenairn:
Looks fine to me, just one question

pkg/controllers/managedzone/suite_test.go
philbrookes: no, overlooked removing this one! I'll remove it now.

philbrookes: done

PR #252, opened 2023-06-14
Add verify-code to CI workflow

mikenairn:


pkg/controllers/gateway/suite_test.go
```diff
@@ -17,10 +17,10 @@ package gateway
 
 import (
 	"context"
+	"encoding/json"
 	"path/filepath"
 	"testing"
 	"time"
-	"encoding/json"
```
mikenairn, line 23: @R-Lawton fyi

hack/make/check.make
```diff
@@ -1,7 +1,7 @@
 ##@ Check
 ## Targets to be used to ensure code quality
 .PHONY: verify-code
-verify-code: vet lint ## Verify code formatting
+verify-code: vet ## Verify code formatting
```
mikenairn, line 4: We should probably make this lint target work without having to have the binary pre-installed, but we also don;t need to run the linter as part of this target on CI anyway since we already have a test for that.

philbrookes (approved):
/lgtm

pkg/controllers/gateway/suite_test.go
R-Lawton: it has to be in alphabetical order?

mikenairn: ha yeah, but running `go fmt` will order them for you.

PR #254, opened 2023-06-14
remove replace.this nonsense with env-vars

jasonmadigan:
For some reason, I thought I'd done this before as part of a recent change. Anyway, doing this again. To get things working, setup an env-var like:

```
export MGC_SUB_DOMAIN=myapp.test.hcpapps.net # replace this
```

### DIFF
--- a/docs/how-to/ocm-control-plane-walkthrough.md
+++ b/docs/how-to/ocm-control-plane-walkthrough.md
@@ -14,21 +14,19 @@ We will start with a single cluster and move to multiple clusters to illustrate 
 
 >**Note:** :exclamation: this walkthrough will setup a zone in your AWS account and make changes to it for DNS purposes
 
->**Note:** :exclamation: `replace.this` is a place holder that you will need to replace with your own domain
-
 ## Installation and Setup
-1. Clone this repo locally 
-1. Setup a `./controller-config.env` file in the root of the repo with the following key values
+* Clone this repo locally 
+* Setup a `./controller-config.env` file in the root of the repo with the following key values
 
     ```bash
     # this sets up your default managed zone
     AWS_DNS_PUBLIC_ZONE_ID=<AWS ZONE ID>
     # this is the domain at the root of your zone (foo.example.com)
-    ZONE_ROOT_DOMAIN=<replace.this>
+    ZONE_ROOT_DOMAIN=test.hcpapps.net # replace this with your root domain
     LOG_LEVEL=1
     ```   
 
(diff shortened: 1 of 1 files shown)


docs/how-to/ocm-control-plane-walkthrough.md
```diff
@@ -171,7 +173,7 @@ Open three windows, which we'll refer to throughout this walkthrough as:
         namespaces:
           from: All
       name: api
-      hostname: sub.replace.this
+      hostname: $MCG_SUB_DOMAIN
```
david-martin, line 176:
```suggestion
      hostname: $MGC_SUB_DOMAIN
```

docs/how-to/ocm-control-plane-walkthrough.md
```diff
@@ -249,7 +251,7 @@ We only configure DNS once a HTTPRoute has been attached to a listener in the ga
         name: prod-web
         namespace: kuadrant-multi-cluster-gateways
       hostnames:
-      - "sub.replace.this"  
+      - "$MCG_SUB_DOMAIN"
```
david-martin, line 254:
```suggestion
      - "$MGC_SUB_DOMAIN"  
```

docs/how-to/ocm-control-plane-walkthrough.md
```diff
@@ -37,6 +35,12 @@ We will start with a single cluster and move to multiple clusters to illustrate
     AWS_REGION=eu-west-1
     ```
 
+* Add some env-vars to your `zsh.rc` or `.bash_profile`
```
david-martin, line 38: It might be that I've run through the walkthrough so many times, but I tend to use a different app host (same subdomain) each time.
So I would probably just export it in the session as opposed to setting it statically in rc or profile. e.g. `export MGC_SUB_DOMAIN=ocmtest01.dm.hcpapps.net`

However, I'd have to export it in each terminal if that was the case. 
I guess that's why you're suggesting this approach?

docs/how-to/ocm-control-plane-walkthrough.md
jasonmadigan: nicely spotted

docs/how-to/ocm-control-plane-walkthrough.md
jasonmadigan: yeah, agreed - I'll make that clearer, I do something similar to you (override often, but have a default)

jasonmadigan: Updated, tried to make this a little clearer

PR #255, opened 2023-06-14
Remove extra placement and manifest work watches from gateway controller

mikenairn:
These are the same:
* https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/controllers/gateway/gateway_controller.go#L423-L459 
* https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/controllers/gateway/gateway_controller.go#L463-L498

### DIFF
--- a/pkg/controllers/gateway/gateway_controller.go
+++ b/pkg/controllers/gateway/gateway_controller.go
@@ -460,42 +460,6 @@ func (r *GatewayReconciler) SetupWithManager(mgr ctrl.Manager, ctx context.Conte
 		Watches(&source.Kind{
 			Type: &corev1.Secret{},
 		}, &ClusterEventHandler{client: r.Client}).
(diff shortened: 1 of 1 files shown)


jasonmadigan (approved):
/lgtm
/approved

david-martin:
Good find @mikenairn 

There's a test for the gateway controller that [exercises the ManifestWork watch](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/controllers/gateway/suite_test.go#L579-L594).
That test is still passing, so 👍 

The PlacementDecision watch is also kind of exercised, though theres no part of the test that specifically relies on it.
The mock PlacementDecision is [created upfront](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/controllers/gateway/suite_test.go#L339-L351)
Could be a candidate to add to the test more directly so the watch is relied on?

mikenairn:
Thanks @jasonmadigan and @david-martin

PR #264, opened 2023-06-19
remove syncer component

philbrookes:
- Removed all syncer related code
- Updated walkthrough to deploy controller as a container
  - Validates permissions
  - Requires less terminals
# verification
- See tests passing
- Run through [OCM walkthrough](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/docs/how-to/ocm-control-plane-walkthrough.md)

# Eulogy
```
There was a young syncer from KCP
who always synced things quite happily
He worked hard no doubt
but KCP didn't work out
so the young syncer was executed rapidly
```


hack/register-cluster.sh
```diff
@@ -1,11 +1,5 @@
 #!/bin/bash
 
-### Usage
```
maleck13, line 3: do we still need this file?

Issue #138, opened 2023-03-27
Allow a grace period strategy when a Gateway is displaced from a cluster


david-martin:
If the placment rules or decisions change for a Gateway it may require the deletion of a Gateway on 1 or more data plane clusters.
If that happens, traffic may still resolve the Gateway IP for the displaced gateway and send traffic to it for some time after DNS records have been updated.

A grace period should be allowed for, before the Gateway is displaced to ensure the traffic dropped is kept low.

Depends on #52

maleck13:
I wonder could this be part of the DNSPolicy "DNSGracePeriod" where we have a sane default of 2x TTL

philbrookes:
PR: https://github.com/Kuadrant/multicluster-gateway-controller/pull/288

PR #270, opened 2023-06-21
DNSPolicy proposal

maleck13:
While most of this is understood by the team and has been discussed in various forums, we did not have anything capturing the details. This doc attempts to bring the design and info together and to ensure we all understand the why and the what of DNSPolicy


maleck13:
@mikenairn  I would like your review in particular please

docs/proposals/DNSPolicy.md
```diff
@@ -0,0 +1,390 @@
...
+     default: 10  #required if weighted is chosen
+     custom:
+     - attribute: cloud  #optional with both geo and weighted
+       value: AWS
+       weight: 10
+     - attribute: cloud
+       value: GCP
+       weight: 20
+    geo:
+      default: IE # required if geo is chosen
```
mikenairn, line 64: It makes sense to make it more obvious that both geo and weight are not mutually exclusive and weights still apply for geo strategy. Since geo is still weighted, should setting a default for weighted be required, or will we just add a default weight if none is specified? If setting a default weight is not required, loadbalancing should probably be optional.

docs/proposals/DNSPolicy.md
```diff
@@ -0,0 +1,409 @@
...
+    name: gateway-name
+  health:
+   ...
+  ttl: 
+    value: 60
+    gracePeriod: 240 #(seconds to wait beyond TTL before allowing deletion of gateway) 
+  loadBalancing:
+     #one or both of geo and weighted are required 
+    weighted:
+     default: 10  #required if weighted
```
mikenairn, line 89: since everything is always weighted, should this just be required?

docs/proposals/DNSPolicy.md
maleck13: Yeah still debating this one, I like having it there as it can then be used to inform other custom weights in the future. If it is hidden, then you either have to set a new one or figure out what the default we used was

docs/proposals/DNSPolicy.md
```diff
@@ -0,0 +1,409 @@
...
+kind: DNSPolicy
+name: default-round-robin
+spec:
+  targetRef: # defaults to gateway gvk and currrent namespace
+    name: gateway-name
+  health:
+   ...
+  ttl: 
+    value: 60
+    gracePeriod: 240 #(seconds to wait beyond TTL before allowing deletion of gateway)
```
mikenairn, line 85: presumably this is a delay on how long we wait before removing an endpoint from the endpoint list rather than deleting a gateway?

docs/proposals/DNSPolicy.md
```diff
@@ -0,0 +1,409 @@
...
+  - attribute: cloud 
+    value: AWS
+    weight: 20
+  - attribute: compute
+    value: GC
+    weight: 30
+```
+
+With the above defined managed cluster, this weighting would give us two values. 10 and 20. Proposal is in this case is that we cannot pick any weight and so the DNSPolicy is invalid and a status condition should be set. In this case we would not change any DNS records.
+An alternative in this case would be to pick the default weight, but update the status to ensure it is clear this has been due to the conflict.
```
mikenairn, line 258: I think the alternative is a better option personally, or we use the the smallest weight value of all the possible ones.

docs/proposals/DNSPolicy.md
philbrookes: As I have understood this, the concept is that this is how long we will wait after the endpoint is removed from the DNSRecord, to remove the gateway from the cluster (i.e. allow time for the removed DNS Record endpoint to propagate).

docs/proposals/DNSPolicy.md
philbrookes: I agree with Mike on this. I think making it possible for DNS Records to enter an "invalid" state reduces our ability to perform one of our main goals (pointing DNS Records at the correct clusters).

docs/proposals/DNSPolicy.md
mikenairn: ok, i see. How are we going to block the deletion of a gateway based on something the dns policy is doing? I read the text below, but I'm still not understanding how it works since the dns policy won't directly affect a gateway other than maybe updating a status.

philbrookes: yeah, that's what I'm trying to figure out right now.

mikenairn: It seems arguably nothing to do with dns. You want to delay the deletion of the target by x amount, once dns is no longer enabled for that gateway. The only way i can you see you doing it is having the dns policy put a finalizer on the Gateway and I'm not sure if that is really what we want is it?

david-martin: Related to #138 (which I see assigned to @philbrookes )
If we agree on the grace period being solely related to keeping a gateway running for latent traffic, perhaps we remove it from DNSPolicy spec and figure out the right solution in #138?

It's more than gateway deletion that would need to be taken into account, so a finalizer may not work. 
If a gateway is 'moved' via placementdecision changes, that grace period should come into effect. 
Owning the finalizer of an individual gateway in a spoke cluster from the hub would have architectural considerations.

maleck13: It is related to DNS by the TTL but I see where you are coming from. So I guess the question is where does it belong?
As this only matters when the placement changes, and an existing cluster is removed from the placement decision, perhaps for now we look to add an annotation to the gateway itself? It may migrate into a policy at somepoint

docs/proposals/DNSPolicy.md
maleck13: Yeah I think I agree with this

docs/proposals/DNSPolicy.md
maleck13: Yes I think we go with, you have to set a default and if the policy is invalid / conflicted we fall back to the default and ensure it is clear what has happened in the status. 
Perhaps in a future iteration we handle this via a validating webhook?

david-martin: An alternative spec to avoid conflict would be to limit the attribute to 1 fixed name.
e.g. `kuadrant.io/lb-attribute-weight`
In the DNSPolicy, the value is the sole differentiator e.g.
```yaml
  - value: cloud-aws
    weight: 20
  - value: compute-gc
    weight: 30
```
It much more restrictive, and forces the user to think about the attribute more, but does avoid conflicts.

docs/proposals/DNSPolicy.md
mikenairn: > It is related to DNS by the TTL

It's related to placement from what i understand, but wants to base a decision on some DNS config(TTL) . It doesn't directly have any influence on what the DNS reconciler will do.

The Gateway controller (our ocm one) can depend on a DNSPolicy and read from it if it wants, but "gracePeriod" for placement decisions probably still doesn't belong in it. The TTL itself can be read from the DNSRecord directly, since these still belong to a Gateway.

philbrookes: so I think this sounds like we're discussing a gateway "DeschedulingPolicy" right? Where we want to say something like "if the DNS Record still exists for any gateway then onDeschedule of a particular gateway wait for the DNS Records TTL + some value in the policy"? Are we thinking then, that I will first implement this in code, and possibly later we'll pull it out to an actual policy, or shall I look into making the DeschedulingPolicy now?

docs/proposals/DNSPolicy.md
maleck13: going with default is required for weight

docs/proposals/DNSPolicy.md
maleck13: decision: use a code based default for now and allow it to be overriden on the gateway with an experimental annotation `kuadrant.io/experimental-grace-period`

docs/proposals/DNSPolicy.md
maleck13: I think for the usecases we have the single attribute is enough. It allows different values on different clusters and allowed different weighting for different gateways

maleck13: it simplifies the implementation and is easier  harder to get wrong from an end user perspective

maleck13: ```kuadrant.io/lb-attribute-custom-weight: "GCP"```

docs/proposals/DNSPolicy.md
```diff
@@ -0,0 +1,292 @@
...
+   ...   
+```  
+
+In order to provide GEO based DNS and allow customisation of the weighting, we need some additional information to be provided by the gateway / cluster admin about where this gateway has been placed. For example if they want to use GEO based DNS as a strategy, we need to know what GEO identifier(s) to use for each record we create and a default GEO to use as a catch all. Also if the desired load balancing approach is to provide custom weighting and no longer simply use Round Robin, we will need a way to identify which records to apply that custom weighting to based on the clusters the gateway is placed on.
+
+To solve this we will allow two new attributes to be added to the `ManagedCluster` resource as annotations:
+
+```
+   kuadrant.io/lb-attribute-GEO-<GEO-code-type>: "IE"
+   kuadrant.io/lb-attribute-custom-weight: "GCP"
```
maleck13, line 96: perhaps these should be labels?

docs/proposals/DNSRecordStructure.md
```diff
@@ -0,0 +1,175 @@
...
+      targets:
+        - aws.lb.com
+    - dnsName: 1bc1.lb-1ab1.www.example.com
+      recordTTL: 60
+      recordType: A
+      targets:
+        - 192.22.2.1
+```
+
+TODO
```
maleck13, line 124: remove

docs/proposals/DNSPolicy.md
```diff
@@ -0,0 +1,280 @@
...
+Custom weighting will use the associated `custom-weight` attribute set on the `ManagedCluster` to decide which records should get a specific weight. The value of this attribute is up to the end user.
+
+example:
+
+```yaml
+apiVersion: cluster.open-cluster-management.io/v1
+kind: ManagedCluster
+metadata:
+ labels:
+   kuadrant.io/lb-attribute-custom-weight: "GCP"
```
philbrookes, line 126: if a cluster has this attribute set (e.g. GCP), but that value isn't mentioned in the DNS Policy (i.e. no GCP value is specified), do we just apply the default weight to that cluster, in that scenario?

docs/proposals/DNSPolicy.md
```diff
@@ -0,0 +1,280 @@
...
+    GEO: #optional
+      default: IE # required with GEO. Choses a default DNS response when no particular response is defined for a request from an unknown GEO.
+```  
+
+### Available Load Balancing Strategies  
+
+GEO and Weighted load balancing are well understood strategies and this API effectively allow a complex requirement to be expressed relatively simply and executed by the gateway controller in the chosen DNS provider. Our default policy will execute a "Round Robin" weighted strategy which reflects the current default behaviour.
+
+With the above API we can provide weighted and GEO and weighted within a GEO. A weighted strategy with a minimum of a default weight is always required and the simplest type of policy. The multi-cluster gateway controller will set up a default policy when a gateway is discovered (shown below). This policy can be replaced or modified by the user.  A weighted strategy can be complimented with a GEO strategy IE they can be used together in order to provide a GEO and weighted (within a GEO) load balancing. By defining a GEO section, you are indicating that you want to use a GEO based strategy (how this works is covered below).
```
philbrookes, line 66:
```
A weighted strategy with a minimum of a default weight is always required and the simplest type of policy
```
Out of curiousity, are we setting a weight value on endpoints in gateways that don't have an attached policy? If so, what value do we use?

docs/proposals/DNSPolicy.md
maleck13: yes

docs/proposals/DNSPolicy.md
maleck13: We create a default policy now, so there will always be a policy

philbrookes: So the default policy is attached to any gateway that has no attached custom policies?

maleck13: yes

maleck13:
@mikenairn ready to merge if you can approve

Issue #113, opened 2023-03-10
Control Plane Gateway status


maleck13:
**What**

The control plane Gateway status should provide a rolled up view of the status of the per gateways it represents. It should also represent the condition of policies attached to those gateways.

- [x] Define what our CP gateway status will contain and communicate
- [x] Implement the rolled up status

Note for the policies at the CP level each of these should add in a status condition to the control plane gateway status which is not part of this work

david-martin:
Proposal for aggregation https://github.com/Kuadrant/multicluster-gateway-controller/pull/150

PR #275, opened 2023-06-23
IP Address & attachedRoutes status aggregation proposal

david-martin:
Follow up on #150 and related to #113


docs/proposals/status-aggregation.md
```diff
@@ -10,39 +10,127 @@ When multiple instances of a resource (e.g. a Gateway) are running across multip
 
 ## Proposal
 
-To solve this problem, I'm proposing we leverage the status conditions in the control plane instance of that resource, aggregating the statuses to convey the necessary information.
+To solve this problem, I'm proposing we leverage the status block in the control plane instance of that resource, aggregating the statuses to convey the necessary information.
 
-For example, if the `ListenersValid` status condition type of a `Gateway` is `True` for all instances of the `Gateway` resource across all clusters, then the `Gateway` in the control plane will have the `ListenersValid` status condition type also set to `True`.
+### Status Conditions
+
```
david-martin, line 16: This first set of changes around status conditions is fixing a mistake with the example `type`. 
ListenersValid is a reason, as opposed to the type, which is `Ready`.

docs/proposals/status-aggregation.md
```diff
@@ -10,39 +10,127 @@ When multiple instances of a resource (e.g. a Gateway) are running across multip
...
+and listeners, such as the number of attached routes for each listener.
+This information is useful at the control plane level.
+For example, a DNS Record should only exist as long as there is at least 1 attached route for a listener.
+It can also be more complicated than that when it comes to multi cluster gateways.
+A DNS Record should only include the IP Addresses of the Gateway instances where the listener has at least 1 attached route.
+This is important when initial setup of DNS Records happen as applications start.
+It doesn't make sense to route traffic to a Gateway where a listener isn't ready/attached yet.
+It also comes into play when a Gateway is displaced either due to changing placement decision or removal.
+
+In summary, the IP Addresses and number of attached routes per listener per Gateway instance is needed in the control plane to manage DNS effectively.
```
maleck13, line 64: would it make sense to say, an IP can only be added to the control plane if there is at least 1 listener on that gateway? That way any IP in the control plane can be used for a DNSRecord for any listener that has > 1 route 

So perhaps we can keep the structure the same at the control plane level?


```
status:
  addresses:
  - type: IPAddress
    value: 172.32.200.0 #ip from gateway instance with attached route for web
  - type: IPAddress
    value: 172.32.201.0 #ip from gateway instance with attached route for web
  listeners:
  - attachedRoutes: 0
    conditions:
    name: api
  - attachedRoutes: 2
    conditions:
    name: web
```

maleck13 (commented):
I wonder if we need to change the status structure. See comment, but perhaps I miss something

docs/proposals/status-aggregation.md
david-martin: Would this allow for the following scenario?

2 apps, (so 2 listeners) a.example.com & b.example.com were configured on 2 different clusters, Cluster X & Cluster Y.
The listener for app a is attached in Cluster X, but not Cluster Y.
The listener for app b is attached in both clusters.
If the IP Address for Cluster Y is added to the dns record for app a, some traffic will get a 503.

docs/proposals/status-aggregation.md
```diff
@@ -10,39 +10,127 @@ When multiple instances of a resource (e.g. a Gateway) are running across multip
...
+    - type: IPAddress
+      value: 172.32.202.0
+    - type: IPAddress
+      value: 172.32.203.0
+    listeners:
+    - attachedRoutes: 1
+      name: api
+    - attachedRoutes: 1
+      name: web
+```
```
mikenairn, line 136: In this example it has addresses from 2 different clusters, but there is no way to map back the ips to a cluster right?

This is probably good enough to ensure that we are updating this gateway when DNS needs to be updated, but it doesn't contain enough information for the DNS reconciler to fully work without querying the ocm placement api to get the addresses per cluster. Wondering if there is anything we can do here to surface the cluster name for each group of addresses?

docs/proposals/status-aggregation.md
maleck13: So perhaps we can add our own condition type that way we don't need to actually modify the CRD?
```
lastTransitionTime: "2023-06-22T07:02:05Z"
message: "cluster1 listener api ready to recieve traffic"
observedGeneration: 1
reason: "listener:api,cluster:c1,addresses: 192.168.22.1,aws.lb.com"
status: "True"
type: MCAddress
```

Not ideal either but at least with this we avoid a change to the type? WDYT @mikenairn @david-martin

mikenairn: @maleck13 presumably you can have many of these MCAddress type status conditions, one for each listener on each cluster?

docs/proposals/status-aggregation.md
david-martin: > In this example it has addresses from 2 different clusters, but there is no way to map back the ips to a cluster right?

Yes. There's no cluster or gateway instance identifier.


> Wondering if there is anything we can do here to surface the cluster name for each group of addresses

What kind of structure is needed/used currently?
Is it a map of cluster names to ip addresses?

docs/proposals/status-aggregation.md
david-martin: > So perhaps we can add our own condition type that way we don't need to actually modify the CRD?

Modifying the CRD isn't ideal.
Without modifying, we're bound by [the spec](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.GatewayStatus).

There's 3 fields we can work with: `addresses`, `conditions` and `listeners`.
Your suggestion to use conditions looks interesting.

> presumably you can have many of these MCAddress type status conditions

I haven't tried to use more than 1 condition with the same `type`, and it isn't explicitily forbidden [in the spec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#condition-v1-meta). 
It kind of feels against what the intention of condition types are.

I wonder if we encoded the listener & cluster into the type so they are unique.
Something like:

```yaml
reason: "192.168.22.1"
status: "True"
type: c1.api/MCAddress
```

and for attachedRoutes:

```yaml
reason: "3"
status: "True"
type: c1.api/AttachedRoutes
```


Alternatively, we could do similar naming in the addresses and listeners fields (instead of conditions) to store the state of all gateways.

e.g.

```yaml
status:
  addresses:
  - type: c1.api/IPAddress
    value: 192.168.26.1
  - type: c1.api/IPAddress
    value: 192.168.26.2
  - type: c2.api/IPAddress
    value: 192.168.27.1
  listeners:
  - name: c1.api
    attachedRoutes: 2
  - name: c2.api
    attachedRoutes: 0
```

I'm trying to think if there's any reason these approaches wouldn't work...

maleck13: the types in the address status have to be an `AddressType` perhaps the value could carry the format?

docs/proposals/status-aggregation.md
maleck13: We use the placement decision to find which clusters and then a manifestwork to find the cluster ips

docs/proposals/status-aggregation.md
mikenairn: Adding a prefix of  cluster name to the listener, and prefixing the cluster and listener name to the value of the address would work in allowing us to get all the required information for setting up DNS. Seems a better option than trying to work it out from conditions, or creating a new CRD IMO.

mikenairn: Maybe we want to use a `NamedAdress` type here https://github.com/robscott/gateway-api/blob/54a3bc3a01acae09330db5dddc179d335e36961a/apis/v1alpha2/gateway_types.go#L530? If I'm understanding what that is saying, it seems kind of close to what we want, a custom way of describing an address.

maleck13: yeah except in the comments that is marked as depreciated now

maleck13: either way lets go for a custom address @david-martin

maleck13: reading it  again we should be able to define our own addressType so this would work as you suggested @david-martin.

mikenairn: I see, i saw that but couldn't work out what it was saying. It looks like we can create our own custom type as long as it passes this validation https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1beta1/shared_types.go#L538. So i guess we could have a `kuadrant.io/IPAddress` and `kuadrant.io/Hostname` and the value of these types has the values with the prefixed meta required to work out what cluster and listener it is.

david-martin: > with the prefixed meta

so something like, keeping the type static and the value with all the gateway specific data?:

```yaml
    value: gateway_name/192.168.26.1
```

mikenairn: Yeah i think it is just cluster name we need here, and a listener for each cluster i.e.
```
  listeners:
  - name: cluster_name.api
    attachedRoutes: 2
  - name: cluster_name.api
    attachedRoutes: 0
```

david-martin:
Resolving conversations as proposal has been updated.

docs/proposals/status-aggregation.md
```diff
@@ -10,39 +10,135 @@ When multiple instances of a resource (e.g. a Gateway) are running across multip
...
+    conditions:
+    name: web
+---
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: Gateway
+metadata:
+  name: gateway
+status:
+  addresses:
+  - type: IPAddress
```
maleck13, line 96: how will we handle host names? Will it be `kuadrant.io/MultiClusterHostAddress`

docs/proposals/status-aggregation.md
```diff
@@ -10,39 +10,135 @@ When multiple instances of a resource (e.g. a Gateway) are running across multip
...
+    - attachedRoutes: 1
+      name: cluster_2.api
+    - attachedRoutes: 1
+      name: cluster_2.web
+```
+
+The MultiCluster Gateway Controller will use a custom implementation of the `addresses` and `listenerers` fields.
+The address `type` is of type [AddressType](https://github.com/kubernetes-sigs/gateway-api/blob/f883de997b88dd6ee138930198542da8a9b2f634/apis/v1beta1/shared_types.go#L552), where the type is a domain-prefixed string identifier. The length has a max of 253 characters.
+The value can be split on the forward slash, `/`, to give the cluster name and the underlying Gateway IPAddress value of type IPAddress.
+Only the IPAddress type will be supported at this time, though there is no obvious reason it couldn't be extended to support a Hostname address type later.
```
maleck13, line 138: not sure why?

david-martin: Simplest answer is that IPAddress is the only type we've seen so far.
I wouldn't want to add support for something that we don't actively test.

mikenairn: We will need HostAddress as well. I suppose at the moment we only use IPAddress, but we will after the proposed DNS record structure changes (being added right now https://github.com/Kuadrant/multicluster-gateway-controller/pull/293) want to support hostnames as well as we will be removing the HostResolver and instead setting a CNAME to the host instead see https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/docs/proposals/DNSRecordStructure.md#geo-agnostic-everything-is-in-a-default--geo-catch-all

maleck13: one thought I had was we should consider dropping kuadrant.io for mgc. IE make our prefix as short  as possible to avoid any issues with long hostnames (edge case I know)

david-martin: > but we will after the proposed DNS record structure changes

Ack.
I'll update the wording to include.
We can add an integration test case as a start.

> make our prefix as short as possible to avoid any issues with long hostnames

The kuadrant.io prefix would only be on the Address type, which would be limited to values of `kuadrant.io/MultiClusterIPAddress` and `kuadrant.io/MultiClusterHostName`.

The length would be a factor for cluster names in the prefix of the address value and the listener name

maleck13: ah yes ofc 👍

mikenairn: > one thought I had was we should consider dropping kuadrant.io for mgc. IE make our prefix as short as possible to avoid any issues with long hostnames (edge case I know)

It's a prefix to the address type so will the length not be constant? Using a domain was what it suggests in it's example https://github.com/kubernetes-sigs/gateway-api/blob/f883de997b88dd6ee138930198542da8a9b2f634/apis/v1beta1/shared_types.go#L538

maleck13 (approved):

PR #277, opened 2023-06-27
Remove dnsProvider from service

mikenairn:
Removes all calls to provider specific logic from the dns service. The dns service no longer requires knowing about the dns provider and is only used for CRUD operations on DNS related k8s resources.

Future work will remove the instance wide DNSProvider resource, and any controllers requiring provider access will instantiate their own instance of the DNSProvider resource.


pkg/dns/aws/dns.go
```diff
@@ -97,18 +96,6 @@ const (
...
-	totalIPs := 0
-	for _, e := range dnsRecord.Spec.Endpoints {
-		totalIPs += len(e.Targets)
-	}
-	for _, e := range dnsRecord.Spec.Endpoints {
-		e.SetProviderSpecific(ProviderSpecificWeight, awsEndpointWeight(totalIPs))
-	}
-	return nil
-}
-
```
mikenairn, line 111: The original intention for `AdjustEndpoints` was that it would allow us to call out o provider specific logic to manipulate endpoints into a provider specific set. As we have progressed the geo work we now realise that it would be much better if the DNSRecord and endpoints are consistent regardless of DNS provider and instead any provider specific logic is dealt with as part of the `Ensure` call.

pkg/dns/aws/dns.go
```diff
@@ -37,7 +37,6 @@ import (
 
 const (
 	ProviderSpecificEvaluateTargetHealth       = "aws/evaluate-target-health"
-	ProviderSpecificWeight                     = "aws/weight"
 	ProviderSpecificRegion                     = "aws/region"
 	ProviderSpecificFailover                   = "aws/failover"
 	ProviderSpecificGeolocationContinentCode   = "aws/geolocation-continent-code"
```
mikenairn, line 42: All of these will eventually be moved in favour of generic options that will be used by all providers. For now i have only moved the weight since it's the only one currently used by mgc.

pkg/dns/service.go
```diff
@@ -260,3 +263,20 @@ func (s *Service) CleanupDNSRecords(ctx context.Context, owner traffic.Interface
...
+//
+// For aws weight value must be an integer between 0 and 255.
+// https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-values-weighted.html#rrsets-values-weighted-weight
+func endpointWeight(numIPs int) string {
+	maxWeight := 120
+	if numIPs > maxWeight {
+		numIPs = maxWeight
+	}
+	return strconv.Itoa(maxWeight / numIPs)
+}
```
mikenairn, line 282: This will likely be removed eventually in favour of using default weights set in policy.

Issue #214, opened 2023-05-23
flesh out unit tests for DNS package


maleck13:
**What**

we are lacking in unit coverage for the DNS package.

PR #281, opened 2023-06-28
GH-214 dns package unit coverage

makslion:
## What 
Increasing test coverage for the DNS package. 
Touched only the `service.go`. 
The `health` controller doesn't have anything in there that would not be handy to test by integration suite. 
`target`, `hostresolv` and `dns` don't have much to unit test in them. 
The files in the AWS subpackage - not sure if it makes sense to also unit test them.

### DIFF
--- a/pkg/dns/service_test.go
+++ b/pkg/dns/service_test.go
@@ -4,12 +4,15 @@ package dns_test
 
 import (
 	"context"
+	"reflect"
 	"testing"
+	"time"
 
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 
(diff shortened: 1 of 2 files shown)


pkg/dns/service_test.go
```diff
@@ -360,3 +416,436 @@ func TestDNS_findMatchingManagedZone(t *testing.T) {
...
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			f := fake.NewClientBuilder().WithScheme(testScheme(t)).WithObjects(tt.dnsRecord).Build()
+			s := dns.NewService(f, &fakeHostResolver{})
+
+			if err := s.SetEndpoints(context.TODO(), tt.addresses, tt.dnsRecord, &v1alpha1.DNSPolicy{}); (err != nil) != tt.wantErr {
```
maleck13, line 592: do we not want to validate that the DNSRecord has been changed as expected?

maleck13 (commented):
just one comment on one test we seem to not validate the DNSRecord

pkg/dns/service_test.go
makslion: Good point. Done

pkg/dns/service_test.go
```diff
@@ -360,3 +416,497 @@ func TestDNS_findMatchingManagedZone(t *testing.T) {
...
+				},
+			},
+			record:  &v1alpha1.DNSRecord{},
+			wantErr: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gw := traffic.NewGateway(tt.gateway)
+			f := fake.NewClientBuilder().WithScheme(testScheme(t)).WithObjects(tt.record).Build()
```
maleck13, line 458: not sure if you are aware, but you can use the fake client to get the DNSRecord after deletion and it should be gone. The fake client acts like an in mem store

maleck13 (approved):

PR #283, opened 2023-06-28
Gh 228

R-Lawton:
## Verification 

1. run make local-setup OCM_SINGLE=true MGC_WORKLOAD_CLUSTERS_COUNT=1    
1. ensure a secret called mgc-aws-credentials in the namespace  multi-cluster-gateways is created with the creds from aws-credentials.env
2. ensure the managed zone cr created contains a reference in the spec to the name and namespace and type of the dnsprovider credentials from above 
```
spec:
   provider:
    name: mgc-aws-credentials
    type: AWS

```
1. run through the walkthrough https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/docs/how-to/ocm-control-plane-walkthrough.md everything should work as expected
2. Status in the managedzone should be updated


config/crd/bases/kuadrant.io_dnsproviders.yaml
```diff
@@ -0,0 +1,52 @@
...
+                    type: string
+                  namespace:
+                    type: string
+                required:
+                - namespace
+                type: object
+            required:
+            - credentials
+            type: object
+        type: object
```
maleck13, line 48: I would have expected the provider type to be a part of this spec?

config/crd/bases/kuadrant.io_managedzones.yaml
```diff
@@ -46,6 +46,19 @@ spec:
...
               internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
             type: string
+          dnsprovider:
+            properties:
+              name:
+                type: string
+              namespace:
+                type: string
+              type:
+                type: string
```
maleck13, line 56: not sure why we would want this in the zone rather than the provider?

config/local-setup/controller/managed_zone.yaml
```diff
@@ -6,3 +6,8 @@ spec:
   id: DUMMY_ID
   domainName: DUMMY_DOMAIN_NAME
   description: "Dev Managed Zone"
+  provider:
+    name: mgc-dns-provider
```
maleck13, line 10: yeah def find this a bit odd, for the provider not to specify which provider it is but instead to specify that in the zone

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
```diff
@@ -71,8 +73,13 @@ func (r *DNSPolicyReconciler) reconcileGatewayHealthChecks(ctx context.Context,
 		// Keep a copy of the DNSRecord to check if it needs to be updated after
 		// reconciling the health checks
 		dnsRecordOriginal := mh.DnsRecord.DeepCopy()
+		var managedzone *v1alpha1.ManagedZone
+		for _, host := range gatewayAccessor.GetHosts() {
```
maleck13, line 77: this looks odd, looks ass though we will only get the last zone as the zone is reassigned with each iteration and then used outside of the loop. Also is there no error returned from GetManagedZoneForHost? Do we perhaps want to do a GetZoneForDNSRecord? as I recall DNSRecord references the zone

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
```diff
@@ -89,8 +96,14 @@ func (r *DNSPolicyReconciler) reconcileGatewayHealthChecks(ctx context.Context,
 	return allResults, nil
 }
 
-func (r *DNSPolicyReconciler) reconcileDNSRecordHealthChecks(ctx context.Context, dnsRecord *v1alpha1.DNSRecord, config *healthChecksConfig) ([]dns.HealthCheckResult, error) {
-	healthCheckReconciler := r.DNSProvider.HealthCheckReconciler()
+func (r *DNSPolicyReconciler) reconcileDNSRecordHealthChecks(ctx context.Context, dnsRecord *v1alpha1.DNSRecord, config *healthChecksConfig, managedzone *v1alpha1.ManagedZone) ([]dns.HealthCheckResult, error) {
+
+	DNSProvider, err := r.DNSProv.CreateDNSProvider(ctx, managedzone)
```
maleck13, line 101: maybe I am missing something here? Not sure why we are creating a DNSProvider when reconciling a health check? I would have thought we would fail to reconcile health checks until there was a DNSProvider present. Same with managed zones and DNSRecords. If there is no provider present we can't really do anything

maleck13: nvm I see it is actually not creating the DNSProvider resource

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
R-Lawton: Yes your right that's my mistake i messed up my looping and forgot my error.  Yes lets go with the GetZoneForDNSRecord instead.

pkg/controllers/dnsrecord/dnsrecord_controller.go
```diff
@@ -39,12 +40,13 @@ const (
 )
 
 var Clock clock.Clock = clock.RealClock{}
+var DNSProvider dns.Provider
```
maleck13, line 43: This looks to be a package level var that is set in the reconcile and used in other functions? This could lead to odd behaviour if multiple threads of execution  as they would change the value of this for all the functions using that var

pkg/controllers/dnsrecord/dnsrecord_controller.go
```diff
@@ -94,6 +102,11 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	reason = "ProviderSuccess"
 	message = "Provider ensured the managed zone"
 
+	DNSProvider, err = r.DNSProv.CreateDNSProvider(ctx, managedZone)
```
maleck13, line 105: as mentioned I think I would switch this to be DNSPRovider.Factory() as it is returns a different impl of the provider interface depending on the input

pkg/dns/dnsprovider/dnsProvider.go
```diff
@@ -0,0 +1,104 @@
...
+func (p *provider) CreateDNSProvider(ctx context.Context, managedZone *v1alpha1.ManagedZone) (dns.Provider, error) {
+	creds, err := p.dnsProviderSecret(ctx, managedZone)
+	if err != nil {
+		return nil, err
+	}
+
+	switch managedZone.Spec.ProviderRef.ProviderType {
+	case "AWS":
+		log.Log.Info("Creating DNS provider for provider type AWS")
+		DNSProvider, _ = aws.NewDNSProvider(*creds)
```
maleck13, line 89: prob should make sure we return the error here

pkg/dns/dnsprovider/dnsProvider.go
```diff
@@ -0,0 +1,104 @@
...
+		return DNSProvider, nil
+
+	case "GCP":
+		log.Log.Info("GCP")
+
+	case "AZURE":
+		log.Log.Info("AZURE")
+
+	default:
```
maleck13, line 98: I think here I would return an error UnsupportedDNSProvider

pkg/controllers/managedzone/managedzone_controller.go
```diff
@@ -38,11 +38,18 @@ const (
 	ManagedZoneFinalizer = "kuadrant.io/managed-zone"
 )
 
+var DNSProvider dns.Provider
+
+type DNS interface {
```
maleck13, line 43: I think I would move this to the dns package and provider.go. Also I don't expect we will add anything more to this interface so you could change it just to be ``` type ProviderFactory func(ctx context.Context, managedZone *v1alpha1.ManagedZone) (Provider, error) ```

maleck13: Then in the controllers you can set ```dnsProviderFactory dns.ProviderFactory``` in the struct

pkg/dns/aws/dns.go
```diff
@@ -56,8 +56,8 @@ type Route53DNSProvider struct {
 var _ dns.Provider = &Route53DNSProvider{}
 
 // NewDNSProvider returns a Route53DNSProvider instance configured for the AWS Route 53 service using the credentials provided
-func NewDNSProvider(route53Config v1alpha1.DNSProviderConfigRoute53) (*Route53DNSProvider, error) {
-	if route53Config.AccessKeyID == "" || route53Config.SecretAccessKey == "" {
+func NewDNSProvider(dnsProviderConfig v1alpha1.DNSProviderConfig) (*Route53DNSProvider, error) {
```
maleck13, line 59: I think I would remove this dnsProviderConfig

pkg/apis/v1alpha1/dnsprovider_types.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+type DNSProviderConfig struct {
+	// Route53 configures this config to communicate with AWS Route 53
+	// +optional
+	Route53 *DNSProviderConfigRoute53 `json:"route53,omitempty"`
+}
+
+// DNSProviderConfigRoute53 is a structure containing the Route 53 configuration for AWS
+type DNSProviderConfigRoute53 struct {
```
maleck13, line 53: same

pkg/apis/v1alpha1/dnsprovider_types.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+type DNSProviderCredentials struct {
+	// +required
+	Namespace    string `json:"namespace"`
+	Name         string `json:"name,omitempty"`
+	ProviderType string `json:"type"`
+}
+
+type DNSProviderConfig struct {
```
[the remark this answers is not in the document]
R-Lawton: Ooops nope ill remove these two

pkg/controllers/dnsrecord/dnsrecord_controller.go
```diff
@@ -156,45 +164,28 @@ func (r *DNSRecordReconciler) deleteRecord(ctx context.Context, dnsRecord *v1alp
 // DNSRecord (dnsRecord.Status.ParentManagedZone).
 func (r *DNSRecordReconciler) publishRecord(ctx context.Context, dnsRecord *v1alpha1.DNSRecord) error {
 
-	managedZone, err := r.getDNSRecordManagedZone(ctx, dnsRecord)
+	managedZone, err := r.HostService.GetDNSRecordManagedZone(ctx, dnsRecord)
 	if err != nil {
-		return err
+		// If the Managed Zone isn't found, just continue
+		return client.IgnoreNotFound(err)
```
maleck13, line 170: why do we ignore not found here? Can there be a valid reason there is no managedzone for a dnsrecord?

R-Lawton: good catch , its a copy and paste error from a different function in that file calling it, ill change that

R-Lawton: its being used in deleteRecord function im persumming its in there in a scenario if a person deletes things out of order and deletes a managedzone the dnsrecord will still delete

pkg/dns/dnsprovider/dnsProvider.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func NewProvider(c client.Client) *providerFactory {
+
+	return &providerFactory{
+		Client: c,
+	}
+}
+
+func (p *providerFactory) loadProviderSecret(ctx context.Context, managedZone *v1alpha1.ManagedZone) (*v1.Secret, *v1alpha1.DNSProvider, error) {
```
maleck13, line 34: small: prob would just name this loadProvider

maleck13 (commented):
code changes look reasonable, will try it here to verify it

maleck13:
@R-Lawton when this merges everyone local setup is going to fail. So might be worth announcing in the slack channel that they need to setup these new env files

mikenairn:
> 

/hold 

There shouldn't be any reason to break local-setup

config/local-setup/controller/kustomization.yaml
```diff
@@ -5,24 +5,25 @@ namePrefix: mgc-
...
     envs:
-      - ../../../controller-config.env
+      - ../../../dns-provider-config.env
 
 secretGenerator:
-  - name: aws-credentials
+  - name: provider-credentials
     envs:
-      - ../../../aws-credentials.env
+      - ../../../provider-credentials.env
```
mikenairn, line 21: There's really no need to change these expected file names. Whatever way you eventually do this, you will be creating something with "AWS" credentials in it for an AWS provider, so "aws-credentials.env" make more sense anyway.

config/local-setup/controller/kustomization.yaml
```diff
@@ -5,24 +5,25 @@ namePrefix: mgc-
...
 
 generatorOptions:
   disableNameSuffixHash: true
 
 configMapGenerator:
-  - name: controller-config
+  - name: provider-config
     envs:
-      - ../../../controller-config.env
+      - ../../../dns-provider-config.env
```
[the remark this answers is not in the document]
R-Lawton: ok cool ill change these back

pkg/dns/dnsprovider/dnsProvider.go
```diff
@@ -0,0 +1,62 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      managedZone.Spec.SecretRef.Name,
+			Namespace: managedZone.Spec.SecretRef.Namespace,
+		}}
+
+	if err := p.Client.Get(ctx, client.ObjectKeyFromObject(providerSecret), providerSecret); err != nil {
+		return nil, err
+	}
+
+	if providerSecret.Type != "kuadrant.io/aws" {
```
maleck13, line 43: lets move this to a constant `DNSProviderAWS = "kuadrant.io/aws"`

pkg/dns/service.go
```diff
@@ -323,3 +324,26 @@ func (s *Service) CleanupDNSRecords(ctx context.Context, owner traffic.Interface
...
+	}
+
+	managedZone := &v1alpha1.ManagedZone{}
+
+	err := s.controlClient.Get(ctx, client.ObjectKey{Namespace: dnsRecord.Namespace, Name: dnsRecord.Spec.ManagedZoneRef.Name}, managedZone)
+	if err != nil {
+		return nil, err
+	}
+
+	managedZoneReady := meta.IsStatusConditionTrue(managedZone.Status.Conditions, "Ready")
```
maleck13, line 342: wondering why we check this here? I feel like whatever is using the Zone should do this check

maleck13 (commented):
minor comments, looking good. Will try it out locally

pkg/dns/dnsprovider/dnsProvider.go
R-Lawton: sure

pkg/dns/service.go
R-Lawton: thats a good question im not sure. that change was merged a few days ago as part of the bug Mike found.

mikenairn: This shouldn't be here, where this method was originally should still be doing it though https://github.com/Kuadrant/multicluster-gateway-controller/pull/283/files#diff-6321f40c52e93a933e0ad7d590b390c749c8e6b757a10b40bf4d21b9a4fa4603R167

PROJECT
```diff
@@ -11,6 +11,14 @@ resources:
...
   path: github.com/Kuadrant/multicluster-gateway-controller/pkg/apis/v1alpha1
   version: v1
+- api:
+    crdVersion: v1alpha1
+    namespaced: true
+  controller: true
+  group: kuadrant.io
+  kind: DNSProvider
+  path: github.com/Kuadrant/multicluster-gateway-controller/pkg/apis/v1alpha1
+  version: v1
```
mikenairn, line 21: Can be removed

pkg/controllers/managedzone/managedzone_controller.go
```diff
@@ -164,7 +163,11 @@ func (r *ManagedZoneReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 func (r *ManagedZoneReconciler) publishManagedZone(ctx context.Context, managedZone *v1alpha1.ManagedZone) error {
 
-	mzResp, err := r.DNSProvider.EnsureManagedZone(managedZone)
+	DNSProvider, err := r.DNSProvider(ctx, managedZone)
+	if err != nil {
```
maleck13, line 167: I think we could check whether this is a not found err, and if it is update the status of the zone with an error condition

pkg/controllers/managedzone/managedzone_controller.go
```diff
@@ -182,7 +185,11 @@ func (r *ManagedZoneReconciler) deleteManagedZone(ctx context.Context, managedZo
 		return nil
 	}
 
-	err := r.DNSProvider.DeleteManagedZone(managedZone)
+	DNSProvider, err := r.DNSProvider(ctx, managedZone)
+	if err != nil {
```
maleck13, line 189: same here, adding a status condition on the zone if we fail to load the provider. Otherwise we will error out in the logs etc but the user will likely not have an external info letting them know what is happening

pkg/apis/v1alpha1/managedzone_types.go
```diff
@@ -39,6 +39,15 @@ type ManagedZoneSpec struct {
 	// Reference to another managed zone that this managed zone belongs to.
 	// +optional
 	ParentManagedZone *ManagedZoneReference `json:"parentManagedZone,omitempty"`
+	// +required
+	SecretRef *SecretRef `json:"provider"`
```
mikenairn, line 43: Why provider? Feels like it should be a bit more explicit that it's a reference to a secret

R-Lawton: no particular reason just a name i chose since its to do with the provider i can change it to make it clearer

pkg/dns/dns.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package dns
 
 import (
+	"context"
 	"github.com/Kuadrant/multicluster-gateway-controller/pkg/apis/v1alpha1"
```
mikenairn, line 21: import grouping. Keep the stdlib imports separate at the top.

pkg/dns/dnsprovider/dnsProvider.go
```diff
@@ -0,0 +1,62 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/Kuadrant/multicluster-gateway-controller/pkg/apis/v1alpha1"
+	"github.com/Kuadrant/multicluster-gateway-controller/pkg/dns"
+	"github.com/Kuadrant/multicluster-gateway-controller/pkg/dns/aws"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
mikenairn, line 15: import grouping. This should be with the other k8s imports.

pkg/dns/dnsprovider/dnsProvider.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func NewProvider(c client.Client) *providerFactory {
+
+	return &providerFactory{
+		Client: c,
+	}
+}
+
+// depending on the provider type given in dnsprovider cr using the credentials supplied returns a dnsprovider.
```
mikenairn, line 31: comment is out of date

pkg/dns/dnsprovider/dnsProvider.go
```diff
@@ -0,0 +1,62 @@
...
+			Name:      managedZone.Spec.SecretRef.Name,
+			Namespace: managedZone.Spec.SecretRef.Namespace,
+		}}
+
+	if err := p.Client.Get(ctx, client.ObjectKeyFromObject(providerSecret), providerSecret); err != nil {
+		return nil, err
+	}
+
+	if providerSecret.Type != "kuadrant.io/aws" {
+		return nil, fmt.Errorf("no dnsprovider types were found")
```
mikenairn, line 44: I don't think you really need this if you add a default in the switch statement below that throws an error.

pkg/dns/service.go
```diff
@@ -20,6 +20,7 @@ import (
 	"github.com/Kuadrant/multicluster-gateway-controller/pkg/_internal/slice"
 	"github.com/Kuadrant/multicluster-gateway-controller/pkg/apis/v1alpha1"
 	"github.com/Kuadrant/multicluster-gateway-controller/pkg/traffic"
+	"k8s.io/apimachinery/pkg/api/meta"
```
mikenairn, line 23: import grouping

pkg/dns/dns.go
R-Lawton: sorry i thought the make fmt sorted all that let me fix it

pkg/dns/dnsprovider/dnsProvider.go
R-Lawton: sure i can remove it, theres a default there do you suggest something different added?

mikenairn: No, it's throwing `errUnsupportedProvider` so should be good. It will never throw that error with this extra check here.

pkg/controllers/managedzone/managedzone_controller.go
R-Lawton: Actually, this gets checked on line 107 so would it not cause an issue with both trying to update it?

pkg/dns/service.go
R-Lawton: i moved the original function but ill move the is status condition back

pkg/dns/dnsprovider/dnsProvider.go
R-Lawton: yes i understand makes sense

pkg/controllers/dnspolicy/dnspolicy_controller_test.go
```diff
@@ -125,6 +129,20 @@ var _ = Describe("DNSPolicy", Ordered, func() {
...
+					Status:             "True",
+					LastTransitionTime: metav1.Now(),
+					Message:            "Provider ensured the managed zone",
+					ObservedGeneration: 1,
+					Reason:             "ProviderSuccess",
+				},
+			},
+		}
+		Expect(k8sClient.Status().Update(ctx, managedZone)).To(BeNil())
+
```
mikenairn, line 145: I don't think you need this here.

pkg/controllers/managedzone/managedzone_controller.go
```diff
@@ -164,7 +165,11 @@ func (r *ManagedZoneReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 func (r *ManagedZoneReconciler) publishManagedZone(ctx context.Context, managedZone *v1alpha1.ManagedZone) error {
 
-	mzResp, err := r.DNSProvider.EnsureManagedZone(managedZone)
+	DNSProvider, err := r.DNSProvider(ctx, managedZone)
```
mikenairn, line 168: DNSProvider -> dnsProvider

pkg/controllers/managedzone/managedzone_controller.go
```diff
@@ -24,8 +24,10 @@ import (
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
+
+	"k8s.io/apimachinery/pkg/runtime"
+
```
mikenairn, line 30: Looks like an unnecessary chnage

pkg/controllers/dnsrecord/dnsrecord_controller.go
```diff
@@ -156,45 +169,32 @@ func (r *DNSRecordReconciler) deleteRecord(ctx context.Context, dnsRecord *v1alp
...
+		return fmt.Errorf("the managed zone is not in a ready state : %s", managedZone.Name)
+	}
 
 	if dnsRecord.Generation == dnsRecord.Status.ObservedGeneration {
 		log.Log.V(3).Info("Skipping managed zone to which the DNS dnsRecord is already published", "dnsRecord", dnsRecord.Name, "managedZone", managedZone.Name)
 		return nil
 	}
-
-	err = r.DNSProvider.Ensure(dnsRecord, managedZone)
+	DNSProvider, err := r.DNSProvider(ctx, managedZone)
```
mikenairn, line 186: DNSProvider -> dnsProvider

pkg/controllers/managedzone/managedzone_controller.go
```diff
@@ -182,7 +187,17 @@ func (r *ManagedZoneReconciler) deleteManagedZone(ctx context.Context, managedZo
 		return nil
 	}
 
-	err := r.DNSProvider.DeleteManagedZone(managedZone)
+	DNSProvider, err := r.DNSProvider(ctx, managedZone)
```
mikenairn, line 190: DNSProvider -> dnsProvider

pkg/dns/dnsprovider/dnsProvider.go
```diff
@@ -0,0 +1,57 @@
...
+		}}
+
+	if err := p.Client.Get(ctx, client.ObjectKeyFromObject(providerSecret), providerSecret); err != nil {
+		return nil, err
+	}
+
+	switch providerSecret.Type {
+	case "kuadrant.io/aws":
+		log.Log.Info("Creating DNS provider for provider type AWS")
+		DNSProvider, err := aws.NewProviderFromSecret(providerSecret)
```
mikenairn, line 45: DNSProvider -> dnsProvider

pkg/dns/dnsprovider/dnsProvider.go
```diff
@@ -0,0 +1,57 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("unable to create dns provider from secret: %v", err)
+		}
+		log.Log.Info("Route53 provider created")
+
+		return DNSProvider, nil
```
mikenairn, line 51: Since this gets called all over the place, you might want to consider what logs you want here, currently you get: 
```
2023-07-12T12:15:36+01:00       INFO    Creating DNS provider for provider type AWS
2023-07-12T12:15:36+01:00       INFO    Route53 provider created
```

repeated a lot.

Maybe a single log that mentions the managed zone it's being create for might be more useful , and possibly a debug log instead of info.

mikenairn:
/hold cancel

This looks good, tried it locally and things work as expected, few small comments. 
Not sure about the e2e test, but when that passes i think we should be good to merge.

Secret is created as expected:
```
kubectl get secret -n multi-cluster-gateways
NAME                   TYPE                DATA   AGE
mgc-aws-credentials    kuadrant.io/aws     3      28m
myapp.mn.hcpapps.net   kubernetes.io/tls   3      3m54s
```

Would be nice if some details of the provider type could be surfaced when listing zones(`kubectl get managedzones -n multi-cluster-gateways`), but not something we need to do now.

pkg/controllers/managedzone/managedzone_controller.go
R-Lawton: Sorry i made those changes but i must not have committed it

pkg/dns/dnsprovider/dnsProvider.go
R-Lawton: Ok cool ill change that

mikenairn (commented):
wtf

pkg/dns/service.go
```diff
@@ -323,3 +324,20 @@ func (s *Service) CleanupDNSRecords(ctx context.Context, owner traffic.Interface
...
+
+	managedZone := &v1alpha1.ManagedZone{}
+
+	err := s.controlClient.Get(ctx, client.ObjectKey{Namespace: dnsRecord.Namespace, Name: dnsRecord.Spec.ManagedZoneRef.Name}, managedZone)
+	if err != nil {
+		return nil, err
+	}
+
+	return managedZone, nil
+}
```
mikenairn, line 343: @maleck13 is talking about removing this service, so we might want to reconsider adding this here. That said, I'm fine with leaving it and making Craig deal with it. You will need to do this in a few places, so a util function seems reasonable somewhere.

R-Lawton: id like to get this pr through so if things arent still concrete yet id say wait to change it when we remove the service

mikenairn: agreed, lets get this merged so we can move on

PR #288, opened 2023-06-30
add grace period to descheduled gateways

philbrookes:
## Changes 
- Allow a grace period of 600 seconds when descheduling a gateway from a spoke cluster
- Update e2e tests
- Add new unit tests
- Few small refactoring changes 
  - Avoiding messy imports
  - Straightening out some of the managed zone matching logic

 ## Verification
follow [walkthrough](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/docs/how-to/ocm-control-plane-walkthrough.md) until gateway is replicated to the spoke cluster.
- deschedule the cluster:
  - kubectl label --overwrite managedcluster kind-mgc-control-plane ingress-cluster=false
  - See gateway and manifestworks still exist: 
    - kubectl get gateway -A
    - kubectl get manifestworks -A
  - Wait 10 minutes
    - See gateway and manifestworks are deleted

### DIFF
--- a/pkg/_internal/gracePeriod/gracePeriod.go
+++ b/pkg/_internal/gracePeriod/gracePeriod.go
@@ -4,10 +4,10 @@ import (
 	"context"
 	"fmt"
 	"strconv"
-	"strings"
 	"time"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	"github.com/Kuadrant/multicluster-gateway-controller/pkg/_internal/metadata"
 	"github.com/Kuadrant/multicluster-gateway-controller/pkg/dns"
@@ -20,13 +20,11 @@ const (
 
 var ErrGracePeriodNotExpired = fmt.Errorf("grace period has not yet expired")
 
-func WasGracePeriodNotExpiredErr(e error) bool {
-	return strings.Contains(e.Error(), "grace period has not yet expired")
-}
-
 func GracefulDelete(ctx context.Context, c client.Client, obj client.Object) error {
+	log := log.Log
 	at := time.Now().Add(DefaultGracePeriod)
 	if err := c.Get(ctx, client.ObjectKeyFromObject(obj), obj); err != nil {
+		log.V(3).Info("error finding object to graceful delete")
 		return err
 	}
 
@@ -35,23 +33,32 @@ func GracefulDelete(ctx context.Context, c client.Client, obj client.Object) err
 		return c.Delete(ctx, obj)
 	}
 
-	//ensure finalizer and annotation are present
+	//ensure annotation is present
 	if !metadata.HasAnnotation(obj, GraceTimestampAnnotation) {
+		log.V(3).Info("no grace annotation set, adding one now")
 		metadata.AddAnnotation(obj, GraceTimestampAnnotation, strconv.FormatInt(at.Unix(), 10))
-		return c.Update(ctx, obj)
-	}
-	deleteAt, err := strconv.Atoi(obj.GetAnnotations()[GraceTimestampAnnotation])
-	if err != nil {
-		//badly formed deleteAt annotation, remove it, so it will be regenerated
-		metadata.RemoveAnnotation(obj, GraceTimestampAnnotation)
 		if err := c.Update(ctx, obj); err != nil {
 			return err
 		}
-	} else {
-		//grace time reached, delete it
-		if int64(deleteAt) <= time.Now().Unix() {
-			return c.Delete(ctx, obj)
-		}
+		return ErrGracePeriodNotExpired
 	}
+	deleteAt, err := strconv.Atoi(obj.GetAnnotations()[GraceTimestampAnnotation])
+	if err != nil {
+		log.V(3).Info("existing grace annotation has bad value, resetting it")
+		metadata.AddAnnotation(obj, GraceTimestampAnnotation, strconv.FormatInt(at.Unix(), 10))
+		if err := c.Update(ctx, obj); err != nil {
+			return err
+		}
+		return ErrGracePeriodNotExpired
+	}
+
+	//grace time reached, delete it
+	if int64(deleteAt) <= time.Now().Unix() {
+		log.V(3).Info("grace period expired, removing object")
+		return c.Delete(ctx, obj)
+	}
+
+	log.V(3).Info("grace period still pending")
+
 	return ErrGracePeriodNotExpired
 }
--- a/pkg/_internal/gracePeriod/gracePeriod.go
+++ b/pkg/_internal/gracePeriod/gracePeriod.go
@@ -1,0 +1,57 @@
+package gracePeriod
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/Kuadrant/multicluster-gateway-controller/pkg/_internal/metadata"
+	"github.com/Kuadrant/multicluster-gateway-controller/pkg/dns"
+)
+
+const (
+	GraceTimestampAnnotation = "kuadrant.io/grace-timeout"
+	DefaultGracePeriod       = time.Second * dns.DefaultTTL * 10
+)
+
+var ErrGracePeriodNotExpired = fmt.Errorf("grace period has not yet expired")
+
+func WasGracePeriodNotExpiredErr(e error) bool {
+	return strings.Contains(e.Error(), "grace period has not yet expired")
+}
+
+func GracefulDelete(ctx context.Context, c client.Client, obj client.Object) error {
+	at := time.Now().Add(DefaultGracePeriod)
+	if err := c.Get(ctx, client.ObjectKeyFromObject(obj), obj); err != nil {
+		return err
+	}
+
+	// if At is before the current time, just delete it
+	if at.Before(time.Now()) || at.Equal(time.Now()) {
+		return c.Delete(ctx, obj)
+	}
+
+	//ensure finalizer and annotation are present
+	if !metadata.HasAnnotation(obj, GraceTimestampAnnotation) {
+		metadata.AddAnnotation(obj, GraceTimestampAnnotation, strconv.FormatInt(at.Unix(), 10))
+		return c.Update(ctx, obj)
+	}
+	deleteAt, err := strconv.Atoi(obj.GetAnnotations()[GraceTimestampAnnotation])
+	if err != nil {
+		//badly formed deleteAt annotation, remove it, so it will be regenerated
+		metadata.RemoveAnnotation(obj, GraceTimestampAnnotation)
+		if err := c.Update(ctx, obj); err != nil {
+			return err
+		}
+	} else {
+		//grace time reached, delete it
+		if int64(deleteAt) <= time.Now().Unix() {
+			return c.Delete(ctx, obj)
+		}
+	}
+	return ErrGracePeriodNotExpired
+}
--- a/pkg/_internal/gracePeriod/gracePeriod_test.go
+++ b/pkg/_internal/gracePeriod/gracePeriod_test.go
@@ -2,6 +2,7 @@ package gracePeriod
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"testing"
 	"time"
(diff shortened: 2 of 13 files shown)


pkg/_internal/gracePeriod/gracePeriod.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/Kuadrant/multicluster-gateway-controller/pkg/_internal/metadata"
+)
+
+const (
+	graceTimestampAnnotation = "kuadrant.io/grace-timeout"
+)
+
+var ErrGracePeriodNotExpired = fmt.Errorf("grace period has not yet expired")
+
+func WasGracePeriodNotExpiredErr(e error) bool {
```
maleck13, line 21: can we keep this consistent with the other error checking functions? ```IsGracePeriodNotExpiredErr(err)``` also why not just check it is equal to `ErrGracePeriodNotExpired`?
```
if err == nil{
return false}
return err == ErrGracePeriodNotExpired
```

pkg/_internal/gracePeriod/gracePeriod.go
```diff
@@ -0,0 +1,57 @@
...
+	if at.Before(time.Now()) || at.Equal(time.Now()) {
+		return c.Delete(ctx, obj)
+	}
+
+	//ensure finalizer and annotation are present
+	if !metadata.HasAnnotation(obj, graceTimestampAnnotation) {
+		metadata.AddAnnotation(obj, graceTimestampAnnotation, strconv.FormatInt(at.Unix(), 10))
+		if err := c.Update(ctx, obj); err != nil {
+			return err
+		}
```
maleck13, line 40: nit: maybe return nil here and drop the else?

pkg/_internal/gracePeriod/gracePeriod.go
```diff
@@ -0,0 +1,57 @@
...
+		return err
+	}
+
+	// if At is before the current time, just delete it
+	if at.Before(time.Now()) || at.Equal(time.Now()) {
+		return c.Delete(ctx, obj)
+	}
+
+	//ensure finalizer and annotation are present
+	if !metadata.HasAnnotation(obj, graceTimestampAnnotation) {
```
maleck13, line 36: is this always a "downstream" object? I find it weird that we are setting the annotation

pkg/_internal/gracePeriod/gracePeriod.go
```diff
@@ -0,0 +1,57 @@
...
+
+func GracefulDelete(ctx context.Context, c client.Client, obj client.Object, at time.Time) error {
+	if err := c.Get(ctx, client.ObjectKeyFromObject(obj), obj); err != nil {
+		return err
+	}
+
+	// if At is before the current time, just delete it
+	if at.Before(time.Now()) || at.Equal(time.Now()) {
+		return c.Delete(ctx, obj)
+	}
```
maleck13, line 33: so I am not getting why we need more than this?

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -269,8 +278,20 @@ func (r *GatewayReconciler) reconcileDownstreamFromUpstreamGateway(ctx context.C
 		}
 	}
 
+	longestTTL := 0
```
maleck13, line 281: this seems really odd here? should it not be a part of the DNS service perhaps? Also wondering if we do this do we really need the TTL on the dns record.

maleck13: Also I am thinking that what we really want is the TTL for the record for the gateway we are about to remove. IE the A or CNAME record that is actually pointing at the gateway instance / lb instance.
For example we could have a CNAME lb.my.zone.com that is a CNAME to ie.lb.my.zone.com that has several weighted records 1 per gateway instance that are A records with IPs. The A records will likely have a lower TTL while the geo CNAMEs will likely have a longer TTL. Anyway point is we want the TTL for the record that is specific to the instance we are removing.

maleck13: So if TTL is in the DNSPolicy it is the TTL  for the record pointing at the gateway record and we use it to then calculate what "our" Geo / Weighted record values should be. cc @mikenairn 
If we do this then the TTL in the DNSPolicy is all we need

pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -235,3 +238,18 @@ func (r *DNSPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		}
+		return
+	}
+
+	if policy.Spec.TTL.Value <= 0 {
+		policy.Spec.TTL.Value = v1alpha1.DNSPoliciesDefaultTTL
+	}
+
+	return
+}
```
mikenairn, line 255: Seems odd to change the spec of the policy in this way

pkg/dns/service.go
```diff
@@ -184,7 +202,7 @@ func (s *Service) SetEndpoints(ctx context.Context, addresses []gatewayv1beta1.G
 			Targets:       []string{ep},
 			RecordType:    "A",
 			SetIdentifier: ep,
-			RecordTTL:     60,
+			RecordTTL:     v1alpha1.TTL(dnsPolicy.Spec.TTL.Value),
```
mikenairn, line 205: This is assuming a policy will always set a TTL. Is that really what we want?

pkg/_internal/gracePeriod/gracePeriod.go
philbrookes: I did initially do it that way, but the error get's wrapped in another error, which is why I changed the name and the matching system to this, rather than reducing the useful info in the error.

pkg/_internal/gracePeriod/gracePeriod.go
philbrookes: the `at` timestamp is recalculated on each reconcile, this is simply handling for if the grace period is zero and no TTL was found, then there's no need to fuss around with annotations, just delete it.

But the annotation below is required if the `at` is anything later than now, so we know when to remove it, the new at would only be honoured if the previous at annotation was somehow corrupted or removed manually.

pkg/controllers/dnspolicy/dnspolicy_controller.go
philbrookes: good point, I'll update the CRD with a default.

pkg/controllers/gateway/gateway_controller.go
maleck13: after more conversation. We will remove TTL from the policy as it is likely to add confusion and it is not clear what the usecases are we would be solving. We will also not expose the grace period but rather set that internally for now

pkg/dns/service.go
maleck13: decided to remove TTL from policy

maleck13 (commented):
Based on further discussion, we are agreed on removing the TTL property from the DNSPolicy as it could cause confusion espically as it would not be applied across all records. Instead we will manage TTLs internally as we are the owners of the DNSRecord.
Additionally we will implement a grace period internally and not expose this currently until we have clearer use cases for needing to change it

pkg/_internal/gracePeriod/gracePeriod.go
philbrookes: This still won't work for me, unless I remove some debugging info along the return path.

The current return flow for this error looks like:
-> return ErrGracePeriodNotExpired
-> return fmt.Errorf(err, "extra debugging info here")
-> check if it's a gracePeriodNotExpired error.

maleck13: yeah so I think you need to use the Wrap.

you can use %w I believe
```
err = fmt.Errorf("some info : %w.", ErrGracePeriodNotExpired)
```

maleck13: %w wraps the error and then go can use the Unwrap which is used underneath in `Is`

maleck13:
prob worth updating the description here now, just for posterity

philbrookes:
> prob worth updating the description here now, just for posterity

done

pkg/_internal/gracePeriod/gracePeriod.go
maleck13: @philbrookes lets change it to use the wrapped error and switch to `IsGracePeriodNotExpiredErr`

pkg/_internal/gracePeriod/gracePeriod_test.go
```diff
@@ -0,0 +1,69 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+	workv1 "open-cluster-management.io/api/work/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	"github.com/Kuadrant/multicluster-gateway-controller/pkg/_internal/metadata"
+)
+
+func init() {
```
maleck13, line 19: seems odd to add it here, wonder should this be in main?

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -164,6 +167,13 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	requeue, programmedStatus, clusters, reconcileErr := r.reconcileDownstreamFromUpstreamGateway(ctx, upstreamGateway, params)
 	// gateway now in expected state, place gateway and its associated objects in correct places. Update gateway spec/metadata
 	if reconcileErr != nil {
+		if err == gracePeriod.ErrGracePeriodNotExpired || errors.Unwrap(err) == gracePeriod.ErrGracePeriodNotExpired {
```
maleck13, line 170: I think this can be simplified
``` if errors.Is(gracePeriod.ErrGracePeriodNotExpired , err)```

File: pkg/dns/service.go [hunk 10]
```diff
@@ -159,14 +174,16 @@ func (s *Service) SetEndpoints(ctx context.Context, addresses []gatewayv1beta1.G
 	old := dnsRecord.DeepCopy()
 	host := dnsRecord.Name
 
-	// check if endpoint already exists in the DNSRecord
+	// check if endpoint already exists in the DNSRecord - we cannot delete and rebuild as health points are injected into
+	// these endpoints elsewhere - TODO refactor this to be all in one place.
```
maleck13, line 178: do you want to make sure there is an issue for this?

pkg/dns/service.go
```diff
@@ -231,6 +243,11 @@ func FindMatchingManagedZone(originalHost, host string, zones []v1alpha1.Managed
 		return nil, "", fmt.Errorf("no valid zone found for host: %v", originalHost)
 	}
 
+	hostParts := strings.SplitN(host, ".", 2)
```
maleck13, line 246: not clear why we added this? Doesn't seem to be used?

maleck13 (commented):
few nit comments generally looks good. Will try it locally

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -61,6 +63,7 @@ type HostService interface {
 	GetManagedZoneForHost(ctx context.Context, domain string, t traffic.Interface) (*v1alpha1.ManagedZone, string, error)
 	SetEndpoints(ctx context.Context, endpoints []gatewayv1beta1.GatewayAddress, dnsRecord *v1alpha1.DNSRecord, dnsPolicy *v1alpha1.DNSPolicy) error
 	CleanupDNSRecords(ctx context.Context, owner traffic.Interface) error
+	GetDNSRecordsFor(ctx context.Context, t traffic.Interface) ([]*v1alpha1.DNSRecord, error)
```
mikenairn, line 66: I assume this change is no longer using this? I guess it's not a big deal to leave it, might be useful in the future. Depends if we want unused code left hanging about.

pkg/dns/service.go
mikenairn: I'd guess this might fall under the work for health checks.

pkg/dns/service.go
philbrookes: The original logic was outputting a pretty unuseful error message to debug the issue (no matching managedzone found), so I reworked it a little to improve the error message.

pkg/controllers/gateway/gateway_controller.go
philbrookes: oh yeah, I'll remove it

pkg/dns/service.go
philbrookes: https://github.com/Kuadrant/multicluster-gateway-controller/issues/294

maleck13 (approved):

PR #293, opened 2023-07-03
Geo DNS Strategy

mikenairn:
Adds a LoadBalancing section to the DNSPolcy spec allowing default weight and Geo code to be set.

`LoadBalancing.Weighted.DefaultWeight` is required (default 120), and if set will be used as the weight for a target clusters ednpoints if no other weight can be determined from cluster attributes. 
`LoadBalancing.Geo.DefaultGeo` is optional, and when set indicates the default geo to use for target cluster endpoints if no other geo can be determined from cluster attributes.

The generated DNSRecord structure has been updated to accommodate Geo and Weighted only strategies using a combination of CNAMEs and A records.

see https://github.com/Kuadrant/multicluster-gateway-controller/pull/270


pkg/dns/service.go
```diff
@@ -24,9 +24,10 @@ import (
 )
 
 const (
-	LabelRecordID          = "kuadrant.io/record-id"
-	LabelGatewayReference  = "kuadrant.io/gateway-uid"
-	ProviderSpecificWeight = "weight"
+	LabelRecordID                         = "kuadrant.io/record-id"
+	LabelGatewayReference                 = "kuadrant.io/Gateway-uid"
+	DefaultCnameTTL          v1alpha1.TTL = 300
+	DefaultClusterGatewayTTL v1alpha1.TTL = 60
```
maleck13, line 30: this makes sense based on what we discussed earlier 👍

pkg/dns/service.go
```diff
@@ -149,53 +130,78 @@ func (s *Service) GetDNSRecord(ctx context.Context, subDomain string, managedZon
...
 
-// SetEndpoints sets the endpoints for the given DNSRecord for each of the gateway addresses.
-func (s *Service) SetEndpoints(ctx context.Context, addresses []gatewayv1beta1.GatewayAddress, dnsRecord *v1alpha1.DNSRecord, dnsPolicy *v1alpha1.DNSPolicy) error {
-	//dnsPolicy is not currently used, but will be once we add other routing strategies
-	ips, err := s.resolveIPS(ctx, addresses)
-	if err != nil {
-		return err
-	}
+// SetEndpoints sets the endpoints for the given MultiClusterGatewayTarget
+func (s *Service) SetEndpoints(ctx context.Context, mcgTarget *MultiClusterGatewayTarget, dnsRecord *v1alpha1.DNSRecord) error {
```
maleck13, line 134: this is quite a complex function now. might be worth adding a bit more description on what it is doing?

pkg/dns/target.go
```diff
@@ -1,11 +1,186 @@
...
+	t.Weight = &defaultWeight
+}
+
+func toBase36hash(s string) string {
+	hash := sha256.Sum224([]byte(s))
+	// convert the hash to base36 (alphanumeric) to decrease collision probabilities
+	base36hash := strings.ToLower(base36.EncodeBytes(hash[:]))
+	// use 6 chars of the base36hash, should be enough to avoid collisions and keep the code short enough
+	return base36hash[:6]
 }
```
mikenairn, line 186: KCP were you using this to generate workspace and logical cluster names. Assuming they knew better than me what they were doing i think this should be sufficient, it may even be possible to reduce the length further,  4 maybe?

maleck13 (commented):
This is looking pretty good to me. The set endpoints is a bit complex to follow. And as you mentioned triggering the DNSPolicy reconcile when a MC changes is going to be a tricky mapping. 

Is it ready for me to give a try?

mikenairn:
> Is it ready for me to give a try?

Yes, this can be tested now, I'll add some notes to the description

pkg/apis/v1alpha1/dnspolicy_types.go
```diff
@@ -33,6 +33,36 @@ type DNSPolicySpec struct {
...
+}
+
+type LoadBalancingGeo struct {
+	// default is the country or continent code to use when no other can be determined for a dns target cluster.
+	//
+	// 2 digit iso3166 country code (https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) or continent code:
+	// AF: Africa; AN: Antarctica; AS: Asia; EU: Europe; OC: Oceania; NA: North America; SA: South America
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=2
+	Default string `json:"default,omitempty"`
```
mikenairn, line 65: @maleck13 Based on yesterdays call do we want to rename these? Looking at the [docs](https://gateway-api.sigs.k8s.io/references/policy-attachment/#inherited-policy-attachment-its-all-about-the-defaults-and-overrides) for the inherited policy attachment i can see where confusion might come in, and maybe just changing it now would be a good idea. 

Maybe `defaultWeight` and `defaultGeo`? It is a little more obvious what it is this way, `default` is not the most descriptive.

/cc @thomasmaas @guicassolato

maleck13: yeah I think changing them now is reasonable. There could come a point where we want to use inherited rather than direct policy attachment

pkg/apis/v1alpha1/dnspolicy_types.go
```diff
@@ -33,6 +33,36 @@ type DNSPolicySpec struct {
...
+	// +optional
+	Geo *LoadBalancingGeo `json:"geo,omitempty"`
+}
+
+type LoadBalancingWeighted struct {
+	// defaultWeight is the record weight to use when no other can be determined for a dns target cluster.
+	//
+	// +kubebuilder:default=120
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=120
```
mikenairn, line 54: Max value supported by aws is 255 https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-values-weighted.html#rrsets-values-weighted-weight, google supports higher values up to 10000 but we probably want to keep this limited to something all providers will support, so this should probably be a max value of 255.

mikenairn: Azure has  a max of 1000

pkg/dns/service.go
mikenairn: Updated, hopefully it makes sense

pkg/controllers/dnspolicy/cluster_eventmapper.go
```diff
@@ -0,0 +1,60 @@
...
+	allGwList := &gatewayapiv1beta1.GatewayList{}
+	err := m.client.List(context.TODO(), allGwList)
+	if err != nil {
+		logger.Info("mapToPolicyRequest:", "error", "failed to get gateways")
+		return []reconcile.Request{}
+	}
+
+	requests := make([]reconcile.Request, 0)
+	for _, gw := range allGwList.Items {
+		val := metadata.GetAnnotation(&gw, gateway.GatewayClustersAnnotation)
```
maleck13, line 47: this is fine, but I also think the clusters are recorded in a status condition

pkg/controllers/dnspolicy/cluster_eventmapper.go
```diff
@@ -0,0 +1,60 @@
...
+	return m.mapToPolicyRequest(obj, "dnspolicy", &DNSPolicyRefsConfig{})
+}
+
+func (m *ClusterEventMapper) mapToPolicyRequest(obj client.Object, policyKind string, policyRefsConfig common.PolicyRefsConfig) []reconcile.Request {
+	logger := m.Logger.V(1).WithValues("object", client.ObjectKeyFromObject(obj))
+
+	clusterName := obj.GetName()
+
+	allGwList := &gatewayapiv1beta1.GatewayList{}
+	err := m.client.List(context.TODO(), allGwList)
```
maleck13, line 39: not a big fan of this context.TODO but not sure what the alternative is.

pkg/controllers/dnspolicy/cluster_eventmapper.go
maleck13:
```
lastTransitionTime: "2023-07-06T10:11:39Z"
    message: 'gateway placed on clusters [kind-mgc-control-plane] '
    observedGeneration: 1
    reason: Programmed
    status: "True"
    type: Programmed
```

mikenairn: They are, but tbh i think we should consider removing that. Parsing a list of clusters out of that string is not a good way to do it, and it actually makes setting that condition harder as sometimes we want a list of clusters, and sometimes we don't. see https://github.com/Kuadrant/multicluster-gateway-controller/pull/298#discussion_r1253378527

pkg/dns/iso3166.go
```diff
@@ -0,0 +1,268 @@
...
+	{548, "Vanuatu", "VU", "VUT"},
+	{862, "Venezuela (Bolivarian Republic of)", "VE", "VEN"},
+	{704, "Viet Nam", "VN", "VNM"},
+	{92, "Virgin Islands (British)", "VG", "VGB"},
+	{850, "Virgin Islands (U.S.)", "VI", "VIR"},
+	{876, "Wallis and Futuna", "WF", "WLF"},
+	{732, "Western Sahara*", "EH", "ESH"},
+	{887, "Yemen", "YE", "YEM"},
+	{894, "Zambia", "ZM", "ZMB"},
+	{716, "Zimbabwe", "ZW", "ZWE"},
```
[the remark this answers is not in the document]
mikenairn: Yeah if there is something out there that looks well maintained we should probably use that, will take a look.

maleck13 (approved):
Tested this locally was able to set up a DNSPolicy set a default GEO and set geo attributes on my clusters and see those geos used in my DNS records. 
Happy for this to merge

maleck13 (approved):
new addition of continent code makes sense

pkg/dns/iso3166.go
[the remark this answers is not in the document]
jasonmadigan: yeah sounds good, can revisit this later when we look at other providers and see what standards they're looking for in terms of API for country codes. That lib looks pretty comprehensive, so I'd guess it'll suit most. But we can revisit later.

mikenairn:
/hold cancel

@maleck13 I'm happy that everything that needs done for this is now done. There are some improvements that can be made, but I'm going to create some follow on issues for those.

maleck13 (approved):

PR #298, opened 2023-07-05
Fix gateway status updates

mikenairn:
Fixes an issue with gateway status updates where it could end up in a loop of constantly updating. This can be seen when placing gateways on more than one cluster.

* Orders the cluster list so that the status condition update always prints the same message for the clusters the gateway is placed on. 
* Use the existing accepted condition if it already exists, currently causes an update on every reconcile since `lastTransaction` is going to be different on a new condition.
* Replaces most of the condition helpers with the meta pkg equivalent helpers.


pkg/_internal/conditions/conditions.go
mikenairn, line 1: We should probably just get rid f this helper and use `k8s.io/apimachinery/pkg/api/meta` everywhere we normally use this.

maleck13: agree with removing this file

maleck13 (commented):
change looks fine. Do we want to remove the conditions.go as part or merge as is? I am happy with either if we have an issue to track

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -147,7 +147,14 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 	if err != nil && IsInvalidParamsError(err) {
-		conditions.SetCondition(upstreamGateway.Status.Conditions, previous.Generation, string(gatewayv1beta1.GatewayConditionProgrammed), metav1.ConditionFalse, string(gatewayv1beta1.GatewayReasonPending), fmt.Sprintf("Invalid parameters in gateway class: %s", err.Error()))
+		programmedCondition := metav1.Condition{
+			Type:               string(gatewayv1beta1.GatewayConditionProgrammed),
+			Status:             metav1.ConditionFalse,
+			Reason:             string(gatewayv1beta1.GatewayReasonPending),
+			Message:            fmt.Sprintf("Invalid parameters in gateway class: %s", err.Error()),
+			ObservedGeneration: previous.Generation,
+		}
+		meta.SetStatusCondition(&upstreamGateway.Status.Conditions, programmedCondition)
```
mikenairn, line 157: Would probably be better if we could just use `buildProgrammedCondition` from wherever we want to set the programmed condition and have it return the expected condition based on the error.

pkg/_internal/conditions/conditions.go
mikenairn: I've updated everything to use the meta pkg. I haven't removed this file as it has a constant `ConditionTypeReady` being used in a few places and I'm not sure of a better place for it.

maleck13 (approved):

PR #309, opened 2023-07-06
Add e2e test to ci

roivaz:
Adds new job to the CI workflow to run the e2e tests against the local-setup environment. I also updated the go-setup to action to v4 as later versions have built-in caching.
The e2e suite takes a while to complete, something in the order of 12-14 minutes, with most of the time being spent on the local-setup script itself and the tests taking just about a couple of minutes. We might need to consider some improvements to the local-setup script to try and reduce this time (hiding the installation of some things like argocd and k8s-dashboard behind flags, some parallelization, etc).

### DIFF
--- a/.github/workflows/ci-e2e.yaml
+++ b/.github/workflows/ci-e2e.yaml
@@ -19,10 +19,18 @@ on:
       - 'LICENSE'
 
 jobs:
+  event_payload:
+    runs-on: ubuntu-latest
+    steps:
+      - run: cat $GITHUB_EVENT_PATH
   authorize:
+    # see https://docs.github.com/en/graphql/reference/enums#commentauthorassociation
+    # for info on author association
     environment:
       ${{ github.event_name == 'pull_request_target' &&
-      github.event.pull_request.head.repo.full_name != github.repository &&
+      github.event.pull_request.author_association != 'OWNER' &&
+      github.event.pull_request.author_association != 'MEMBER' &&
+      github.event.pull_request.author_association != 'COLLABORATOR' &&
       'external' || 'internal' }}
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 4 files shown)


david-martin:
> hiding the installation of some things like argocd and k8s-dashboard behind flags, some parallelization, etc

Agreed.
I think we could default to not installing them, similar to submariner, so it's the absolute minimum for local dev and testing.
Any thoughts on how to do the parallelisation?

roivaz:
For parallelization I haven't given it much thought, but most of the things that are installed could be launched at the same time on the cluster. 
We could try just with bash... at the risk of making it too complex to debug. 
We could also make a go wrapper that can execute the different bash functions in parallel, it would be probably easier to coordinate the different threads and report back to the user. 
Also ansible might be an option here, creating ansible roles that install each component and letting them run in parallel.
I'm not sure at this point about what would be the most maintainable solution tbh.

roivaz:
/hold 
the job is failing to retrieve the repository secrets, it must be something very obvious but I'm not seeing it right now ...

roivaz:
I'm closing this temporarily to avoid more workflow runs while I test things in my fork

/close

roivaz:
@david-martin ended up implementing this https://iterative.ai/blog/testing-external-contributions-using-github-actions-secrets, it's the solution that has less impact in how the team works. It's basically using `pull_request_target` event for the e2e actions so it runs in the context of the base repository. An authorize step that makes use of the "environments" feature in GH is required when the PR is opened from a fork and will check if the user opening the PR is a OWNER, MEMBER or COLLABORATOR in the repo. If it is not one of those, a member must authorize the "deployment" for the e2e tests to run.

You can try and open a PR against my fork (https://github.com/roivaz/multicluster-gateway-controller), it should request authorization to run the e2e tests as you are not a member of my personal org.

david-martin:
/lgtm
/approve

Tested with a PR to a fork.
We can test (and fix if needed) once on main.

PR #312, opened 2023-07-07
Add custom weights to DNSPolicy

### LINKED ISSUE #284: Custom Weighting DNS based load balancing policy
**What**

We currently only support Round Robin / Even weighted load balancing. We want to support the ability to define custom weighting based load balancing based on attributes of the underlying clusters.

**Use Case**

```
As a gateway admin, I want to send 10% of traffic to my AWS cloud and 90% to my google cloud resources as I have cheaper costs in Google but still have committed spend in AWS
````

The proposed DNSPolicy resource has a weighted section. It is this that we need to implement

 weighted: # always required
     default: 10  #always required
     custom: #optional
     - value: AWS  #optional with both GEO and weighted. With GEO the custom weight is applied to gateways within a GEOgraphic region
       weight: 30
     - value: GCP
       weight: 20

- All records by default will receive the default weight
- when a custom weight is specified it will be applied to records associated with IPS/Hosts from clusters that have the a `kuadrant.io/lb-attribute-custom-weight:` label with the specified value.

https://github.com/Kuadrant/multicluster-gateway-controller/pull/270/files

The DNSRecord structure is covered under the above proposal also

### PR DESCRIPTION
**Pull request by mikenairn**
Adds an option to the `DNSPolicy.LoadBalancing.Weighted` spec to add custom weights for attributes set on target clusters via labels. The custom attributes are checked during the calculation of the target clusters weight, and if attributes match, the custom weight is applied.

Example

```
spec:
  loadBalancing:
    weighted:
      custom:
      - value: AWS
        weight: 60
      - value: GCP
        weight: 255
      defaultWeight: 120
```
```
kubectl label --overwrite managedcluster cluster-1 kuadrant.io/lb-attribute-custom-weight=AWS
kubectl label --overwrite managedcluster cluster-2 kuadrant.io/lb-attribute-custom-weight=GCP
kubectl label --overwrite managedcluster cluster-3 kuadrant.io/lb-attribute-custom-weight-
```

The above would favour the GCP cluster(cluster-2) over any other, and the AWS cluster(cluster-1) would recieve the least traffic with cluster-3 falling back to the default weight getting a medium amount of traffic.

closes #284

### CODE REVIEW

File: pkg/dns/target.go [hunk 1]
```diff
@@ -86,7 +85,8 @@ type ClusterGateway struct {
 }
 
 type ClusterAttributes struct {
-	Geo *GeoCode
+	CustomWeight *string
```
maleck13, line 88: I like that we are doing more of this with optional stuff it is a pattern you often see in other k8s controllers and APIs +1

maleck13 (commented):
changes look good, will give it a spin locally

maleck13:
built up a dnspolicy
```
kind: DNSPolicy
metadata:
  creationTimestamp: "2023-07-10T11:21:55Z"
  finalizers:
  - kuadrant.io/dns-policy
  generation: 4
  name: prod-web
  resourceVersion: "11685"
  uid: e3d3cb98-18cf-4ff8-a97a-19228f2995d8
spec:
  loadBalancing:
    geo:
      defaultGeo: C-AN
    weighted:
      custom:
      - value: AWS
        weight: 60
      - value: GCP
        weight: 180
      defaultWeight: 120
  targetRef:
    kind: Gateway
    name: prod-web
status:
  conditions:
  - lastTransitionTime: "2023-07-10T11:49:19Z"
    message: Gateway is DNS Enabled
    reason: GatewayDNSEnabled
    status: "True"
    type: Ready
  observedGeneration: 4
```
and got the expected setup in AWS rout53
Screenshot 2023-07-10 at 12 51 44

maleck13 (approved):

PR #315, opened 2023-07-07
Implement the Gateway status aggregation proposal for addresses and l…

### LINKED ISSUE #113: Control Plane Gateway status

david-martin:
…isteners

Closes #113 

The [proposal](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/docs/proposals/status-aggregation.md#addresses-and-listeners-status).

The gateway integration tests are updated to test address (both IPAddress and HostnameAddress types) and listener aggregation.
However, there is mocking of the ManifestWork status (as ocm doesn't run in integration tests), so I'd advice verification of the ocm walkthrough as well locally.
After running through the whole walkthrough, the hub gateway status should look something like this:

```yaml
status:
  addresses:
    value: kind-mgc-control-plane/172.32.200.0
    value: kind-mgc-workload-1/172.32.201.0
  conditions:
  - lastTransitionTime: "2023-07-07T15:48:37Z"
    message: Handled by kuadrant.io/mgc-gw-controller
    observedGeneration: 1
    reason: Accepted
    status: "True"
    type: Accepted
  - lastTransitionTime: "2023-07-07T15:53:49Z"
    message: 'gateway placed on clusters [kind-mgc-control-plane kind-mgc-workload-1] '
    observedGeneration: 1
    reason: Programmed
    status: "True"
    type: Programmed
  listeners:
  - attachedRoutes: 1
    conditions: []
    name: kind-mgc-control-plane.api
    supportedKinds: []
  - attachedRoutes: 1
    conditions: []
    name: kind-mgc-workload-1.api
    supportedKinds: []
```

Note that listener status `conditions` and `supportedKinds` are required fields, but have no implementation (or proposed implementation) at this time. 
They will always be an empty list.

### DIFF
--- a/pkg/controllers/gateway/gateway_controller.go
+++ b/pkg/controllers/gateway/gateway_controller.go
@@ -55,10 +55,12 @@ import (
 )
 
 const (
-	GatewayClusterLabelSelectorAnnotation = "kuadrant.io/gateway-cluster-label-selector"
-	GatewayClustersAnnotation             = "kuadrant.io/gateway-clusters"
-	GatewayFinalizer                      = "kuadrant.io/gateway"
-	ManagedLabel                          = "kuadarant.io/managed"
+	GatewayClusterLabelSelectorAnnotation                            = "kuadrant.io/gateway-cluster-label-selector"
+	GatewayClustersAnnotation                                        = "kuadrant.io/gateway-clusters"
+	GatewayFinalizer                                                 = "kuadrant.io/gateway"
+	ManagedLabel                                                     = "kuadarant.io/managed"
+	MultiClusterIPAddressType             gatewayv1beta1.AddressType = "kuadrant.io/MultiClusterIPAddress"
+	MultiClusterHostnameAddressType       gatewayv1beta1.AddressType = "kuadrant.io/MultiClusterHostnameAddress"
 )
 
 type HostService interface {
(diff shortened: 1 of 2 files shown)


pkg/controllers/gateway/gateway_controller.go
```diff
@@ -199,6 +201,57 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	for _, listener := range specListeners {
+		for _, cluster := range clusters {
+			attachedRoutes, err := r.Placement.ListenerTotalAttachedRoutes(ctx, upstreamGateway, string(listener.Name), cluster)
+			if err != nil {
+				// May not have the status yet, let's ignore, but output info logs about it
+				log.Info("AttachedRoutes unknown for listener. Ignoring", "listener", listener.Name, "cluster", cluster, "message", err)
+				continue
+			}
+			allListenerStatuses = append(allListenerStatuses, gatewayv1beta1.ListenerStatus{
+				Name:           gatewayv1beta1.SectionName(fmt.Sprintf("%s.%s", cluster, string(listener.Name))),
```
mikenairn, line 246: I think will need to be able to parse the cluster and listener name from this value, do we think just doing that based on a period delimiter will be sufficient? I assume you could create clusters or listeners with periods in the names which could cause issues.

pkg/controllers/gateway/suite_test.go
```diff
@@ -18,10 +18,12 @@ package gateway
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"path/filepath"
 	"testing"
 	"time"
 
+	. "github.com/Kuadrant/multicluster-gateway-controller/test/util"
```
mikenairn, line 26: nit: should be in the group with the rest of the mgc imports

pkg/controllers/gateway/gateway_controller.go
david-martin: A [Listener SectionName](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.SectionName) can be a valid subdomain, so it could contain more than 1 period. The validation doesn't allow slashes or other symbols.

The [cluster name check](https://github.com/open-cluster-management-io/clusteradm/blob/main/pkg/cmd/join/preflight/checks.go#L97-L104) doesn't allow periods, slashes, underscores or other symbols except just hyphens.

Given we are bound by SectionName validation first, and the fact periods are not allowed in cluster names, I think having a period separator is reasonable.
To parse a multi cluster SectionName, you can split on the first period.
Can you think of anything we're missing here, or if there's a better way to handle this?

david-martin: > One caveat here is that cluster names with a hyphen in them will not pass the SectionName validation.
perhaps we need to allow for this edge case.

Despite this, that validation doesn't seem to be baked into the CRD validation, so it worked in my testing.
However it doesn't comply with the spec...

david-martin: Here's the validation pattern of a listener name in the crd

```
^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
```

> The validation doesn't allow slashes or other symbols

I was incorrect.
The validation allows hyphens

david-martin: To summarise the above, you can parse out the cluster name and listener name by splitting on the first period.
There may be more than 1 period, but that shouldn't be a problem if splitting stops at the first.

e.g. 

- mycluster.listenername => `mycluster` & `listenername`
- mycluster.listener.name => `mycluster` & `listener.name`
- my-cluster.listener.name => `my-cluster` & `listener.name`
- my-cluster.listener-name.org => `my-cluster` & `listener-name.org`
- my.cluster.listener.name => Invalid cluster name. Can't have periods. If this somehow happened, the cluster name would be assumed to be `my` with the listener name being `cluster.listener.name`


@mikenairn Given the above, are there any scenarios that are not allowed for?

mikenairn: Yes, i think this is fine. I was just wondering when i seen it what issues we might hit. Seems though we should be fine as long as clusters stick to valid cluster names.

mikenairn:
Tried this locally and it worked as expected. If you rebase main the tests should pass.

PR #319, opened 2023-07-10
Proposal health checks

philbrookes:


docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,81 @@
...
+apiVersion: kuadrant.io/v1alpha1
+kind: DNSHealthCheckProbe
+metadata:
+  name: example-probe
+spec:
+  port: "..."
+  host: “...”
+  ipAddress: "..."
+  path: "..."
+  protocol: "..."
```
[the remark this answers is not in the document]
philbrookes: The host and IP would both be required, the host is passed in as the Host header in a requests against the IP address. 

I suppose it could be compiled into a URL, but the different parts are coming from different places, i.e. the DNS Policy and the DNS Record.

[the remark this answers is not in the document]
jasonmadigan: I think I get you, thanks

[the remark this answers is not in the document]
mikenairn: re. https://github.com/Kuadrant/multicluster-gateway-controller/pull/240/files#r1252649712

Should it be address then instead of  `ipAddress` and possibly have a type like a gateway address. It's not always going to be an IP Address now, it could be a an external Lb Hostname.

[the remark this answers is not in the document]
mikenairn: Might be worth adding a description of what these fields are to this doc, wasn't totally clear until Jason mentioned it.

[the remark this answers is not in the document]
philbrookes: I've made those changes now. Thanks.

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,89 @@
...
+
+The MGC has the ability to extend the DNS configuration of the gateway with the DNSPolicy resource. This resource allows 
+users to configure health checks. As a result of configuring health checks, the controller creates the health checks in 
+Route53, attaching them to the related DNS records. This has the benefit of automatically disabling an endpoint if it 
+becomes unhealthy, and enabling it again when it becomes healthy again.
+
+This feature has a few shortfalls:
+1. It’s tightly coupled with Route53. If other DNS providers are supported they must either provide a similar feature, 
+or health checks will not be supported
+2. Lacks the ability to reach endpoints in private networks
```
maleck13, line 13: I would add, requires us the gateway controller to implement, maintain and test multiple providers

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,89 @@
...
+* Transparently keep support for other health check providers like Route53
+
+## Proposal
+
+#### `DNSPolicy` resource
+
+The presence of the `healthCheck` means that for every DNS endpoint (that is either an A record, or a CNAME to an external host), 
+a health check is created based on the health check configuration in the DNSPolicy.
+
+A `failureThreshold` field will be added to the health spec, allowing users to configure a number of consecutive health
```
maleck13, line 34: I think it would be worth adding an example here of the policy definition

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,89 @@
...
+The presence of the `healthCheck` means that for every DNS endpoint (that is either an A record, or a CNAME to an external host), 
+a health check is created based on the health check configuration in the DNSPolicy.
+
+A `failureThreshold` field will be added to the health spec, allowing users to configure a number of consecutive health 
+check failures that must be observed before the endpoint is considered unhealthy.
+
+#### `DNSHealthCheckProbe` resource
+
+The DNSHealthCheckProbe resource configures a health probe in the controller to perform the health checks against an 
+endpoint.
```
maleck13, line 41:
```suggestion
identified final A or CNAME record
```

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,89 @@
...
+metadata:
+  name: example-probe
+spec:
+  port: "..."
+  host: “...”
+  address: "..."
+  path: "..."
+  protocol: "..."
+  interval: "..."
+status:
```
maleck13, line 55: I feel like this status should reflect which policy triggered its creation

maleck13: also should we have a way to identify which gateway and listener this probe is for?

docs/proposals/provider-agnostic-dns-health-checks.md
maleck13: also wondering as part of defining a health check if we should have an additional headers map so if some custom header needs to be sent it can?

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,89 @@
...
+The DNSHealthCheckProbe resource configures a health probe in the controller to perform the health checks against an 
+endpoint.
+
+```yaml
+apiVersion: kuadrant.io/v1alpha1
+kind: DNSHealthCheckProbe
+metadata:
+  name: example-probe
+spec:
+  port: "..."
```
maleck13, line 49: if we allow a headers map then it would populate here as well

docs/proposals/provider-agnostic-dns-health-checks.md
maleck13: for example a service account token

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,89 @@
...
+- **Address** The address to connect to (e.g. IP address or hostname of a clusters loadbalancer)
+- **Host** The host to request in the Host header
+- **Path** The path to request
+- **Protocol** The protocol to use for this request
+- **Interval** How frequently this check would ideally be executed.
+
+The reconciliation of this resource results in the configuration of a health probe, which targets the endpoint and 
+updates the status. The status is propagated to the providerSpecific status of the equivalent endpoint in the DNSRecord
+
+### Changes to current controllers
```
maleck13, line 73: do we not want to outline or call out that there will be a new health check agent? Seems like the fundamental proposal here is to introduce a new health check agent component?

maleck13: I think we should also at a high level explain how the agent will perform the health checks and where the agent will live

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,89 @@
...
+
+In order to support this new feature, the following changes in the behaviour of the controllers are proposed.
+
+#### DNSPolicy controller
+
+Currently, the reconciliation loop of this controller creates health checks in the configured DNS provider 
+(Route53 currently) based on the spec of the DNSPolicy, separately from the reconciliation of the DNSRecords. 
+The proposed change is to reconcile health check CRs based on the combination of DNS Records and DNS Policies.
```
maleck13, line 81:
```suggestion
The proposed change is to reconcile health check probe CRs based on the combination of DNS Records and DNS Policies.
```

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,89 @@
...
+#### Spec Fields Definition
+- **Port** The port to use
+- **Address** The address to connect to (e.g. IP address or hostname of a clusters loadbalancer)
+- **Host** The host to request in the Host header
+- **Path** The path to request
+- **Protocol** The protocol to use for this request
+- **Interval** How frequently this check would ideally be executed.
+
+The reconciliation of this resource results in the configuration of a health probe, which targets the endpoint and 
+updates the status. The status is propagated to the providerSpecific status of the equivalent endpoint in the DNSRecord
```
maleck13, line 71: is the health probe the agent? or part of the agent? not clear

maleck13: which component is propagating the status back to the DNSRecord?

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,89 @@
...
+(Route53 currently) based on the spec of the DNSPolicy, separately from the reconciliation of the DNSRecords. 
+The proposed change is to reconcile health check CRs based on the combination of DNS Records and DNS Policies.
+
+Instead of Route53 health checks, the controller will create `DNSHealthCheckProbe` resources.
+
+#### DNSRecord controller
+
+When reconciling a DNS Record, the DNS Record reconciler will retrieve the relevant DNSHealthCheckProbe CRs, and consult
+the status of them when determining what value to assign to a particular endpoint's weight. If the HealthCheckProbe is 
+reporting an unhealthy response, then the weight will be assigned 0, otherwise the usual weight assigning process will
```
maleck13, line 89: do we want to call out what the plan is if all endpoints are unhealthy.

maleck13: although it would mean you don't need to have the endpoint then. As in if the health check is returning 404s then we will just treat it as though there is no health check and return all endpoints. So I guess it depends on what we see as the most valuabe. IE do you want to force the health check to be present or just take advantage of a health check if it is present for failover. 
I can see arguments for both, but you could certainly say, "if you want to take take advantage of automatic endpoint failure mitigation add the health check otherwise your on your own"

docs/proposals/provider-agnostic-dns-health-checks.md
philbrookes: yes, I have actually already written the code this way, but I will update the proposal to reflect that.

docs/proposals/provider-agnostic-dns-health-checks.md
philbrookes: There is no "seperate" agent now, right? Or I've significantly misunderstood our most recent discussion. The code is running inside the existing controller, rate limited by a queue and running on a separate goroutine.

docs/proposals/provider-agnostic-dns-health-checks.md
philbrookes: The status is written back to the health check probe CR by the probe worker which is executing the probe from a queue.

docs/proposals/provider-agnostic-dns-health-checks.md
philbrookes: yes ok, I will outline the proposed response to the various scenarios.

docs/proposals/provider-agnostic-dns-health-checks.md
maleck13: ok I guess a separate service is not needed at this point. But I can see the potential need for it in the future if we want to install into the spoke or scale independently. Really this means adding a separate main and coming up with a mechanism for deploying it. So as they can share the same code base, initially having the health checks start as part of the main controller seems fine for the initial release. Lets capture this in the proposal though

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,140 @@
...
+    endpoint: /health
+    failureThreshold: 5
+    port: 443
+    protocol: https
+    additionalHeaders:
+      - name: "..."
+        value: "..."
+      - name: "..."
+        value: "..."
+  targetRef:
```
david-martin, line 59: I'm wondering how health check targets an entire Gateway.
I would have thought per listener/httproute would be align with an application specific health endpoint.

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,140 @@
...
+
+In order to support this new feature, the following changes in the behaviour of the controllers are proposed.
+
+#### DNSPolicy controller
+
+Currently, the reconciliation loop of this controller creates health checks in the configured DNS provider 
+(Route53 currently) based on the spec of the DNSPolicy, separately from the reconciliation of the DNSRecords. 
+The proposed change is to reconcile health check probe CRs based on the combination of DNS Records and DNS Policies.
+
+Instead of Route53 health checks, the controller will create `DNSHealthCheckProbe` resources.
```
david-martin, line 117: Is it expected a user (gateway admin or application developer) would create a DNSHealthCheckProbe?

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,140 @@
...
+
+Example DNS Policy with a defined health check.
+```yaml
+apiVersion: kuadrant.io/v1alpha1
+kind: DNSPolicy
+metadata:
+  name: prod-web
+  namespace: multi-cluster-gateways
+spec:
+  healthCheck:
```
david-martin, line 49: Is this intended to be an abstraction of the DNSHealthCheckProbe resource?

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,140 @@
...
+The DNSHealthCheckProbe resource configures a health probe in the controller to perform the health checks against an 
+identified final A or CNAME endpoint. When created by the controller as a result of a DNS Policy, this will have an 
+owner ref of the DNS Policy that caused it to be created.
+
+```yaml
+apiVersion: kuadrant.io/v1alpha1
+kind: DNSHealthCheckProbe
+metadata:
+  name: example-probe
+spec:
```
david-martin, line 76: Some of the fields here look similar to [kubernetes readiness & liveness probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/).
Is that the inspiration for these fields, or some other standard or service?
e.g. https://docs.datadoghq.com/integrations/dns_check/ or https://documentation.solarwinds.com/en/success_center/pingdom/content/topics/what-is-a-check-.htm

docs/proposals/provider-agnostic-dns-health-checks.md
maleck13: it will check each listener on that gateway. @philbrookes  perhaps we should allow specifying what response codes are ok IE perhaps they are ok with 404 but not a 503

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,140 @@
...
+This document describes a proposal to extend the current health check implementation to overcome these shortfalls.
+
+### Goals
+
+* Ability to configure health checks in the DNSPolicy associated to a Gateway
+* DNS records are disabled when the associated health check fails
+* Current status of the defined health checks is visible to the end user
+
+### Nongoals
+* Ability for the health checks to reach endpoints in private networks
```
maleck13, line 25: in separate private networks? If the hub and spoke are in the same private network, would work fine

maleck13: also a none goal here should be, having gateway level health checks cover wildcard listeners

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,140 @@
...
+* Transparently keep support for other health check providers like Route53
+
```
maleck13, line 27: I think adding some use cases here would be good also:

```
As a gateway administrator, I would like to define a health check that each service sitting behind a particular listener across the production clusters has to implement to ensure we can automatically respond, failover and mitigate a failing instance of the service
```
@david-martin ^ FYI

docs/proposals/provider-agnostic-dns-health-checks.md
philbrookes: no, these are created by the controller as a result of the settings in a dns policy.

docs/proposals/provider-agnostic-dns-health-checks.md
philbrookes: it's more like the template for the settings for all the health checks created as a result of this DNS Policy.

docs/proposals/provider-agnostic-dns-health-checks.md
philbrookes: yes agreed with both points, I will add the non-goal.

docs/proposals/provider-agnostic-dns-health-checks.md
philbrookes: @sergioifg94 wrote the original spec for these, I am not sure where the inspiration came from.

docs/proposals/provider-agnostic-dns-health-checks.md
david-martin: I think that's a reasonable thing a gateway administrator would like to have.

In reality, I wonder if there's a different use case that's needed as well?
Will our users be building new apps (or think it reasonable to extend them all to conform to a health check API)?
Or will there be a lot of 'lift and shift' of existing apps. Perhaps there are existing apps that are not moving, but instead are being routed to via vpns, submariner, skupper etc.. from a central gateway.
In these cases, would it make more sense to allow per listener/host health checks to be defined based on the app?

maleck13: I think allowing policy to target listeners is something we will want. This is not something that policy attachment supports in its targeting yet though. I think this is something we should discuss with them using use cases like this. Something else I have being given thought to is whether we could allow a policy to be defined in the spoke that is leveraged by the control plane (IE health set on the HTTPRoute rather than the gateway for some use cases) This feels  complex though and is something we would want to give careful thought. Esp when it comes to making sense of what is happening

mikenairn: Does direct policy attachment dictate that you can only have one policy attached to a target? I know the kuadrant policies do this, but is that a requirement? I was wondering if we could have multiple polices attached to a single gateway (direct attachment), but specify what hosts it should be applied to in a similar way to the RateLimit and AuthConfig policies https://github.com/Kuadrant/kuadrant-operator/blob/main/api/v1beta1/ratelimitpolicy_types.go#L103 and https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/ratelimitpolicy_controller.go#L161. We don't do any of this in the DNSPolicy currently and just accept a policy will act against any listener, but it could be scoped. Problem with that and multiple polices would be you'd only want one policy acting on any given listener.

philbrookes: > would it make more sense to allow per listener/host health checks to be defined based on the app

I don't this host / listener specific details belong in a gateway policy. I have always considered policies to be the abstract definition of some rule that can be applied to multiple gateways. I feel that specifying particular hosts in a policy contradicts that philosophy. 

i.e. a policy with specifically defined hosts is only really useful to one particular gateway, which undermines the purpose of it being a policy at all.

david-martin: > I don't think host / listener specific details belong in a gateway policy.

While I don't disagree with this (the scope of a policy targeting a gateway), I am questioning the usefulness of defining a health check that applies to all listeners vs. the ability to define a health check per listener/host/application.

philbrookes: I see, this concept is like a... health check for a specific route or endpoint, which I suppose would be a different type of object to a DNSHealthCheckPolicy. I see value in that alright, but it probably warrants it's own issue / proposal separate to this one?

maleck13: Yes different issue, @david-martin what is the value of having to define a health check for each listener vs all listeners in a gateway?
Where I could potentially see it being useful is 
1) if the applications are in a multi-tenanted environment and don't all have to conform to a given "production standard" with the current design the check at the gateway level could still be useful to those that want it. Adding 404 as an allowed response code 
2) an extension to the above, if the apps all have different auth they want to apply

I think the design here is focused around an single org where a "ops/sre" team are managing the environment and so want to specify a health check for all listeners in that environment. Thinking about allowing this type of policy at the data plane level, is worth exploration though. Just as is allowing RLP and AuthPolicy at the MC CP level.

If you're in a agreement in the value of this health check design lets follow this up with additional issues ?

maleck13: @roivaz interested in your thoughts here from an operational point of view

david-martin: > If you're in a agreement

I'd love to have this decision be more data driven.
Based on my own previous experiences, individual applications tend to be developed by different teams/people at different stages of the overall group of applications evolution. Different patterns may be necessary or evolve over time as well with regards endpoints (e.g. REST vs graphql vs custom). I do like the story of conforming to the same set of endpoints across applications with different hostnames for dns health checks.
In reality, how often is that possible?

Perhaps this could be similar to other policies like Auth where it can be defined at a coarse grain in the control plane, and be more fine grained in the data plane when it isn't specific enough. For example, I'd enforce that all applications/listeners/hosts should give a 200 OK on `/` for them to be healthy. If that needs to be changed, that's something configurable per application/listener/host... somewhere.

maleck13: Yeah I think what you are describing is a potential evolution that we should consider. We know the use case exits for DNS level health checks as it is offered by all the main providers. So it is that data point that drives the current design.

roivaz: I agree with @david-martin  that very rarely a wide group of applications conform to the same pattern regarding health checks, even though in kubernetes the /healthz and /readyz have become very popular, many applications/orgs still use custom patterns and those can vary between teams.

Also, things like timeouts or retries should be configurable per listener, because load might be different across different services, and different services might respond very differently in terms of latency, even orders of magnitude different. 

I think an organization would find it very hard to have a set of default values that can apply to every listener in a gateway unless all services exposed through it are very similar. This would end up forcing users to group similar services in the same gateway and deploying other gateways for those that don't follow the pattern or behave very differently on terms of latency.

maleck13: ok so I think here we are also expanding the feature set? So there are some potential follow ons:

1)  allow per listener health check to be defined (this is not supported by policy attachment currently) we would need to do some form of custom targeting 
2) potentially allow overriding the health check at the HTTPRoute level
3) consider additional policies that cover things like timeout and retry

What I would say is while limited in its current design nothing we are doing prevents us adding this type of functionality afterwards. 
IE (new targeting mechanism is back compatible as if not specified default is all listeners)
introducing support to override the health check lower down the hierarchy 
etc.
For now having some form of simple health check support seems reasonable. If you don't want to enforce it you can set 404 as an acceptable response? WDYT

david-martin: I think so.

> If you don't want to enforce it you can set 404 as an acceptable response?

I see what you're getting at, and that gets around the problem.
However, it should also possible to completely disable healthchecks for a gateway (maybe that's already possible in the spec by not setting it?)
If it can be disabled, it shouldn't get in the way of users that want to wait for the more flexible functionality per listener.

maleck13: Yes if not set there should be no health checks

maleck13 (commented):
I think once we resolve how we will handle the DNS records that are marked as unhealthy we should update this to cover that and we can then probably go ahead and merge it 

@philbrookes

docs/proposals/provider-agnostic-dns-health-checks.md
```diff
@@ -0,0 +1,151 @@
...
+  healthCheck:
+    endpoint: /health
+    failureThreshold: 5
+    port: 443
+    protocol: https
+    additionalHeaders:
+      - name: "..."
+        value: "..."
+      - name: "..."
+        value: "..."
```
roivaz, line 64: For the use case of setting an auth header in the request we should allow the value of the header to be retrieved from a secret, otherwise we make it harder for a user to keep their custom resources under version control.

philbrookes: This is a great suggestion, thanks Roi. I've updated the proposal.

maleck13 (approved):

PR #320, opened 2023-07-10
dependency pre-req and some associated makefile updates

laurafitzgerald:
There's no issue for this. Should I create one?

Spotted a small error `ls: /Users/lfitzger/MCG/multicluster-gateway-controller/bin/operator-sdk: No such file or directory` when running any of the make commands before the operator-sdk one which prompted to me to make one change which resulted in a few more inside the depencies.make file.

# Verification
Running `make dependencies` should result in the following binaries installed into bin directory

 clusteradm     controller-gen ginkgo         helm           istioctl       kind           kustomize      operator-sdk   setup-envtest  subctl         yq

and a successful message whether you previously had them installed or not.

### DIFF
--- a/hack/make/dependencies.make
+++ b/hack/make/dependencies.make
@@ -1,4 +1,4 @@
-## Build Dependencies
+##@ Build Dependencies
 
 ## system information
 ARCH ?= amd64
@@ -38,6 +38,7 @@ GINKGO_VERSION ?= v2.6.1
 .PHONY: dependencies
 dependencies: kustomize operator-sdk controller-gen envtest kind helm yq istioctl clusteradm subctl ginkgo
 	@echo "dependencies installed successfully"
+	@echo "consider running `export PATH=$PATH:$(pwd)/bin` if you haven't already done"
 
 KUSTOMIZE_INSTALL_SCRIPT ?= "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"
 .PHONY: kustomize
--- a/hack/make/dependencies.make
+++ b/hack/make/dependencies.make
@@ -1,4 +1,4 @@
-##@ Build Dependencies
+## Build Dependencies
 
 ## system information
 ARCH ?= amd64
@@ -20,6 +20,7 @@ OPERATOR_SDK ?= $(LOCALBIN)/operator-sdk
 CLUSTERADM ?= $(LOCALBIN)/clusteradm
 SUBCTL ?= $(LOCALBIN)/subctl
 GINKGO ?= $(LOCALBIN)/ginkgo
+YQ ?= $(LOCALBIN)/yq
 
 
 ## Tool Versions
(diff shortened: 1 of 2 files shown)


hack/make/dependencies.make
```diff
@@ -34,39 +35,42 @@ CLUSTERADM_VERSION ?= 0.6.0
 SUBCTL_VERSION ?= release-0.15
 GINKGO_VERSION ?= v2.6.1
 
+.PHONY: dependencies
+dependencies: kustomize operator-sdk controller-gen envtest kind helm yq istioctl clusteradm subctl ginkgo
+	@echo "dependencies installed successfully"
```
laurafitzgerald, line 40: Would it make sense to include the update to PATH here?

laurafitzgerald: e.g. `export PATH=$PATH:$(pwd)/bin`

laurafitzgerald: or even an echo to recommend it if it's not already done.

hack/make/dependencies.make
```diff
@@ -1,4 +1,4 @@
-##@ Build Dependencies
+## Build Dependencies
```
mikenairn, line 1: This is used to add a group in the help when running `make help` see https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/Makefile#L40

laurafitzgerald: ack. i'll add it back in. 👍

hack/make/dependencies.make
```diff
@@ -81,25 +85,26 @@ $(HELM):
 
 .PHONY: yq
 yq: $(YQ)
-	test -s $(LOCALBIN)/yq || GOBIN=$(LOCALBIN) go install github.com/mikefarah/yq/v4@$(YQ_VERSION)
+$(YQ):
+	test -s $(YQ) || GOBIN=$(LOCALBIN) go install github.com/mikefarah/yq/v4@$(YQ_VERSION)
```
mikenairn, line 89: Why remove the `yq` phony target? I assume local-setup is not working now?

hack/make/dependencies.make
mikenairn: I don't think we want to do that. Do we tell people somewhere to do that?

hack/make/dependencies.make
mikenairn: My mistake, you didn't remove it, but you added this for some reason https://github.com/Kuadrant/multicluster-gateway-controller/pull/320/files#diff-edf16a6750c4ce8ab32c9359e0f357d15f8c79fb01baba58ae1dba663e556b50R89. What's that doing?

mikenairn: ignore me, its fine.

hack/make/dependencies.make
jasonmadigan: I suppose some extra complexity if we export ourselves (other than potentially clobbering existing paths if they've got some of these installed already) is that we'd need to cater for a bunch of shells (`zsh`, `bash`).

We could output a path to include in their shell of choice at the end of local-setup?

hack/make/dependencies.make
laurafitzgerald: hmm. i thought I was getting errors without this but running it again it seems it's not necessary. A similar syntax is used in other make commands in this file. Although I'm not sure also why it would be required.

hack/make/dependencies.make
laurafitzgerald: ` Do we tell people somewhere to do that?` we don't but some of the individual commands did have an output to say consider doing it. I guess that's standard to make sure the binary is available.

hack/make/dependencies.make
mikenairn: What you did is correct, it's using make to check that the file already exists and do nothing if it does (Actually using make the way it was intended to be used). The `test -s $(YQ) ` part is actually redundant now, you could remove it, but i see it also being used on other targets unnecessarily .

jasonmadigan (approved):

PR #321, opened 2023-07-10
Fix flaky test

mikenairn:
Update the dnspolicy geo test to create a geo dns policy rather than updating an existing one. The update should work, but for whatever reason testenv doesn't alway appear to do it or the events are missed.


jasonmadigan (approved):
Saw this test failure with https://github.com/Kuadrant/multicluster-gateway-controller/actions/runs/5509711179/jobs/10044861769#step:4:124 - thanks for fixing.

PR #322, opened 2023-07-11
Add prometheus and thanos for metrics federation.

### LINKED ISSUE #304: Update the local setup to (optionally) install a federated metrics solution that exposes metrics in the hub in prometheus format (using thanos & prometheus)

david-martin:
Closes #304 

Adds an optional env var, METRICS_FEDERATION, defaults to empty.
If set during local-setup, thanos is installed in the hub cluster and prometheus in the spoke clusters.
The prometheus instance in the spoke clusters is configured to remote write metrics to the thanos receive instance in the hub.

A thanos query instance is available via an Ingress in the hub at https://thanos-query.172.32.0.2.nip.io

To verify, follow the included walkthrough.


config/thanos/manifests/thanos-query-serviceMonitor.yaml
```diff
@@ -0,0 +1 @@
+null
```
maleck13, line 1: needed?

david-martin: No, but these files in the manifests directory are generated.
I can remove them, but if we regen they'll have to be removed again.
They don't seem to cause issue with kustomize

I'm happy to remove them now if you want

david-martin: I updated the jsonnet file to generate these 2 files since I don't see any reason not to.
Changes pushed.

jasonmadigan:
Will stick a second set of eyes on

jasonmadigan (approved):
/lgtm
/approve

Issue #323, opened 2023-07-11
CRD validation for domain name, possibly other fields as well


jasonmadigan:
Imagine a world where some lunatic blindly copy-pastes from one of our walkthroughs, and ends up with a `controller-config.env` that looks something like:

```bash
# this sets up your default managed zone
AWS_DNS_PUBLIC_ZONE_ID=welpwelpwelpwelp
# this is the domain at the root of your zone (foo.example.com)
ZONE_ROOT_DOMAIN=hoppityboppity.hcpapps.net # replace this with your root domain
LOG_LEVEL=1
```

Said lunatic decides he knows better than @philbrookes, and runs the controller locally, rather than in a container:

```bash
(export $(cat ./controller-config.env | xargs) && export $(cat ./aws-credentials.env | xargs) && make build-controller install run-controller)
```

Following the OCM walkthrough verbatim, they'll eventually end up with the following managedzones:

```bash
$ kubectl get managedzones -A
NAMESPACE                NAME         DOMAIN NAME                                           ID                                  RECORD COUNT   NAMESERVERS                                                                                         READY
multi-cluster-gateways   mgc-dev-mz   hoppityboppity.hcpapps.net # replace this with your root domain   /hostedzone/Z02205993US0IQE3YLO37   2              ["ns-1583.awsdns-05.co.uk","ns-1100.awsdns-09.org","ns-796.awsdns-35.net","ns-242.awsdns-30.com"]   True
```

Obviously, domainName here (and possibly other fields) should be validated.

PR #325, opened 2023-07-11
Fixing some walkthrough docs

jasonmadigan:
Stepping through the OCM walkthrough today to get things back up and running, I ran into some issues:

* The `./controller-config.env` you end up with as part of the walkthrough is malforme
* Created a related issue here https://github.com/Kuadrant/multicluster-gateway-controller/issues/323 to not allow malformed domain names for `managezone`'s
* Added a `kubectl patch` for the numberOfClusters command later in the walkthrough

### DIFF
--- a/docs/how-to/ocm-control-plane-walkthrough.md
+++ b/docs/how-to/ocm-control-plane-walkthrough.md
@@ -16,28 +16,43 @@ We will start with a single cluster and move to multiple clusters to illustrate 
 
 ## Installation and Setup
 * Clone this repo locally 
-* Setup a `./controller-config.env` file in the root of the repo with the following key values
+* Setup a `./controller-config.env` env-var file in the root of the repo with the following keys. Fill in your own values as appropriate. You will need access to a domain or subdomain in Route 53 in AWS:
 
-    ```bash
-    AWS_DNS_PUBLIC_ZONE_ID=<AWS ZONE ID> # this sets up your default managed zone
-    ZONE_ROOT_DOMAIN=test.hcpapps.net # this is the domain at the root of your zone (foo.example.com)
-    LOG_LEVEL=1
-    ```   
+  ```bash
+  AWS_DNS_PUBLIC_ZONE_ID=Z01234567US0IQE3YLO00
+  ZONE_ROOT_DOMAIN=jbloggs.hcpapps.net
+  LOG_LEVEL=1
+  ```
+
+  | Env Var                  | Example Value           | Description                                           |
+  |--------------------------|-------------------------|-------------------------------------------------------|
+  | `ZONE_ROOT_DOMAIN`       | `jbloggs.hcpapps.net`   | Hostname for the root Domain                          |
+  | `AWS_DNS_PUBLIC_ZONE_ID` | `Z01234567US0IQE3YLO00` | AWS Route 53 Zone ID for specified `ZONE_ROOT_DOMAIN` |
+  | `LOG_LEVEL`              | `1`                     | Log level for the Controller                          |
 
 * setup a `./aws-credentials.env` with credentials to access route 53
 
-    For example:
+  For example:
+
     ```bash
     AWS_ACCESS_KEY_ID=<access_key_id>
     AWS_SECRET_ACCESS_KEY=<secret_access_key>
     AWS_REGION=eu-west-1
     ```
 
+  | Env Var                 | Example Value          | Description                                                    |
+  |-------------------------|------------------------|----------------------------------------------------------------|
+  | `AWS_ACCESS_KEY_ID`     | `AKIA1234567890000000` | Access Key ID, with access to resources in Route 53            |
+  | `AWS_SECRET_ACCESS_KEY` | `Z01234567US0000000`   | Access Secret Access Key, with access to resources in Route 53 |
+  | `AWS_REGION`            | `eu-west-1`            | AWS Region                                                     |
+
+
 * We're going to use an environment variable, `MGC_SUB_DOMAIN`, throughout this walkthrough. Simply run the below in each window you create:
 
-```bash
-export MGC_SUB_DOMAIN=myapp.test.hcpapps.net # replace this
-```
+  For example:
+  ```bash
+  export MGC_SUB_DOMAIN=myapp.jbloggs.hcpapps.net
+  ```
 
 * Alternatively, to set a default, add the above environment variable to your `.zshrc` or `.bash_profile`. To override this as a once-off, simply `export MGC_SUB_DOMAIN`.
 
@@ -140,7 +155,7 @@ Open two windows, which we'll refer to throughout this walkthrough as:
     make build-controller kind-load-controller deploy-controller
     ```
 
-    *Alternatively, in `T2`, you can run the controller locally instead by running:*
+    *Alternatively, in a new window we'll call `T3`, you can run the controller locally instead by running:*
 
     ```bash
     (export $(cat ./controller-config.env | xargs) && export $(cat ./aws-credentials.env | xargs) && make build-controller install run-controller)
@@ -345,13 +360,11 @@ So now we have a working gateway with DNS and TLS configured. Let place this gat
     kubectl label managedcluster kind-mgc-workload-1 ingress-cluster=true
     ```
 
-1. This has added our workload-1 cluster to the ingress clusterset. Next we need to modify our placement. In `T1`, run:
+1. This has added our workload-1 cluster to the ingress clusterset. Next we need to modify our placement to update our `numberOfClusters` to 2. To patch, in `T1`, run:
 
     ```bash
-    kubectl edit placement http-gateway -n multi-cluster-gateways
+    kubectl patch placement http-gateway -n multi-cluster-gateways --type='json' -p='[{"op": "replace", "path": "/spec/numberOfClusters", "value": 2}]'
     ```
-
-1. This will open your `$EDITOR` - in here, edit the spec change the `numberOfClusters` to be 2, and save.
 
 1. In `T2` window execute the following to see the gateway on the workload-1 cluster:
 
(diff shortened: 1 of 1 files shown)


docs/how-to/ocm-control-plane-walkthrough.md
```diff
@@ -19,10 +19,8 @@ We will start with a single cluster and move to multiple clusters to illustrate
 * Setup a `./controller-config.env` file in the root of the repo with the following key values
 
     ```bash
-    # this sets up your default managed zone
-    AWS_DNS_PUBLIC_ZONE_ID=<AWS ZONE ID>
-    # this is the domain at the root of your zone (foo.example.com)
-    ZONE_ROOT_DOMAIN=test.hcpapps.net # replace this with your root domain
+    AWS_DNS_PUBLIC_ZONE_ID=<AWS ZONE ID> # this sets up your default managed zone
+    ZONE_ROOT_DOMAIN=test.hcpapps.net # this is the domain at the root of your zone (foo.example.com)
```
laurafitzgerald, line 23: @jasonmadigan i applied the blindly copying logic to this and it still results in a malformed domain in the managedzone cr after running `make local-setup` before doing any further steps. is the idea to address it by adding validation as suggested in https://github.com/Kuadrant/multicluster-gateway-controller/issues/323 or would it be better to copy the text that we've got in the README here https://github.com/Kuadrant/multicluster-gateway-controller#1-running-the-controller-in-the-cluster that might be a bit more bullet proof for noobs.

jasonmadigan: I had to wipe a bunch of resources and restart the controller for a non-malformed config to be picked up

laurafitzgerald: I didn't get to the point of starting the controller. Steps I took were
- copy and paste into the file 
- run make local-setup
- kubectl get managedzone -n multi-cluster-gateways -o yaml
- and the spec.domainname included the commented out section.

laurafitzgerald: cool. i'll continue with the steps. 👍

laurafitzgerald: when I get to `(export $(cat ./controller-config.env | xargs) && export $(cat ./aws-credentials.env | xargs) && make build-controller install run-controller)` I get the following error `export: not valid in this context: (foo.example.com)`

jasonmadigan: It sort of sounds like your `controller-config.env` is still broken - can you try it without any included comments?

laurafitzgerald: after fixing the file to have the correct values i'm facing this error `{"error": "InvalidInput: 1 validation error detected: Value 'Z0860675CQXHTYO3UOX3 # this sets up your default managed zone' at 'resourceId' failed to satisfy constraint: Member must have length less than or equal to 32\n\tstatus code: 400, request id: 708873d1-7ee5-424e-b1df-4906b6099964"}`

docs/how-to/ocm-control-plane-walkthrough.md
```diff
@@ -136,18 +134,29 @@ Open three windows, which we'll refer to throughout this walkthrough as:
...
+    ```bash
+    (export $(cat ./controller-config.env | xargs) && export $(cat ./aws-credentials.env | xargs) && make build-controller install run-controller)
+    ```
+
 ### Check the managed zone
 
 1. Lets ensure our `managedzone` is present. In `T1`, run the following:
 
     ```bash
     export KUBECONFIG=$(pwd)/local/kube/control-plane.yaml
```
laurafitzgerald, line 154: When I ran this command it actually knocked of my KUBECONFIG. I hadn't run the command on line 67 because it said optional. i was able to run other `kubectl` commands in the meantime.

docs/how-to/ocm-control-plane-walkthrough.md
jasonmadigan: re: `"InvalidInput: 1 validation error detected: Value 'BLAHBLAH # this sets up your default managed zone'` where is BLAHBLAH coming from? sounds like it could be from either aws-credentials.env or controller-config.env

laurafitzgerald: yeah. BLAHBLAH is just the key value but I changed it here. in the error message it's a proper ID. I think changing it make sense. i really applied the blindly copying mindset so that a newbie wouldn't get caught out. 👍

docs/how-to/ocm-control-plane-walkthrough.md
```diff
@@ -16,38 +16,51 @@ We will start with a single cluster and move to multiple clusters to illustrate
...
+
+
 * We're going to use an environment variable, `MGC_SUB_DOMAIN`, throughout this walkthrough. Simply run the below in each window you create:
 
-```bash
-export MGC_SUB_DOMAIN=myapp.test.hcpapps.net # replace this
-```
+  For example:
+  ```bash
+  export MGC_SUB_DOMAIN=myapp.jbloggs.hcpapps.net
```
laurafitzgerald, line 54:
```suggestion
  export MGC_SUB_DOMAIN=jbloggs.hcpapps.net
```

docs/how-to/ocm-control-plane-walkthrough.md
```diff
@@ -16,38 +16,51 @@ We will start with a single cluster and move to multiple clusters to illustrate
...
+* Setup a `./controller-config.env` env-var file in the root of the repo with the following keys. Fill in your own values as appropriate. You will need access to a domain or subdomain in Route 53 in AWS:
 
-    ```bash
-    # this sets up your default managed zone
-    AWS_DNS_PUBLIC_ZONE_ID=<AWS ZONE ID>
-    # this is the domain at the root of your zone (foo.example.com)
-    ZONE_ROOT_DOMAIN=test.hcpapps.net # replace this with your root domain
-    LOG_LEVEL=1
-    ```   
+  ```bash
```
laurafitzgerald, line 21: This is a sight to behold. 😍

docs/how-to/ocm-control-plane-walkthrough.md
jasonmadigan: actually, this one should definitely be <something>.`ZONE_ROOT_DOMAIN` from above - basically, this is used a little later on as the "app" host for the example apps that are deployed towards the end of the walkthrough

laurafitzgerald: This one is used in some of the crs. which is why i think the `myapp.` should be removed.

docs/how-to/ocm-control-plane-walkthrough.md
```diff
@@ -336,14 +360,12 @@ So now we have a working gateway with DNS and TLS configured. Let place this gat
     kubectl label managedcluster kind-mgc-workload-1 ingress-cluster=true
     ```
 
-1. This has added our workload-1 cluster to the ingress clusterset. Next we need to modify our placement. In `T1`, run:
+1. This has added our workload-1 cluster to the ingress clusterset. Next we need to modify our placement to update our `numberOfClusters` to 2. To patch, in `T1`, run:
 
     ```bash
-    kubectl edit placement http-gateway -n multi-cluster-gateways
+    kubectl patch placement http-gateway -n multi-cluster-gateways --type='json' -p='[{"op": "replace", "path": "/spec/numberOfClusters", "value": 2}]'
```
laurafitzgerald, line 366: Verifying this patch worked for me.

docs/how-to/ocm-control-plane-walkthrough.md
laurafitzgerald: you can ignore this suggestion.

docs/how-to/ocm-control-plane-walkthrough.md
```diff
@@ -136,18 +149,29 @@ Open three windows, which we'll refer to throughout this walkthrough as:
...
     ```bash
     make build-controller kind-load-controller deploy-controller
     ```
 
+    *Alternatively, in a new window we'll call `T3`, you can run the controller locally instead by running:*
+
+    ```bash
+    (export $(cat ./controller-config.env | xargs) && export $(cat ./aws-credentials.env | xargs) && make build-controller install run-controller)
+    ```
+
```
mikenairn, line 163: I'm not sure if you should say this here, depends who this walkthough is targeted at i guess, but running the controller like that is really for devs i would say and if you are just trying it out you should deploy it.

It also turns out people just blindly copy all these things and paste away into terminals so you are inviting people into a situation where they have the controller running in two places which wouldn't be ideal.

docs/how-to/ocm-control-plane-walkthrough.md
```diff
@@ -16,38 +16,51 @@ We will start with a single cluster and move to multiple clusters to illustrate
...
     ```
 
+  | Env Var                 | Example Value          | Description                                                    |
+  |-------------------------|------------------------|----------------------------------------------------------------|
+  | `AWS_ACCESS_KEY_ID`     | `AKIA1234567890000000` | Access Key ID, with access to resources in Route 53            |
+  | `AWS_SECRET_ACCESS_KEY` | `Z01234567US0000000`   | Access Secret Access Key, with access to resources in Route 53 |
+  | `AWS_REGION`            | `eu-west-1`            | AWS Region                                                     |
+
+
 * We're going to use an environment variable, `MGC_SUB_DOMAIN`, throughout this walkthrough. Simply run the below in each window you create:
```
david-martin, line 50: No harm in reminding of this when T2 is opened later in the guide.
Can be in a follow up PR since you're planning on making further changes around the getting started experience.

docs/how-to/ocm-control-plane-walkthrough.md
jasonmadigan: that's a good point actually. We could do with a decent dev setup guide which is quite different to a "here's how to use this thing" guide.

PR #326, opened 2023-07-12
Domain name crd validation

jasonmadigan:
Fixes https://github.com/Kuadrant/multicluster-gateway-controller/issues/323

Adds hostname validation to `domainName` in the `managedzone` CR

Regex:

https://regex101.com/r/ZIYOaE/1

### DIFF
--- a/config/crd/bases/kuadrant.io_managedzones.yaml
+++ b/config/crd/bases/kuadrant.io_managedzones.yaml
@@ -61,6 +61,7 @@ spec:
                 type: string
               domainName:
                 description: Domain name of this ManagedZone
+                pattern: ^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$
                 type: string
               id:
                 description: ID is the provider assigned id of this  zone (i.e. route53.HostedZone.ID).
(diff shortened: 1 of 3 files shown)


philbrookes:
it looks like this will allow the likes of `.co.uk`?

jasonmadigan:
> co.uk

Thought about this a bit alright - lonesome second-level domains are going to be really hard to validate without having to refer to a list of them. I think I was mostly looking to validate against the most egregious of bad hostnames (e.g. the famous Google-killer: `hoppityboppity.hcpapps.net # replace this with your root domain`)

philbrookes:
> 

yes ok, this is still a drastic improvement on the existing validation (nothing). We do have some code internally to not make use of managedzones that are just TLDs like `co.uk` but our protection against misuse is thin at best.

Probably makes more sense for something like that to exist in a webhook.

philbrookes (approved):
/lgtm

config/crd/bases/kuadrant.io_managedzones.yaml
```diff
@@ -61,6 +61,7 @@ spec:
                 type: string
               domainName:
                 description: Domain name of this ManagedZone
+                pattern: ^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])$
```
laurafitzgerald, line 64: Just noting that I've validated this. Alongside the ocm walkthrough I faced the following error after copying the file directly 
`{"error": "InvalidInput: 1 validation error detected: Value 'IDVALUE # this sets up your default managed zone' at 'resourceId' failed to satisfy constraint: Member must have length less than or equal to 32\n\tstatus code: 400, request id: a52d2233-67a4-4c6a-a3ba-610312cef431"}

PR #327, opened 2023-07-12
Fix dnsrecord updates

mikenairn:
Fixes an issue where the dnsrecord can be updated unnecessary.

* Updates set endpoints to ensure the list is always the same (ordered by set id).
* Remove unnecessary dnsrecord update, check for status update and ignore conflict errors.


philbrookes (approved):
/lgtm

Issue #260, opened 2023-06-15
Move all controller integration tests into their own test package


mikenairn:
**What**

Move all controller testenv tests into a single "integration" test package.

**Why**

We have a mix of tests written for the ginkgo framework using testenv (integration test) and others that are using the standard go testing libs (unit tests), all in the same package. The problems with that are better described here https://github.com/kubernetes-sigs/kubebuilder/issues/2171, and part of the reason build tags were added https://github.com/Kuadrant/multicluster-gateway-controller/pull/244 so we could use ginkgo flags for the suite tests (verbose logging etc..).

All the ginkgo tests just add the specs into the suite_test file, where it would be better if we could separate those out into individual spec files i.e.  gateway_controller_test.go etc.. but that would be an issue if we were also adding unit tests for that file. We are also adding a suite_test for each individual controller, which is probably because of the way kubebuider generates a controller expecting them all to be in the same directory, where we have them in individual packages.

**How**

* Create a single test_suite file that initialises all controllers in a dedicated test package `test/integration`?
* Add individual test files for each controller in the same package (Move specs from individual test_suite files)
* Add `integration` and `unit` build tags to all tests
* Add `test-integration` and `test-unit` make targets
* Update CI to run unit and integration tests on all PRs

Issue #282, opened 2023-06-28
Uniform unit and integration tests 


makslion:
## What
Our unit and integration tests are currently not uniform across the repository. Meaning that they use different approaches for testing controllers and depending on the person doing tests preference was given to either of the two. This creates inconsistency and could make future development and testing tricky.

## How
Following the [discussion](https://redhat-internal.slack.com/archives/C047YHJJG2D/p1687895228432679) from Slack, it would be a good idea to bring those tests to the same idea/concept. 
This would involve: 
1. Transfering test cases from unit to integration and back so we have a consistent scope for each test type. Covered in #352 
2. Refactoring existing unit tests to follow the same "pattern" 
3. Unifying `fake` elements for unit tests and separating them from a "use case" (making them more reusable) 
4. Joining and unifying helper functions that could be reused across different packages into the `test` package 
5. Ensuring integration tests follow the same idea of having cases in separate from the setup files. Covered in #260 

## Done
- [x] Unit tests and integration tests have the same scope across the repository delegated to the #352 
- [ ] Booth test types following the same pattern 
- [ ] Extra logic that could be reused is stored in the `test` package
- [ ] the `fake` implementations of interfaces reside next to the real-life ones. (one exception to be handled by #330 )

PR #329, opened 2023-07-12
Moved integration test to its package

Ygnas:
#260


test/integration/dnspolicy_controller_test.go
```diff
@@ -327,8 +333,7 @@ var _ = Describe("DNSPolicy", Ordered, func() {
 					// must exist
 					Expect(err).ToNot(HaveOccurred())
 					return existingGateway.GetAnnotations()
-				}, time.Second*5, time.Second).ShouldNot(HaveKey(DNSPolicyBackRefAnnotation))
-
+				}, time.Second*5, time.Second).ShouldNot(HaveKeyWithValue(DNSPoliciesBackRefAnnotation, policiesBackRefValue))
```
david-martin, line 336: Is this because the annotation still exists, but with a different value?

test/integration/managedzone_controller_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/Kuadrant/multicluster-gateway-controller/pkg/apis/v1alpha1"
+	//+kubebuilder:scaffold:imports
+)
+
+var _ = Describe("ManagedZoneReconciler", func() {
+	Context("testing ManagedZone controller", func() {
```
david-martin, line 30: I'm guessing this is a copy/paste of the original file with the Before/AfterSuite removed?
No other changes?

test/integration/suite_test.go
```diff
@@ -102,8 +122,10 @@ var _ = BeforeSuite(func() {
 	Expect(err).ToNot(HaveOccurred())
 
 	dnsProvider := &dns.FakeProvider{}
+	certificates := tls.NewService(k8sManager.GetClient(), "glbc-ca")
 	dns := dns.NewService(k8sManager.GetClient())
-	plc := NewTestOCMPlacer()
+	plc := placement.NewOCMPlacer(k8sManager.GetClient())
+	testPlc := NewTestOCMPlacer()
```
david-martin, line 128: Let's capture a follow up issue to avoid 2 different placer impls if possible.
There's some important logic in the real placer that the gateway controller relies on before we even expect ocm to be running e.g. the ManifestWork parsing of statusfeedback rules.

david-martin (approved):
A few comments, nothing major.
I'll approve, but would be good to get a 2nd set of eyes on it from @mikenairn if possible.

makslion:
what about the `dnspolicy` and `gateway` controllers? :thinking: they should also hold a few integration tests that should be evicted, right?

makslion:
And I believe we should also (as mentioned in #260 ) split our Test Suite check into integration and unit - since we have two separate make targets for each of them.

test/integration/suite_test.go
makslion: two fake placers will be addressed here https://github.com/Kuadrant/multicluster-gateway-controller/issues/282 - they will be joined into one big happy fake placer - no need for the follow-up task if I understand things correctly.

test/integration/helper_test.go
```diff
@@ -22,12 +22,17 @@ import (
...
+	ConsistentlyTimeoutMedium = time.Second * 60
+	TestRetryIntervalMedium   = time.Millisecond * 250
+	TestPlacedGatewayName     = "test-placed-gateway"
+	TestPlacedClusterName     = "test-placed-cluster"
+	TestAttachedRouteName     = "test.example.com"
+	TestAttachedRouteAddress  = "172.0.0.3"
+	nsSpoke1Name              = "test-spoke-cluster-1"
+	nsSpoke2Name              = "test-spoke-cluster-2"
+	defaultNS                 = "default"
+	gatewayFinalizer          = "kuadrant.io/gateway"
```
makslion, line 35: the gateway finalizer defined in the `gateway` package. Is it a good idea to have the same value hardcoded in two different places? And would id make sense to keep all the test constants in one [place](https://github.com/Ygnas/multicluster-gateway-controller/blob/integration-test_package/test/util/helper.go#L21-L34)?

test/integration/suite_test.go
Ygnas: I have created issue #330

test/integration/dnspolicy_controller_test.go
mikenairn: There are 2 annotations we need to check and it was incorrectly checking the same one twice. The `DNSPoliciesBackRefAnnotation` won't be removed completely on deletion, but should be empty.

test/integration/suite_test.go
mikenairn: I'd have expected this to be done as part of this move, but a follow up is fine

Makefile
```diff
@@ -72,7 +72,7 @@ test-unit: manifests generate fmt vet envtest ## Run unit tests.
 
 .PHONY: test-integration
 test-integration: manifests generate fmt vet envtest ## Run integration tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test ./... -tags=integration -coverprofile cover-integration.out -ginkgo.v -v -timeout 0
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test ./test/... -tags=integration -coverprofile cover-integration.out -ginkgo.v -v -timeout 0
```
mikenairn, line 75: Might be better to just use ginkgo command here, see [e2e:](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/Makefile#L83)

```
KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" $(GINKGO) -tags=integration -v ./test/integration
```

You'll also need to change the e2e target to run the specs in the e2e dir only:

```
$(GINKGO) -tags=e2e -v ./test/e2e
```

mikenairn:
Looks good, small issue with the e2e tests, but if you update the make target it should work.

mikenairn:
/lgtm
/approve

The integration tests are still being run as part of `make test`, so are still being tested via CI, so `Update CI to run unit and integration tests on all PRs` is covered. A good follow on to this though would be to create separate CI jobs for both using the specific targets `make test-unit` and `make test-integration`

PR #331, opened 2023-07-13
Fix dnsrecord when no addresses in any targets

### LINKED ISSUE #328: DNSRecord can be updated with invalid endpoints
Since the record structure was changed as part of https://github.com/Kuadrant/multicluster-gateway-controller/pull/293, it's now possible that a DNSRecord can be updated with only part of the desired records in  place.

If for whatever reason no ClusterTarget returns a valid ip or host address, the top level CNAMES will still be created, but none of the others.

`SetEndpoints` should be updated to create these in reverse order and make sure that a valid target record is created before creating any of the other endpoints.

mikenairn:
closes #328  

Fixes an issue where a DNSRecord could get created with some endpoints, but have no actual targets for them to point to.

If we have cluster targets that have no addresses when building up our list of Endpoints for the DNSRecord they are now completely ignored and no CNAME or A records related to that cluster are created.  If no targets in a geo have valid addresses, the geo is ignored and no records created for it.  If no records can be created for any targets (no targets have addresses), nothing is created and an empty endpoints list returned.

### DIFF
--- a/pkg/dns/service.go
+++ b/pkg/dns/service.go
@@ -182,37 +182,15 @@ func (s *Service) SetEndpoints(ctx context.Context, mcgTarget *MultiClusterGatew
 	}
 
 	var (
-		newEndpoints []*v1alpha1.Endpoint
-		endpoint     *v1alpha1.Endpoint
+		newEndpoints    []*v1alpha1.Endpoint
+		endpoint        *v1alpha1.Endpoint
+		defaultEndpoint *v1alpha1.Endpoint
 	)
-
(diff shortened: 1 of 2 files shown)


david-martin:
/lgtm
/approve

Changes make sense.

Tested & verified locally and in route53.
To verify, I deleted the metallb IPAddressPool in the hub and spoke clusters to force no IPAddress assignment on the Gateway. (The Gateway has a Service of type LoadBalancer that would normally get assigned an IP by metallb).
`kubectl -n metallb-system delete IPAddressPool example`

PR #332, opened 2023-07-13
Fix make fmt

mikenairn:
The make fmt command was using the wrong path causing odd issues.


david-martin:
/lgtm
/approve

This explains some odd issues I've seen locally, but others I asked didn't.
I thought it was an issue with gofmt version.

PR #334, opened 2023-07-13
Separate CI jobs for both unit and itegration tests

### LINKED ISSUE #333: Separate CI jobs for unit and integration tests using specific targets
Currently, our integration tests are being run as part of the `make test` command, causing both unit and integration tests to run together in our CI. It makes it harder to determine the specific source of any test failures 

**Proposed Solution:**
Improving our testing, it would be beneficial to create separate CI jobs for our unit and integration tests.
We can leverage the specific targets `make test-unit` and `make test-integration` that are designed to handle these tests individually.

This separation would provide a clearer distinction between test types, making it easier to troubleshoot failures and monitor the results.

_@mikenairn in https://github.com/Kuadrant/multicluster-gateway-controller/issues/329#issuecomment-1634078773_

Ygnas:
closes #333

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -61,8 +61,8 @@ jobs:
       - name: Run make verify-manifests
         run: |
           make verify-manifests
-  test_suite:
-    name: Test Suite
+  integration_test_suite:
+    name: Integration Test Suite
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


mikenairn:
@Ygnas You need to add the `ginkgo` target here https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/Makefile#L74, see https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/Makefile#L82

PR #337, opened 2023-07-14
Gh 273 support for wildcard listeners and clean up dns and tls when listeners removed

maleck13:
**What**

adds support for wildcard hostnames in gateway listeners. Also adds support for cleaning up DNS records when a listener is removed. 

As part of this work the code is moving to focus around listeners rather than host names and logic for DNS is being focused in just the DNSPolicy controller

While TLS is still handled by the gateway controller, if we end up with a TLSPolicy I can imagine we would also want to follow the same pattern as DNSPolicy

### DIFF
--- a/config/cert-manager/kustomization.yaml
+++ b/config/cert-manager/kustomization.yaml
@@ -21,4 +21,4 @@ helmCharts:
       cainjector:
         enabled: false
       # Customize active controllers
-      extraArgs: ["--controllers=*,-ingress-shim"]
+      extraArgs: ["--controllers=*,-ingress-shim",--enable-certificate-owner-ref=true]
--- a/pkg/controllers/dnspolicy/dnspolicy_controller.go
+++ b/pkg/controllers/dnspolicy/dnspolicy_controller.go
@@ -56,7 +56,7 @@ func (c *DNSPolicyRefsConfig) PolicyRefsAnnotation() string {
 // DNSPolicyReconciler reconciles a DNSPolicy object
 type DNSPolicyReconciler struct {
 	reconcilers.TargetRefReconciler
-	DNSProvider dns.Provider
+	DNSProvider dns.DNSProviderFactory
 	Placement   gateway.GatewayPlacer
 }
 
--- a/pkg/controllers/dnspolicy/dnspolicy_controller.go
+++ b/pkg/controllers/dnspolicy/dnspolicy_controller.go
@@ -56,8 +56,7 @@ func (c *DNSPolicyRefsConfig) PolicyRefsAnnotation() string {
 // DNSPolicyReconciler reconciles a DNSPolicy object
 type DNSPolicyReconciler struct {
 	reconcilers.TargetRefReconciler
-	DNSProvider dns.DNSProviderFactory
-	HostService gateway.HostService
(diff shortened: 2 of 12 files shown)


config/cert-manager/kustomization.yaml
```diff
@@ -21,4 +21,4 @@ helmCharts:
       cainjector:
         enabled: false
       # Customize active controllers
-      extraArgs: ["--controllers=*,-ingress-shim"]
+      extraArgs: ["--controllers=*,-ingress-shim",--enable-certificate-owner-ref=true]
```
maleck13, line 24: this ensure secrets are cleaned up when certificates are removed so we no longer need to do that in our controller

pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -245,5 +246,9 @@ func (r *DNSPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			&source.Kind{Type: &clusterv1.ManagedCluster{}},
 			handler.EnqueueRequestsFromMapFunc(clusterEventMapper.MapToDNSPolicy),
 		).
+		// Watches(
```
maleck13, line 249: might not do this in this PR but we may want to queue DNSPolicy if changes happen to the DNSRecord (example it gets deleted by accident currently it wont get recreated without some for  of change to the gateway or the policy)

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -79,42 +124,269 @@ func (r *DNSPolicyReconciler) reconcileGatewayDNSRecords(ctx context.Context, ga
...
 
-		if err := r.HostService.SetEndpoints(ctx, mcgTarget, dnsRecord); err != nil {
+		if err := r.SetEndpoints(ctx, mcgTarget, dnsRecord, &listener); err != nil {
 			return fmt.Errorf("failed to add dns record dnsTargets %s %v", err, mcgTarget)
 		}
+	}
+	return nil
+}
+
+func (r *DNSPolicyReconciler) SetEndpoints(ctx context.Context, mcgTarget *dns.MultiClusterGatewayTarget, dnsRecord *v1alpha1.DNSRecord, listener *gatewayv1beta1.Listener) error {
```
mikenairn, line 161: This can probably be private. This is also out of date with what is on main. I wonder if you are moving all these things from the service if we shouldn't do that in it's own PR without functional changes. We'd be able to get it merged quicker.

pkg/controllers/dnspolicy/dnspolicy_controller.go
mikenairn: Probably want to think about this one. Can see the issue alright, but we ideally don't want to be triggering the policy reconcile unnecessarily i.e. it's just update the record

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
maleck13: ok will spit some of this out

maleck13:
depends on https://github.com/Kuadrant/multicluster-gateway-controller/pull/338

PR #343, opened 2023-07-17
Handle MangedCluster delete events

### LINKED ISSUE #303: Handle the deletion of a managedcluster resource
**What**

When a managedcluster is deleted, we need to ensure that the records for any gateway on that cluster are cleaned up. 
Currently the records are not cleaned up and the controller has several reconcile errors

If the managedcluster is deleted:
- grace period should be ignored
- dns records should be tidied up

It should also be possible to modify things like the DNSPolicy and GEO attributes of existing clusters.
Currently the DNSPolicy controller wont apply a new country code as it is failing to find a cluster

```
ERROR	Reconciler error	{"controller": "dnspolicy", "controllerGroup": "kuadrant.io", "controllerKind": "DNSPolicy", "DNSPolicy": {"name":"prod-web","namespace":"multi-cluster-gateways"}, "namespace": "multi-cluster-gateways", "name": "prod-web", "reconcileID": "97742d88-5b84-44c7-92b0-b96ba4e0985e", "error": "get cluster gateway failed: ManagedCluster.cluster.open-cluster-management.io \"kind-mgc-workload-1\" not found"}

```

david-martin:
Fixes #303

When a ManagedCluster is deleted, don't attempt to update the DNSPolicy straight away.
Instead, reconcile the Gateway first, which will trigger a DNSPolicy reconcile if there are any changes related to placement changes e.g.
IPAddresses changing.

When a ManagedCluster is deleted, don't wait for the grace period before deleting a Gateway (via ManifestWork).
Instead, delete the ManifestWork straight away.
Due to a known issue in OCM, https://github.com/open-cluster-management-io/ocm/issues/140 the ManifestWork is not removed after the deletionTimestamp is added in this sceanrio.
To allow for this, the gateway placement logic will filter out any ManifestWork resources that have a deletionTimestamp set.

To verify:

* Follow through the full ocm walkthrough
* Delete 1 of the ManagedClusters e.g. `kubectl delete managedcluster kind-mgc-workload-1`
* Verify the DNSPolicy status is `Ready` and doesn't have an error like [this](https://gist.github.com/david-martin/49421cde1fd12ac4d94fd3a9cfade228#file-dnspolicies-yaml-L24-L30)
* Verify the Gateway (and the kuadrant-multi-cluster-gateways namespacae) no longer exists in the `kind-mgc-workload-1` cluster
* Verify the [dnspolicy errors](https://gist.github.com/david-martin/49421cde1fd12ac4d94fd3a9cfade228#file-mgc_controller-log-L21-L27) are no longer happening in the logs
* Verify the DNSRecord only has 1 set of endpoints for 1 IPAddress [instead of 2](https://gist.github.com/david-martin/49421cde1fd12ac4d94fd3a9cfade228#file-dnsrecords-yaml-L29-L71).

### DIFF
--- a/pkg/controllers/dnspolicy/cluster_eventmapper.go
+++ b/pkg/controllers/dnspolicy/cluster_eventmapper.go
@@ -33,6 +33,14 @@ func (m *ClusterEventMapper) MapToDNSPolicy(obj client.Object) []reconcile.Reque
 func (m *ClusterEventMapper) mapToPolicyRequest(obj client.Object, policyKind string, policyRefsConfig common.PolicyRefsConfig) []reconcile.Request {
 	logger := m.Logger.V(1).WithValues("object", client.ObjectKeyFromObject(obj))
 
+	if obj.GetDeletionTimestamp() != nil {
+		// Ignore ManagedCluster delete events.
+		// Create/Update events are OK as ManagedCluster custom attributes may change, affecting DNSPolicies.
+		// However, deleting a ManagedCluster shouldn't affect the DNSPolicy directly until the related
+		// Gateway is deleted from that ManagedCluster (and reconciled then via watching Gateways, not ManagedClusters)
+		return []reconcile.Request{}
+	}
+
 	clusterName := obj.GetName()
 
 	allGwList := &gatewayapiv1beta1.GatewayList{}
--- a/pkg/placement/ocm.go
+++ b/pkg/placement/ocm.go
@@ -173,7 +173,22 @@ func (op *ocmPlacer) Place(ctx context.Context, upStreamGateway *gatewayv1beta1.
 			return existingClusters, err
 		}
 
-		if err := gracePeriod.GracefulDelete(ctx, op.c, w); err != nil {
+		// Check if the ManagedCluster still exists,
+		// otherwise delete without any grace period.
+		// This can happen if a ManagedCluster is deleted,
+		// as opposed to just a placement decision changing
+		ignoreGrace := false
+		managedCluster := &clusterv1.ManagedCluster{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: cluster,
+			},
+		}
+		err = op.c.Get(ctx, client.ObjectKeyFromObject(managedCluster), managedCluster)
+		if k8serrors.IsNotFound(err) {
+			log.V(3).Info(fmt.Sprintf("ManagedCluster not found '%s', ignoring grace period", cluster))
+			ignoreGrace = true
+		}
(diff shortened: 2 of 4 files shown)


maleck13:
@david-martin I agree on the responsibility split proposed. 

IE the clean up of the manifestwork is something that OCM should handle when a managed cluster is removed.  

what is the downside of the klusterlet not removing the finalizer? Do we keep re-queuing it? I am wondering if we take care of our resources and handle the MC being gone, do we need to set the delete option? Or can we just leave it at that?

pkg/controllers/dnspolicy/cluster_eventmapper.go
```diff
@@ -33,6 +33,14 @@ func (m *ClusterEventMapper) MapToDNSPolicy(obj client.Object) []reconcile.Reque
 func (m *ClusterEventMapper) mapToPolicyRequest(obj client.Object, policyKind string, policyRefsConfig common.PolicyRefsConfig) []reconcile.Request {
 	logger := m.Logger.V(1).WithValues("object", client.ObjectKeyFromObject(obj))
 
+	if obj.GetDeletionTimestamp() != nil {
+		// Ignore ManagedCluster delete events.
+		// Create/Update events are OK as ManagedCluster custom attributes may change, affecting DNSPolicies.
+		// However, deleting a ManagedCluster shouldn't affect the DNSPolicy directly until the related
+		// Gateway is deleted from that ManagedCluster (and reconciled then via watching Gateways, not ManagedClusters)
+		return []reconcile.Request{}
+	}
```
mikenairn, line 42: makes sense :+1:

pkg/placement/ocm.go
```diff
@@ -173,7 +173,22 @@ func (op *ocmPlacer) Place(ctx context.Context, upStreamGateway *gatewayv1beta1.
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name: cluster,
+			},
+		}
+		err = op.c.Get(ctx, client.ObjectKeyFromObject(managedCluster), managedCluster)
+		if k8serrors.IsNotFound(err) {
+			log.V(3).Info(fmt.Sprintf("ManagedCluster not found '%s', ignoring grace period", cluster))
+			ignoreGrace = true
+		}
+		if err := gracePeriod.GracefulDelete(ctx, op.c, w, ignoreGrace); err != nil {
```
mikenairn, line 191: In order for  this to get here on cluster deletion, it must be updating the PlacementDecision as expected and the cluster that was deleted removed?

mikenairn: nvm, this is answered here https://github.com/Kuadrant/multicluster-gateway-controller/issues/303#issuecomment-1635890565

PR #344, opened 2023-07-18
some code cleanup suggestions

laurafitzgerald:
These are just some suggestions and completed as part of some short upskilling. Happy for them to be rejected if they're not suitable at this time. 

- Rename of Reconcilers and associated objects so that the naming isn't duplicating the package name
- Split out gatewayclass to it's own controller
- Move params functions out to it's own package under local
- Consistent naming on package imports k8serrors metav1
- Small spelling fixes.


laurafitzgerald:
/hold
some tests are failing so taking a 👀

mikenairn:
/hold

Some of these changes look fine, but as discussed on slack, some of them are forcing us further away from kubebuilder where we might actually want to consider going bask the other way.  

I'd undo all the renaming of things in the controllers  until we have discussed this more.

PR #345, opened 2023-07-18
initial draft of a TLSPolicy

jasonmadigan:
Early draft for a `TLSPolicy`, which I'd like to square away the details on before writing an implementation. There are still some open considerations that need work.

One particular area of doubt is whether we should have alternative strategies for the types of certificates we issue, or instead let the listeners/hosts determine the type of certificate we issue. `cert-manager` doesn't seem to make distinctions here, rather it issues certs (be they wildcard or for known hosts), via `dnsNames`


docs/proposals/TLSPolicy.md
```diff
@@ -0,0 +1,81 @@
+# WIP
+# TLSPolicy API
+Currently, TLS certificates issued by the MGC are issued and managed using `cert-manager`. Configuration for these certificates are opague to the end-user - we'd like to gradually expose a TLSPolicy to cover a range of use-cases where administrators require more flexibile configuration.
+
+## Problem
+
+Gateway Administrators require a comprehensive way to manage Automated, ACME-based TLS for Gateways. As gateways multiply and scale, there's a need for a flexible, policy-based API for handling Automatic TLS. Currently (July 2023), the MGC sets up automated TLS whenver a HTTPRoute is created. There are currently no hooks into the creation of automated TLS certificates, nor user-exposed configuration options in terms ACME providers, custom renewal windows/schedules, key size / algorithm and other configuration options available via `cert-manager`.
```
maleck13, line 7:
```suggestion
Gateway Administrators require a comprehensive way to manage Automated, ACME-based TLS for Gateways. As gateways multiply and scale, there's a need for a flexible, policy-based API for handling Automatic TLS. Currently (July 2023), the MGC sets up automated TLS whenever a Gateway listener is created. There are currently no hooks into the creation of automated TLS certificates, nor user-exposed configuration options in terms ACME providers, custom renewal windows/schedules, key size / algorithm and other configuration options available via `cert-manager`.
```

docs/proposals/TLSPolicy.md
```diff
@@ -0,0 +1,83 @@
...
+
+1. **Policy**
+
+```yaml
+apiVersion: kuadrant.io/v1alpha1
+kind: TLSPolicy
+spec:
+  targetRef: # defaults to gateway gvk and currrent namespace
+    name: gateway-name
+  duration: 2160h # 90d - certificate validity period
```
david-martin, line 66: I'm assuming these fields map to fields in the [Certificate spec](https://cert-manager.io/docs/reference/api-docs/#cert-manager.io/v1.Certificate)?

Would be good to have sensible defaults (e.g. what's used now) to ease the 'getting started' experience.

Open question if the TLSPolicy should embed the types directly from cert-manager (hard dependency) or copy them.

docs/proposals/TLSPolicy.md
```diff
@@ -0,0 +1,83 @@
...
+## Example Policies
+
+1. **Policy**
+
+```yaml
+apiVersion: kuadrant.io/v1alpha1
+kind: TLSPolicy
+spec:
+  targetRef: # defaults to gateway gvk and currrent namespace
+    name: gateway-name
```
david-martin, line 65: Targeting Gateways makes sense until a different use case is clear.

docs/proposals/TLSPolicy.md
```diff
@@ -0,0 +1,83 @@
...
+- Facilitate easy selection of a TLS provider.
+- Allow admins to set the renewal window for TLS certificates.
+- Enable the use of different TLS strategies.
+- Consider additional TLS configurations such as enforcing HTTPS listeners and supporting wildcard hosts.
+
+## Non-Goals
+
+- Detailed implementation of the API.
+- Developing a working prototype of the API.
+- Identifying specific technologies or tools for implementing the API.
```
david-martin, line 36:
```suggestion
- Identifying specific technologies or tools for implementing the API.
- Abstracting cert-manager Issuer/ClusterIssuer configuration.
```

docs/proposals/TLSPolicy.md
jasonmadigan: agreed @david-martin. Yeah, I've made these more closely map to the certificate spec recently. I'm hoping a working PoC will help me flush out an answer on embedding types from cert-manager or copy them, I'm unsure right now.

docs/proposals/TLSPolicy.md
```diff
@@ -0,0 +1,84 @@
...
+
+
+
+## Goals
+
+- Design an API that allows admins to handle Automatic TLS for any given gateway.
+- Facilitate easy selection of a TLS provider.
+- Allow admins to set the renewal window for TLS certificates.
+- Enable the use of different TLS strategies.
+- Consider additional TLS configurations such as enforcing HTTPS listeners and supporting wildcard hosts.
```
maleck13, line 30: I think it should be a goal that TLS is not managed unless a TLSPolicy is specified. By default TLS will be the responsibility of the gateway admin. If they want us to take that on they can create a TLSPolicy.
Perhaps not part of this, but we prob need to handle listeners that are not backed by a MZ IE something directly handled by the admin for a CNAME for example or self managed DNS. 
@mikenairn interested in your opinion here? We might not have to start here but we will need to handle a listener that is not backed by a zone in a graceful way

docs/proposals/TLSPolicy.md
```diff
@@ -0,0 +1,84 @@
...
+kind: TLSPolicy
+spec:
+  targetRef: # defaults to gateway gvk and currrent namespace
+    name: gateway-name
+  duration: 2160h # 90d - certificate validity period
+  renewBefore: 360h # 15d - certificate renewal buffer in hours
+  PrivateKey:
+    Algorithm: RSA # certman.RSAKeyAlgorithm
+    Encoding: PKCS1 # certman.PKCS1
+    Size: 2048
```
maleck13, line 72: hmm expected to see an issuer property here?

maleck13 (commented):
I think we are good to look at poc impl. I would look closely at the DNSPolicy controller as we should follow a similar pattern.

david-martin:
After reading through the proposal and doing a little research on the [cert-manager feature that automatically creates Certificates for Gateway listeners](https://cert-manager.io/docs/usage/gateway/), I have a few suggestions.

First, the cert-manager feature linked is alpha, so I don't think we should change our impl to depend on that for generating Certificates. Perhaps at a later stage it could make sense.

I'm not a fan of how you need to use [annotations](https://cert-manager.io/docs/usage/gateway/#supported-annotations) to configure all the aspects of how a Certificate is created/managed.
I prefer the CRD approach, so I'm +1 on the TLSPolicy proposal we have here.

I do really like the approach of letting the owner of the Gateway resource specify the tls secret name to use.
If we changed our implementation to do something similar, it could help decouple the tls logic from the gateway controller, and make implementing a new tlspolicy controller easier.
The gateway controller no longer needs to worry about transforming the Gateway resource to add the tls block.
It also doesn't need to trigger the Certificate creation or manage the placement of Secrets in the spoke clusters.

Instead, that would all be delegated to the new tlspolicy controller. 
When a TLSPolicy is reconciled, the targeted gateway is looked up, and any certificateRefs for listeners.
For each certificateRef, ensure the Certificate exists.
Also ensure the resulting secret, if it exists yet, is placed on the same clusters & namespace as the gateway.

This approach would need some consideration about overall Gateway status in the hub after the gateway controller has do it's bit, not knowing (or caring) if the tlspolicy controller has also done it's bit.
For example, should the Gateway status be 'Programmed' if the gateways are placed correctly, regardless of if the secrets exist & have been placed yet. I'm thinking yes, as that's a different stage of the Gateway and should be reflected in individual listener attachedRoute numbers (and perhaps in future indivudal listener status conditions if we aggregate them).

Thoughts @maleck13 @mikenairn @jasonmadigan

docs/proposals/TLSPolicy.md
mikenairn: I would need to go and check what we currently do for TLS, but I'm not seeing where ManagedZones comes into it. DNSPOlicy does require that records are created in a known hosted zone, but the TLSPolicy wouldn't since it would be (i expect) created with it's own cert manager Issuer reference, with provider credentials added in the usual cert manager issuer way to deal with challenges. 

Maybe I'm missing the point, quite possible.

docs/proposals/TLSPolicy.md
```diff
@@ -0,0 +1,84 @@
...
+- Allow admins to set the renewal window for TLS certificates.
+- Enable the use of different TLS strategies.
+- Consider additional TLS configurations such as enforcing HTTPS listeners and supporting wildcard hosts.
+
+## Non-Goals
+
+- Detailed implementation of the API.
+- Developing a working prototype of the API.
+- Identifying specific technologies or tools for implementing the API.
+- Abstracting cert-manager Issuer/ClusterIssuer configuration.
```
mikenairn, line 37: On this, is this in reality going to be a CertManagerTLSPolicy? Regardless of naming, i expect it is, so would we ever want to abstract these?

docs/proposals/TLSPolicy.md
mikenairn: My understanding would have led me to expect the same. I almost feel like that's the main thing we want to allow people to select through the policy.

TLSPolicy in dev just references the self signed cert issuer, or are we expecting that the policy would manage the creation of everything for you? :thinking: This might be nice, but probably completely unrealistic for a first pass, the options we would have to provide to create a vaid Issuer in cert manager would be huge .

docs/proposals/TLSPolicy.md
maleck13: Yeah good point, so when setting up cert-manager issuer you would provide it with a credential. The listeners hosts would have to be in the same DNS provider as with the current design only one issuer per gateway would be allowed. But I agree that we don't need to tie to the existence of a managed zone

docs/proposals/TLSPolicy.md
maleck13: yes self signed in dev. Issuer is expected with this policy as it stands

docs/proposals/TLSPolicy.md
david-martin: This is in non goals

docs/proposals/TLSPolicy.md
mikenairn: I see, yes we are limiting it to only one listener. I think this comes back to the issue that we have already talked about and being able to apply polices to listeners, rather than just at the Gateway level.

I guess we could hook it into the ManagedZone somehow :thinking:, although I'm not totally sure how that would work, or if we really want that.

docs/proposals/TLSPolicy.md
mikenairn: Yeah, i suppose since it was here it felt like it was something you aren't doing now but might do later.

If this is tightly coupled to cert-manager, I'm wondering do we need to reflect that in the naming of it, or at least make it more obvious in the wording of this proposal. Probably not, prefer just TLSPolicy tbh, but just putting it out there.

docs/proposals/TLSPolicy.md
```diff
@@ -0,0 +1,84 @@
...
+# TLSPolicy API
+Currently, TLS certificates issued by the MGC are issued and managed using `cert-manager`. Configuration for these certificates are opaque to the end-user - we'd like to gradually expose a TLSPolicy to cover a range of use cases where administrators require more flexible configuration.
+
+## Problem
+
+Gateway Administrators require a comprehensive way to manage Automated, ACME-based TLS for Gateways. As gateways multiply and scale, there's a need for a flexible, policy-based API for handling Automatic TLS. Currently (July 2023), the MGC sets up automated TLS whenever a Gateway listener is created. There are currently no hooks into the creation of automated TLS certificates, nor user-exposed configuration options in terms ACME providers, custom renewal windows/schedules, key size, algorithm and other configuration options available via `cert-manager`.
+
+## Use Cases
+
+**Choosing a TLS Provider**: As an admin, I want to select a specific ACME provider for the TLS certificates in a particular gateway.
```
mikenairn, line 11: Should this be more focused on Cert Manager Issuers that ACME providers? The majority of issuers are more than likely ACME, but not all of them. Self Signed cert issuer for instance aren't ACME.

jasonmadigan: Good point. I think it should be more focused on cert-manager issuers, rather than plain ACME ones as you say.

mikenairn:
> I'm not a fan of how you need to use [annotations](https://cert-manager.io/docs/usage/gateway/#supported-annotations) to configure all the aspects of how a Certificate is created/managed. I prefer the CRD approach, so I'm +1 on the TLSPolicy proposal we have here.

This list almost looks like what we want in the spec of the policy.

> I do really like the approach of letting the owner of the Gateway resource specify the tls secret name to use. If we changed our implementation to do something similar, it could help decouple the tls logic from the gateway controller, and make implementing a new tlspolicy controller easier. The gateway controller no longer needs to worry about transforming the Gateway resource to add the tls block. It also doesn't need to trigger the Certificate creation or manage the placement of Secrets in the spoke clusters.

Hadn't really considered that the TLSPolicy is actually going to have to directly update the spec of the target resource which I'm not sure we want to do, and we can't rely on the controller knowing about policies since we want it to work with any Gateway. Do you see that in order for a TLSPolicy to actually do anything we would require the admin to add a tls section with a secret named in the Gateway ahead of time and the policy would just produce a reconcile error result until they do?
 
> Instead, that would all be delegated to the new tlspolicy controller. When a TLSPolicy is reconciled, the targeted gateway is looked up, and any certificateRefs for listeners. For each certificateRef, ensure the Certificate exists. Also ensure the resulting secret, if it exists yet, is placed on the same clusters & namespace as the gateway.

Indeed, this is key for me in all policies we plan to create, the policy and target need to be completely decoupled.

david-martin:
> Do you see that in order for a TLSPolicy to actually do anything we would require the admin to add a tls section with a secret named in the Gateway ahead of time

Yes

> and the policy would just produce a reconcile error result until they do?

Doesn't have to be an error.
'Invalid' listeners can be skipped by the tlspolicy controller.

See the Use cases examples in https://cert-manager.io/docs/usage/gateway/#use-cases

maleck13:
> Do you see that in order for a TLSPolicy to actually do anything we would require the admin to add a tls section with a secret named in the Gateway ahead of time

This is interesting. I guess this is reasonable. And allows the state of the gateway to be better reflected in the control plane. 

So is a valid user flow here

I create  a gateway with listeners and secret refs defined with a placement label and a TLSPolicy (IE via argo) . It gets synced but is not "programmed" at that stage as the secrets wont exist on the spoke. TLSPolicy controller requests certs , cert-manager setsup the secrets. TLSPolicy controller ensures these are synced to the spokes.  So in this case if the secret already exists, does the TLSPolicy controller move on?

I create a gateway with listeners and secret refs and secrets with a placement label but no TLSPolicy. We ensure the secrets are distributed with the gateways to the spoke clusters. Later I learn about TLSPolicy. I create one but we don't do anything cause their are already secrets specified and present. 90 days later TLS certs expire.

So wondering how we deal with a secret / cert that already exists. Valid concern?

david-martin:
> So in this case if the secret already exists, does the TLSPolicy controller move on?

If the secret exists in the spoke cluster, that depends how OCM handle that when the ManifestWork specifies a resource that already exists. 
The fact the TLSPolicy was created indicates the user wants the secret to be managed. It shouldn't exist already, and if it does I would expect an error or for it to be overwritten.


> We ensure the secrets are distributed with the gateways to the spoke clusters

I don't see mgc as solving this piece unless a TLSPolicy exists.
If there's no TLSPolicy, then mgc doesn't manage or place any secrets referenced in that gateway. 
It's just the gateway that's managed in this case.

> So wondering how we deal with a secret / cert that already exists. Valid concern?

Yes, valid.
If there's a TLSPolicy, we either error or overwrite the secret.
If there's no TLSPolicy, not an mgc problem.

docs/proposals/TLSPolicy.md
jasonmadigan: I'll drop an updated copy of this when I generate the CRDs

docs/proposals/TLSPolicy.md
maleck13: There is potentially a complex relationship here
1) I choose an issuer in the TLSPolicy and that applies to all listeners
2) some of the listeners are backed by a zone in a different provider 

I think having to have a managed zone setup before using a policy may be the wrong user experience https://github.com/orgs/Kuadrant/discussions/7

mikenairn:
> @mikenairn agree. I'll close this, re-open there and we might work together to land the proposal there, given your impl. sound ok?

Sounds good to me

jasonmadigan:
OK, moving this to https://github.com/Kuadrant/architecture/pull/22 - will try to port over as much context as we can.

Issue #330, opened 2023-07-13
Consolidate Placer Implementations


Ygnas:
We are currently maintaining two separate Placer implementations within mgc - `NewOCMPlacer()` and `NewTestOCMPlacer()`.

-  `NewOCMPlacer()`, our real Placer, contains crucial logic that the gateway controller depends on even before we even expect the operational control manager (OCM) to be running.
- `NewTestOCMPlacer()`, doesn't incorporate this logic, creating a discrepancy between our production and test environments.

Let's capture this issue and consider ways to consolidate these two Placer implementations into one, where possible, without losing any critical functionality.

@david-martin in https://github.com/Kuadrant/multicluster-gateway-controller/pull/329#discussion_r1261333413

david-martin:
Thanks for capturing this @Ygnas 

> doesn't incorporate this logic, creating a discrepancy between our production and test environments.

This is key.
The boundary of the 'fake' piece should be as close as possible to what we're trying to mock out. In this case it's whatever OCM would do, not what our OCM placer implementation would do.
This may need some thinking on how best to achieve that without duplicating code, and without being overly burdensome on those that would write & maintain tests.

philbrookes:
This issue is stale because it has been open for 30 days with no activity.

makslion:
It would also be nice to have this issue enable DNSPolicy integration tests to follow more `real life` path. Right now there is a workaround that manually updates GatewayCR (placer can't handle it) status and references invalid gateway class to prevent reconciliation. Introduced in [this](https://github.com/Kuadrant/multicluster-gateway-controller/pull/465) PR

philbrookes:
This issue is stale because it has been open for 60 days with no activity.

philbrookes:
This issue was closed because it has been inactive for 30 days since being marked as stale.

philbrookes:
This issue is stale because it has been open for 60 days with no activity.

philbrookes:
This issue was closed because it has been inactive for 30 days since being marked as stale.

PR #346, opened 2023-07-18
Uniform unit tests

Ygnas:


pkg/_internal/policy/policy_test.go
```diff
@@ -5,60 +5,73 @@ package policy
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			expected := GetTargetRefValueFromPolicy(tt.args.policy)
+			if tt.want != expected {
+				t.Errorf("GetTargetRefValueFromPolicy returned %v; expected %v", tt.want, expected)
+			}
+		})
+	}
```
makslion, line 76: two questions: 
1. The `_internal`  package has unit cases and they are not being executed during the `make test-unit`.  Should we run those tests? If not maybe we should get rid of them? 
2. If we are refactoring one unit test file in here, should we refactor all of them? 
And hardcoded values - especially the ones that are being used for the verification.

pkg/controllers/gateway/cluster_eventhandler_test.go
```diff
@@ -16,40 +16,50 @@ import (
...
 		enqueuedGateways []gatewayv1beta1.Gateway
 	}{
 		{
-			name: "Queued one",
+			name:   "Queued one",
+			scheme: testutil.GetValidTestScheme(),
 			gateways: []gatewayv1beta1.Gateway{
 				{
 					ObjectMeta: metav1.ObjectMeta{
 						Name:      "test-gateway",
```
makslion, line 47: Do we need this name at all?

pkg/controllers/gateway/gateway_controller_test.go
```diff
@@ -606,9 +609,9 @@ func TestGatewayReconciler_reconcileTLS(t *testing.T) {
 			r := &GatewayReconciler{
 				Client:       tt.fields.Client,
 				Scheme:       tt.fields.Scheme,
-				Certificates: test.NewTestCertificateService(tt.fields.Client),
-				HostService:  test.NewTestHostService(tt.fields.Client),
-				Placement:    test.NewTestGatewayPlacer(),
+				Certificates: faketls.NewTestCertificateService(tt.fields.Client),
+				HostService:  fakedns.NewTestHostService(tt.fields.Client),
+				Placement:    fakeplacement.NewTestGatewayPlacer(),
```
makslion, line 614: Do we need [this](https://github.com/Kuadrant/multicluster-gateway-controller/blob/3deb7ec77a561be047e05ba150e71cdef89d58b5/pkg/controllers/gateway/gateway_controller_test.go#L746) any more? I'd reuse the Pointer helper function just like [here](https://github.com/Kuadrant/multicluster-gateway-controller/blob/3deb7ec77a561be047e05ba150e71cdef89d58b5/pkg/_internal/policy/policy_test.go#L61)

pkg/controllers/gateway/params_test.go
```diff
@@ -8,20 +8,18 @@ import (
...
 )
 
 func TestGetParams(t *testing.T) {
 	cases := []struct {
-		name string
-
+		name         string
 		gatewayClass *gatewayv1beta1.GatewayClass
 		paramsObj    client.Object
 		assertParams func(*Params, error) error
```
makslion, line 25: This is neat assertion. Could we reuse it in other assertion functions and move this implementation to the helper package?

pkg/controllers/gateway/params_test.go
```diff
@@ -37,14 +35,14 @@ func TestGetParams(t *testing.T) {
...
 						Kind:      "ConfigMap",
 						Name:      "test-params",
-						Namespace: testutil.Pointer(gatewayv1beta1.Namespace("test-ns")),
+						Namespace: testutil.Pointer(gatewayv1beta1.Namespace(testutil.Namespace)),
 					},
 				},
 			},
 			paramsObj: &corev1.ConfigMap{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      "test-params",
```
makslion, line 44: As long as the name is the same it should be fine, but again - could it be defined somewhere once and then the variable reused?

pkg/dns/service_test.go
```diff
@@ -26,24 +26,22 @@ import (
...
-						Namespace: "test",
-					},
-					Spec: v1alpha1.ManagedZoneSpec{
-						DomainName: "b.c.com",
-					},
-				}
+			MZ: &v1alpha1.ManagedZone{
+				ObjectMeta: v1.ObjectMeta{
+					Name:      "b.c.com",
+					Namespace: "test",
```
makslion, line 40: a hardcoded namespace that seems to be important for the test case.

pkg/dns/target_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build unit
+
 package dns
 
 import (
```
makslion, line 5: For this file - hardcoded names of the CRs - and the clusters array seems to consist of the same elements shuffled around. There must be a way to define each cluster once and then throw it into the test case

test/util/helper.go
```diff
@@ -1,5 +1,3 @@
-//go:build unit
```
makslion, line 1: instead of getting rid of it because integration needs it, could we do the `//go:build unit || integration`?

Ygnas: Didn't even know this is possible. Thanks

pkg/dns/service_test.go
```diff
@@ -178,7 +170,7 @@ func TestDNS_GetDNSRecords(t *testing.T) {
 			gw := traffic.NewGateway(tc.Gateway)
 			f := fake.NewClientBuilder().WithScheme(testScheme(t)).WithObjects(tc.DNSRecord).Build()
 			s := dns.NewService(f)
-			record, err := s.GetDNSRecord(context.TODO(), tc.SubDomain, tc.MZ(), gw)
+			record, err := s.GetDNSRecord(context.TODO(), tc.SubDomain, tc.MZ, gw)
 			tc.Assert(t, record, err)
```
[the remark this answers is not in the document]
makslion: no, we can't

pkg/_internal/policy/policy_test.go
mikenairn: Good question. I don't think we want to delete the tests, but they do indeed not run:
```
$ go test ./pkg/_internal/... -tags=unit
go: warning: "./pkg/_internal/..." matched no packages
no packages to test
```

It must be down to the build and the fact it's got this underscore in the name, if you rename it, the tests work fine:

```
$ go test -count=1 ./pkg/internal/... -tags=unit
?       github.com/Kuadrant/multicluster-gateway-controller/pkg/internal/clusterSecret  [no test files]
?       github.com/Kuadrant/multicluster-gateway-controller/pkg/internal/conditions     [no test files]
?       github.com/Kuadrant/multicluster-gateway-controller/pkg/internal/controller     [no test files]
ok      github.com/Kuadrant/multicluster-gateway-controller/pkg/internal/env    0.002s
--- FAIL: TestGracefulDelete (0.00s)
    --- FAIL: TestGracefulDelete/deleting_in_future_sets_expected_annotation (0.00s)
        gracePeriod_test.go:45: expected graceful delete error to be nil, got: grace period has not yet expired
FAIL
FAIL    github.com/Kuadrant/multicluster-gateway-controller/pkg/internal/gracePeriod    0.012s
ok      github.com/Kuadrant/multicluster-gateway-controller/pkg/internal/metadata       0.004s
ok      github.com/Kuadrant/multicluster-gateway-controller/pkg/internal/policy 0.011s
?       github.com/Kuadrant/multicluster-gateway-controller/pkg/internal/slice  [no test files]
FAIL
```
I'm not up on these internal packages, and i cant really find any info about them, but other repos don't use underscore in the name. @maleck13 You added that, what was the reason for the underscore in the package name there? Was that required?

pkg/placement/fake/ocm.go
```diff
@@ -0,0 +1,67 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"sigs.k8s.io/gateway-api/apis/v1beta1"
+
+	"github.com/Kuadrant/multicluster-gateway-controller/pkg/dns"
+	testutil "github.com/Kuadrant/multicluster-gateway-controller/test/util"
+)
+
+type FakeGatewayPlacer struct{}
```
david-martin, line 17: Is this different than https://github.com/Kuadrant/multicluster-gateway-controller/blob/e9870c7a822920d24df19fdcf9bb941ebbb165f8/test/integration/helper_test.go#L41 ?

Makefile
```diff
@@ -68,15 +68,14 @@ lint: ## Run golangci-lint against code.
 
 .PHONY: test-unit
 test-unit: manifests generate fmt vet envtest ## Run unit tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test ./... -tags=unit -coverprofile cover-unit.out -v
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $(shell find ./pkg/_internal -mindepth 1  -type d) ./...  -tags=unit -coverprofile cover-unit.out
```
david-martin, line 71: I see how this solves the problem of running the tests.
Why not just rename the folder?
Is there something special about the name _internal that it has to be kept that way?

david-martin (commented):
There's a lot going on in this PR, so I'm hesitant to give it more thorough review without understanding the rationale for the changes.
Perhaps I'm not the best person for a thorough review.

I didn't see anything in the PR description, but I think this PR is related to #282 ?
Or at least 1 of the tasks in there?

My best interpretation at this time is that the PR does the following:
1. Fixes some unit tests that were broken
2. Includes unit tests from the _internal folder in the test run as they weren't being run before
3. A lot of refactoring of existing test code (no new tests). I'm not sure the reason for these changes
4. New fake packages with mock implementations of some interfaces. I couldn't see where these were being used, if at all.
5. Anything else I missed

I'm good with 1 & 2 (left some comments on those changes)

I'm unsure about 3, 4 & 5. 
Perhaps @makslion is better suited to review those changes.
I'd argue for having a separate PR to land 1 & 2 as they seem more important/needed right now,
then giving other changes the time they need to be reviewed & understood

Makefile
```diff
@@ -68,15 +68,14 @@ lint: ## Run golangci-lint against code.
...
 
 .PHONY: test-integration
 test-integration: ginkgo manifests generate fmt vet envtest ## Run integration tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" $(GINKGO) -tags=integration -v ./test/integration
 
 .PHONY: test
-test: manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test ./... -tags=integration,unit -coverprofile cover.out
+test: test-unit test-integration ## Run tests.
```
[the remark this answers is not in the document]
Ygnas: Isn't calling the other tests that include them not sufficient? And one of them requires `ginkgo`

pkg/placement/fake/ocm.go
Ygnas: Yes, there are 3 implementations of it.  I'm looking at it now, but it will be #330?

Makefile
mikenairn: Agree, we need to see if we can just rename that directory. I'm not sure of the reason behind using the underscore there. @maleck13 might know better. Other repos that i can see have `internal` package directories, but they don't contain the underscore see [cert-manager](https://github.com/cert-manager/cert-manager/tree/master/internal ) for example.

Probably a follow up issue/pr to resolve that particular issue would be better.

Makefile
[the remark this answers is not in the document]
makslion: those dependencies will be called in each of the sub calls

pkg/placement/fake/ocm.go
makslion: Yes, it is. And #330 would addres it

Makefile
makslion: we could do either way. I, personally, don't have a preference there.

mikenairn: That said, if this is a temporary workaround and it's going to make the tests run on CI, I'm OK with just doing it this way for now, although a bit odd.

makslion:
> There's a lot going on in this PR, so I'm hesitant to give it more thorough review without understanding the rationale for the changes. Perhaps I'm not the best person for a thorough review.
> 
> I didn't see anything in the PR description, but I think this PR is related to #282 ? Or at least 1 of the tasks in there?
> 
> My best interpretation at this time is that the PR does the following:
> 
> 1. Fixes some unit tests that were broken
> 2. Includes unit tests from the _internal folder in the test run as they weren't being run before
> 3. A lot of refactoring of existing test code (no new tests). I'm not sure the reason for these changes
> 4. New fake packages with mock implementations of some interfaces. I couldn't see where these were being used, if at all.
> 5. Anything else I missed
> 
> I'm good with 1 & 2 (left some comments on those changes)
> 
> I'm unsure about 3, 4 & 5. Perhaps @makslion is better suited to review those changes. I'd argue for having a separate PR to land 1 & 2 as they seem more important/needed right now, then giving other changes the time they need to be reviewed & understood

This PR is from the #346 issue, yes. 
The 1 & 2 - the only broken test I'm aware of is one from the `_internal` and it is a problem only because we are trying to run those now. Initially, this PR was about refactoring and making unit tests pretty, more readable and logical (the 3) Reasoning would be to make a foundation for unit testing in the future and present a "pattern" to be followed. Unit tests aren't something that gets a lot of eyes, but I'd still go for having them as readable and fault-proof as feasible. 
The 4 They aren't new - moving things around and that's all. 
The 5 - not sure you missed anything. But would like someone's else eyes on the "make test pretty" part as it heavily resides in the "personal preferences" field.

pkg/placement/fake/ocm.go
david-martin: Ack

david-martin:
> "make test pretty" part as it heavily resides in the "personal preferences" field

@mikenairn do you have any strong preference on this aspect of the PR changes?

I'd try to stay away from changing anything that can be attributed to personal preference,
and instead follow some standards or best practices.

david-martin:
> The 4 They aren't new - moving things around and that's all.

@makslion Ack. I see it now how they are moved around.

makslion:
> > "make test pretty" part as it heavily resides in the "personal preferences" field
> 
> @mikenairn do you have any strong preference on this aspect of the PR changes?
> 
> I'd try to stay away from changing anything that can be attributed to personal preference, and instead follow some standards or best practices.

Would you suggest we drop any of the current changes as not part of the best practices?

pkg/_internal/metadata/annotations_test.go
```diff
@@ -292,3 +225,31 @@ func Test_hasAnnotation(t *testing.T) {
 		})
 	}
 }
+
+func assertAnnotations(expectedAnnotations map[string]string) func(obj metav1.Object, t *testing.T) {
```
david-martin, line 229: 👍 reduces code repetition

pkg/_internal/metadata/finalizers_test.go
```diff
@@ -278,3 +220,25 @@ func Test_hasFinalizer(t *testing.T) {
 		})
 	}
 }
+
+func assertFinalizers(expectedFinalizers []string) func(obj metav1.Object, t *testing.T) {
```
david-martin, line 224: 👍 reduces code repetition

pkg/_internal/policy/policy_test.go
```diff
@@ -9,55 +9,70 @@ import (
...
-		res := GetTargetRefValueFromPolicy(policy)
-		expected := "test-gateway,test-gateway-ns"
-		if res != expected {
-			t.Errorf("GetTargetRefValueFromPolicy returned %v; expected %v", res, expected)
-		}
-	})
+			want: "test-gateway,test-gateway-ns",
+		},
+	}
+	for _, tt := range tests {
```
david-martin, line 70: Looks like we're changing the t.Run approach in this test to a struct that is iterated over.
The struct approach seems to be the pattern in use elsewhere in the repo, so seems reasonable to make this change.

pkg/controllers/dnspolicy/dns_helper_test.go
```diff
@@ -1076,3 +975,23 @@ func Test_dnsHelper_getManagedHosts(t *testing.T) {
 		})
 	}
 }
+
+func assertSub(domain string, subdomain string, err string) func(t *testing.T, expectedzone *v1alpha1.ManagedZone, expectedsubdomain string, expectedErr error) {
```
david-martin, line 979: 👍 reduces code repetition

test/util/helper.go
```diff
@@ -85,6 +85,17 @@ func AssertNoErrorReconciliation() func(res ctrl.Result, err error, t *testing.T
 	}
 }
 
+func AssertError(expectedError string) func(t *testing.T, err error) {
```
david-martin, line 88: 👍 reduces code repetition

david-martin:
> Would you suggest we drop any of the current changes as not part of the best practices?

Did another pass to better understand the reason behind the different changes.
No, I wouldn't suggest dropping any changes.

Something to think about though:
Hypothetically, the amount of tests 6 months from now have doubled.
The new tests are added by various people, and follow different test patterns, some have test code duplication, some have test helpers, there are new mocks/fakes that look different that existing ones and some tests don't run in the CI process.

How could we avoid such a scenario (and also avoid the need for another PR like this to fix things up)?
Even if we're not following some explicitly published best practices, it feels like there are practices that are being tended towards in this project.


Anyhow, I have no issue with these changes landing.
I will put it on hold in case @mikenairn has any other feedback first before merging.

/lgtm
/approve
/hold

mikenairn:
> > "make test pretty" part as it heavily resides in the "personal preferences" field
> 
> @mikenairn do you have any strong preference on this aspect of the PR changes?

I have no string preference, i think  "make test pretty" is subjective, and IMO golang tests are never going to be. Tests in general, again IMO, don't need to be overly re-factored or optimised and clearly seeing the inputs and expected outputs of specs is more valuable.

Changes look fine to me.

/approve

PR #349, opened 2023-07-18
Gh 271

philbrookes:
Update DNS Policy to have health check spec, as follows:
```
healthCheck:
    expectedResponses:
    - 200
    - 201
    additionalHeaders:
    - name: header
      value: content
    endpoint: /healthz
    failureThreshold: 5
    port: 443
    protocol: https
```

These values are used in connection with a gateway and it's relevant DNS Records to create Health Check CRs.

The health check CRs are read by the health check controller and queuers are created for each of these which will queue a probe check of the endpoint every 60 seconds.

This probe check will update the status of the health check CR with the result of the probe.

### Verification
Create a cluster and follow the [walkthrough](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/docs/how-to/ocm-control-plane-walkthrough.md) as far as deploying a workload to the spoke cluster.

Edit the DNS Policy:
```
kubectl edit dnspolicy prod-web -n multi-cluster-gateways
```

Update the DNS policy spec with the following:
```
healthCheck:
    additionalHeaders:
    - name: additional
      value: headers
    endpoint: /
    failureThreshold: 5
    port: 443
    protocol: https
```

With no expectedResponses defined, the controller will default to expecting either 200 or 201.

Wait for 60 seconds and see that the status of the health check CR is unhealthy:
```
kubectl get dnshealthcheckprobes -A
NAMESPACE                NAME               HEALTHY   LAST CHECKED
multi-cluster-gateways   172.32.200.0-api   false     53s
```

Edit the DNS Policy and add the following to the health check spec:
```
    expectedResponses:
    - 104
```

**N.B. 104 is required due to the complications of curling the docker container from within the docker container - hopefully we can discover a better solution to this allowing us to use 200**

Wait for the probe to execute (upto 60 seconds) and see the status of the health check is healthy:
```
kubectl get dnshealthcheckprobes -A
NAMESPACE                NAME               HEALTHY   LAST CHECKED
multi-cluster-gateways   172.32.200.0-api   true      3s
```

Now, delete the health check spec of the DNS Policy, then see that  no health check probes exist:
```
kubectl get dnshealthcheckprobes -A
No resources found
```


pkg/apis/v1alpha1/dnshealthcheckprobe_types.go
```diff
@@ -0,0 +1,83 @@
...
+type DNSHealthCheckProbeSpec struct {
+	Port              int               `json:"port,omitempty"`
+	Host              string            `json:"host,omitempty"`
+	Address           string            `json:"address,omitempty"`
+	Path              string            `json:"path,omitempty"`
+	Protocol          HealthProtocol    `json:"protocol,omitempty"`
+	Interval          metav1.Duration   `json:"interval,omitempty"`
+	AdditionalHeaders AdditionalHeaders `json:"additionalHeaders,omitempty"`
+	FailureThreshold  *int              `json:"failureThreshold,omitempty"`
+	ExpectedReponses  []int             `json:"ExpectedReponses,omitempty"`
```
sergioifg94, line 33:
```suggestion
	ExpectedReponses  []int             `json:"expectedReponses,omitempty"`
```

Typo in the `json` flag

sergioifg94 (changes requested):
Code changes look good, just found the typo below

pkg/apis/v1alpha1/dnshealthcheckprobe_types.go
philbrookes: nice, thanks. Fixed now.

philbrookes:
ping @mikenairn

cmd/controller/main.go
```diff
@@ -157,6 +174,15 @@ func main() {
...
 	}
+
+	if err = (&dnshealthcheckprobe.DNSHealthCheckProbeReconciler{
+		Client:        mgr.GetClient(),
+		HealthMonitor: healthMonitor,
+		Queue:         healthCheckQueue,
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "DNSHealthCheckProbe")
+		os.Exit(1)
+	}
```
mikenairn, line 185: What's the purpose of having this started here and in the health command? Is the intention to run that separately or as part of the controller, or is the health one to be run on cluster?

cmd/health/main.go
```diff
@@ -0,0 +1,75 @@
...
+
+	utilruntime.Must(v1alpha1.AddToScheme(clientgoscheme.Scheme))
+	//+kubebuilder:scaffold:scheme
+}
+
+func main() {
+	var metricsAddr string
+	var probeAddr string
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
```
mikenairn, line 33: This will conflict with the defaults for the controller. Depends what the intention of this is whether that's an issue or not.

pkg/controllers/dnshealthcheckprobe/dnshealthcheckprobe_controller.go
```diff
@@ -0,0 +1,120 @@
...
+			Protocol:          protocol,
+			IPAddress:         previous.Spec.Address,
+			AdditionalHeaders: previous.Spec.AdditionalHeaders,
+			ExpectedReponses:  previous.Spec.ExpectedReponses,
+			Notifier:          notifier,
+			Queue:             r.Queue,
+		})
+	}
+
+	return ctrl.Result{}, nil
```
mikenairn, line 104: Should the status not be getting updated here somewhere?

pkg/controllers/dnshealthcheckprobe/dnsheathcheckprobe_controller_test.go
```diff
@@ -0,0 +1,90 @@
...
+					return false
+				}
+				return probeObj.Status.LastCheckedAt.Time.After(lastUpdate.Time)
+			}, timeout+(time.Second*20), interval).Should(BeTrue())
+
+			Expect(probeObj.Status.Healthy).Should(BeFalse())
+			Expect(probeObj.Status.Reason).Should(Equal("Status code: 500"))
+		})
+	})
+})
```
mikenairn, line 90: This test should be moved into `test/integeration`, also needs a `//go:build integration` build tag added or it won't run executed by CI.

pkg/controllers/dnshealthcheckprobe/notifier.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"context"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/Kuadrant/multicluster-gateway-controller/pkg/apis/v1alpha1"
+	"github.com/Kuadrant/multicluster-gateway-controller/pkg/health"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
mikenairn, line 11: nit: grouping and order of imports is inconsistent with others.

pkg/controllers/dnshealthcheckprobe/notifier.go
```diff
@@ -0,0 +1,51 @@
...
+	} else {
+		probeObj.Status.ConsecutiveFailures = 0
+	}
+
+	probeObj.Status.LastCheckedAt = metav1.NewTime(result.CheckedAt)
+	probeObj.Status.Healthy = result.Healthy
+	probeObj.Status.Reason = result.Reason
+	probeObj.Status.Status = result.Status
+
+	return n.apiClient.Status().Update(ctx, probeObj)
```
mikenairn, line 50: Should this check for conflict and re-queue?

pkg/controllers/dnshealthcheckprobe/suite_test.go
```diff
@@ -0,0 +1,142 @@
+// go:build integrationhealth
```
mikenairn, line 1: Did you intend to add a different tag, or should it be `//go:build integration`?

pkg/controllers/dnshealthcheckprobe/suite_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"path/filepath"
+	"testing"
+	"time"
+
+	ctrl "sigs.k8s.io/controller-runtime"
```
mikenairn, line 13: nit: Can be with the rest of the k8s imports

pkg/controllers/dnshealthcheckprobe/suite_test.go
```diff
@@ -0,0 +1,142 @@
...
+		errCh <- http.ListenAndServe(fmt.Sprintf(":%d", s.Port), mux)
+	}()
+
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case err := <-errCh:
+		return err
+	}
+}
```
mikenairn, line 142: All the ginkgo tests should now be in `test/integeration` this won't be run in CI otherwise.

pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -116,6 +117,8 @@ func (r *DNSPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	if !controllerutil.ContainsFinalizer(dnsPolicy, DNSPolicyFinalizer) {
 		if err := r.AddFinalizer(ctx, dnsPolicy, DNSPolicyFinalizer); client.IgnoreNotFound(err) != nil {
 			return ctrl.Result{Requeue: true}, err
+		} else if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, err
```
mikenairn, line 121: Why was this required?

pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -194,6 +197,11 @@ func (r *DNSPolicyReconciler) deleteResources(ctx context.Context, dnsPolicy *v1
 			return err
 		}
 	}
+	for _, gw := range gatewayDiffObj.GatewaysWithInvalidPolicyRef {
+		if err := r.DeleteTargetBackReference(ctx, client.ObjectKeyFromObject(dnsPolicy), gw.Gateway, DNSPolicyBackRefAnnotation); err != nil {
+			return err
+		}
+	}
```
mikenairn, line 204: This shouldn't be required, the direct back ref is being removed [here](https://github.com/Kuadrant/multicluster-gateway-controller/pull/349/files#diff-162ceed42032f121fd671b5884a611158c00f31e292b7602e48d5671ca7fe01fR195-R199) when the target still exists.

This should only be getting called once for the policy we are currently deletingResources for. And the polices (plural) annotation gets dealt with here https://github.com/Kuadrant/multicluster-gateway-controller/pull/349/files#diff-162ceed42032f121fd671b5884a611158c00f31e292b7602e48d5671ca7fe01fR207

This btw is the same for all current kuadrant polices see [AuthPolicy](https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/authpolicy_controller.go#L173-L181) and https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/ratelimitpolicy_controller.go#L212-L220 which this is based. If there is some issue that this was trying to fix, I'd expect that to be an issue in all of them.

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
```diff
@@ -2,203 +2,181 @@ package dnspolicy
...
-	// Delete Health checks for each gateway no longer referred by this policy
-	for _, gw := range gwDiffObj.GatewaysWithInvalidPolicyRef {
-		log.V(1).Info("reconcileHealthChecks: gateway with invalid policy ref", "key", gw.Key())
-		_, err := r.reconcileGatewayHealthChecks(ctx, gw.Gateway, nil)
+	var expectedProbes []*v1alpha1.DNSHealthCheckProbe
+	if dnsPolicy.DeletionTimestamp != nil {
+		expectedProbes = []*v1alpha1.DNSHealthCheckProbe{}
+	} else {
+		var err error
+		expectedProbes, err = r.generateExpectedProbes(ctx, gwDiffObj, dnsPolicy)
```
mikenairn, line 27: I'd have expected this to keep the pattern of using the three groups of gateways from `gwDiffObj`  and calling an appropriate `reconcileGatewayHealthChecks` method to deal with create/updates/deletes of the probes. This is a pattern that was setup in this polices reconcilers (dnsrecord and health) based on other kuadrant polices see [RateLimit](https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/ratelimitpolicy_limits.go#L49-L87l)  and [AuthConfig](https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/authpolicy_istio_authorization_policy.go#L46-L53). I think we'd want good reason form straying from that pattern here since we are trying to keep all these polices as consistent as possible in how they work.

test/integration/dnspolicy_controller_test.go
```diff
@@ -9,14 +9,16 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
```
mikenairn, line 13: nit: Should be with the rest of the k8s imports

cmd/controller/main.go
sergioifg94: The initial idea was to run it separately as a different container running in the spoke clusters, but we dropped that in favour of having it in the hub

cmd/health/main.go
sergioifg94: I think we'll remove this file for now as we are running the probe controller alongide the MGC

pkg/controllers/dnshealthcheckprobe/dnshealthcheckprobe_controller.go
sergioifg94: The status isn't directly updated (yet) in the Reconcile but by the notifier that is passed when the probes are created:
https://github.com/philbrookes/multi-cluster-traffic-controller/blob/gh-271/pkg/controllers/dnshealthcheckprobe/notifier.go

It would be nice to have some other status block to represent whether the probe was registered or not or any other non-health related errors though

pkg/controllers/dnspolicy/dnspolicy_controller.go
philbrookes: oh I added this while trying to figure out what was wrong with the e2e tests. I'll remove it now.

pkg/controllers/dnspolicy/dnspolicy_controller.go
philbrookes: yes, this was also my clutching at straws for the e2e tests.

test/integration/dnspolicy_controller_test.go
philbrookes: done

pkg/controllers/dnshealthcheckprobe/notifier.go
philbrookes: fixed.

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
philbrookes: I'm struggling to see a way to use that pattern, and have the resulting process be as reliable. The current code will look at all gateways referenced by the policy (new or existing) and generate the CRs as expected for those gateways based on the current DNS Policy configuration.

It will then take those and create or exist as expected, and remove any created as a result of this DNS Policy that are no longer expected.

As I have understood your comment, you're suggesting I change this to:
- Creates in a for loop for "newly targeted gateways"
- Updates in a for loop for "already targeted gateways"
- Deletes in a for loop for "no longer targeted gateways"

However, this feels inferior to me, as it leaves a greater possibility of the create / delete events to not be accurately triggered and have either new health checks not created, or expired health checks not removed. Have I misunderstood somewhere?

philbrookes: oh, I think I see what you mean. Have it call a function per gateway which will generate a list of expected health checks per gateway and create / update / delete as expected, but treating each gateway separately?

philbrookes: discussed on a call, I will update this to match the existing policy reconcilers.

pkg/controllers/dnspolicy/dnspolicy_controller.go
mikenairn: You keeping this in? Just interested to know what it solves if we do need it, not sure we do this anywhere else.

mikenairn (approved):

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
```diff
@@ -2,203 +2,164 @@ package dnspolicy
...
+						Protocol:          *dnsPolicy.Spec.HealthCheck.Protocol,
+						Interval:          metav1.Duration{Duration: 60 * time.Second},
+						AdditionalHeaders: dnsPolicy.Spec.HealthCheck.AdditionalHeaders,
+						FailureThreshold:  dnsPolicy.Spec.HealthCheck.FailureThreshold,
+						ExpectedReponses:  dnsPolicy.Spec.HealthCheck.ExpectedResponses,
+					},
+				}
+				healthChecks = append(healthChecks, healthCheck)
+			}
+		}
```
mikenairn, line 162: @philbrookes I was thinking about this again, and was wondering if it would be better here, rather than processing a gateways listeners (which the DNSrecord reconcile is already doing), to instead just work off the list of DNSRecords that were created by this DNSPolicy?

HealthChecks are more tightly coupled to DNSRecords than the Gateway itself, and since we have just processed the records to create/update/delete them, it seems logical, to me at least, that we could just iterate round them here.

Note: This might be easier once we have these changes in https://github.com/mikenairn/multi-cluster-traffic-controller/commit/655596b5845d8c4ce4ef0b7848846f70b989acf8.

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
```diff
@@ -2,203 +2,164 @@ package dnspolicy
...
+	} else {
+		for _, p := range existingProbes.Items {
+			if !slice.Contains(expectedProbes, func(expectedProbe *v1alpha1.DNSHealthCheckProbe) bool {
+				return expectedProbe.Name == p.Name && expectedProbe.Namespace == p.Namespace
+			}) {
+				if err := r.Client().Delete(ctx, &p); err != nil {
+					return err
+				}
+			}
 		}
```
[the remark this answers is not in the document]
mikenairn: Looks to be rolled into this https://github.com/Kuadrant/multicluster-gateway-controller/pull/363

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
philbrookes: for this to work, the dns policy would have to attach to the DNS Record rather than the gateway, otherwise how would the DNS Record reconciler know what configuration to use in the DNS Health Check Probe for each endpoint?

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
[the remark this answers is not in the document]
philbrookes: ok, I think this can be tracked in a separate issue.

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
mikenairn: You have the policy here, you are doing this as part of the policy reconciliation. I'm just saying you don't need to get records by looking them up based on the listeners of the gateway, the DNSRecord reconcile has already done that, and any difference in the logic in the two could result in different things happening. You could list the dnsrecords in the same way this delete is doing here https://github.com/Kuadrant/multicluster-gateway-controller/pull/379/files#diff-560235894bef952ec7cc558026db8ebfadbab2c1667248e75eb0c575d2ff2e0bR124-R134 and create your probes for each.

philbrookes: ah, I see. I thought you were suggesting moving this logic into the DNS Record reconciler. Yes, this makes sense, I'll add it.

mikenairn: OK, maybe there are other considerations. I wasn't suggesting we had to change it for this PR to merge though, so we can maybe think on that one.

philbrookes: oh no, I take that back, the port is specified in the DNS Policy. I'll add this change now.

pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -76,8 +76,6 @@ func (r *DNSPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	if err := r.Client().Get(ctx, req.NamespacedName, previous); err != nil {
 		if err := client.IgnoreNotFound(err); err == nil {
 			return ctrl.Result{}, nil
-		} else {
-			return ctrl.Result{}, err
 		}
```
mikenairn, line 79: Did you mean to remove this?

PR #351, opened 2023-07-19
Remove check for no placed clusters in dnspolicy

mikenairn:
Removes the logic in the dnspolicy reconcile to break out immediately if there are no placed clusters. This is incorrectly preventing DNSRecords that are already created from being removed if the target Gateway is removed from all clusters, or all clusters are deleted.


david-martin:
/approve
/lgtm

This makes sense.
The reconcile logic should handle an empty list of clusters & bring the current state towards the desired state.

PR #355, opened 2023-07-19
Update RecordCount on ManagedZone

mikenairn:
The managed zone API response doesn't contain information about the number of records in a zone, instead we can just make a request to get all the records and set it to the length.

Other minor updates:

* Fix issue with nameservers not updating correctly
* Improve handling of ManagedZone and DNSRecord deletion
* Use pointer receiver
* Order, public before private
* Use `ensureTrailingDot`
* Add `toManagedZoneOutput` to convert a google ManagedZone into the mgc expected output


mikenairn:
@R-Lawton Leaving it up to you to merge this or not

Issue #286, opened 2023-06-29
Improve getting started experience


maleck13:
**What**

Currently out best "Getting started guide" is to go through the OCM walkthrough. While this works, it is a lot of work. 

**Proposal**

create a page similar to this:

https://open-cluster-management.io/getting-started/quick-start/

that has pre-reqs and a script based setup. This page should also explain what we are going to setup and perhaps link to the more detailed walkthrough doc

Prereqs:
- Kind installed

**Create a new script ** that is for locally trying the system

that prompts for the root domain, and AWS creds if not present
installs OCM and Istio (which we already do)
sets up the ground work in the OCM walkthough (IE two kind clusters, Istio installed, kuadrant (MGC and Data plane) installed via latest images and hooked up, clusterset set up and bound, placement created that targets the clusterset)
Essentially once it is done, they get to create a gateway, create a HTTPRoute and see it work. We should then have follow ons for using the RateLimit policy and creating / updating the DNS Policy 

- [x] new script to get you to the point you can create a gateway
- [ ] update the getting started page for kuadrant.io and github readme ( Link to the follow ons ie (rate limiting))

david-martin:
@maleck13 Is this different than #280, which is part of the [Kuadrant.io Refresh](https://github.com/Kuadrant/multicluster-gateway-controller/issues/205)

maleck13:
yeah I would say this is a quick start whereas the install guide should prob be more focused on actual installation rather than setting up placements and clusters for you

laurafitzgerald:
As part of this task i'm starting with stepping through https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/docs/how-to/ocm-control-plane-walkthrough.md
I've made a copy of that document here https://docs.google.com/document/d/1cLtLNT9HqGqfQCntpEUIqaX22LMUrHQ8VrRuqIqHswc/edit where myself and @eoinfennessy are making notes about potential improvements as well as discussion about a script.

laurafitzgerald:
Hey @maleck13 or anyone. i've been looking at the existing script local-setup.sh to see what can be done for a quick start one line setup. There's a lot that that existing one relies on locally. configuration files and helm charts kept in the config directory and helper functions. before starting into it I want to double check on an approach. i'm new to helm so wondering if those resources can be gotten dynamically somehow? the alternative feels like we'll end up with a massive script which duplicates some of the existing work. did you have any other approach in mind as a preference for this?

Issue #256, opened 2023-06-15
Separate cluster creation and provisioning in local-setup


mikenairn:
**What**

Separate the cluster creation and provisioning parts of local-setup so that they can be run independently of one another.

**Why**

Currently local-setup will delete any existing clusters, and one by one, create a kind cluster and install the relevant pieces onto it for the kind of cluster it is intended to be (control plane vs data plane). If at any point local-setup fails, even after successfully creating a cluster, you have to just run local-setup again starting from scratch. A better option here is to have cluster creation come first, and then run a tailored run list for each cluster based on it's type. This run list should be idempotent and can run over and over if needs be.

**How**

WIP PR here for inspiration, but needs more discussion https://github.com/Kuadrant/multicluster-gateway-controller/pull/102. Depends how heavily we want to invest in using the likes of kustomize for everything.

- [ ] Have the kind cluster setup run separately and allow it to be run independently 
- [ ] Have a kind cluster tear down

Issue #119, opened 2023-03-13
Change local development subnet


sergioifg94:
## Description

The current development subnet is `172.32.0.0/16`. This might conflict as it's owned by [TMobile](https://search.arin.net/rdap/?query=172.32.0.1)

## How

Update the local-setup and the resources/code that reference the subnet to an address in [the private space](https://search.arin.net/rdap/?query=172.16.0.1)

philbrookes:
This issue is stale because it has been open for 30 days with no activity.

philbrookes:
This issue was closed because it has been inactive for 30 days since being marked as stale.

PR #362, opened 2023-07-21
Quickstart script

eoinfennessy:
Closes https://github.com/Kuadrant/multicluster-gateway-controller/issues/286

### What

A Quickstart setup script that can be curl-ed and piped to bash:
2. Prompts user for AWS credentials, root domain, hosted zone ID, etc.
3. Follows similar (stripped down) steps to the existing `local-setup.sh` script, but remotely builds kustomize directories and remotely sources utils scripts

### Verifcation steps.

Run `curl https://raw.githubusercontent.com/laurafitzgerald/multicluster-gateway-controller/quickstart-script/hack/quickstart-setup.sh | bash`

The script should complete successfully. 

The user will be prompted on first run to enter some env vars that are usually expected in env files.  Once entered the script should continue. 

The resulting setup should be sufficient to follow steps in [ocm-walkthrough](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/docs/how-to/ocm-control-plane-walkthrough.md#open-terminal-sessions) skipping the make-local setup command. 

### Associated Document
While this isn't created yet we aim to have something like https://open-cluster-management.io/getting-started/quick-start/ and probably under https://docs.kuadrant.io/multicluster-gateway-controller/  

The form of the document will depend on some further discussions. A slack discussion is opened about this. 

### Next Steps

@laurafitzgerald and I have been chatting about possible next steps for this and how it may work on the website, including the idea of allowing different users types (admins, developers, system architects) the flexibility to either automate next steps using additional one-liners, or walk through the steps manually in a linked guide, e.g:

> Next steps: Follow [this guide to learn how to create a gateway](http://www.kuadrant.io/) or automate gateway creation and continue to learn how to deploy an application by running the following command...


maleck13:
how do I try it? what is the command to run?

eoinfennessy:
> how do I try it? what is the command to run?

For now, you can run it with the following:

`curl https://raw.githubusercontent.com/laurafitzgerald/multicluster-gateway-controller/quickstart-script/hack/quickstart-setup.sh | bash`

jasonmadigan:
Nice! 

```bash
curl https://raw.githubusercontent.com/laurafitzgerald/multicluster-gateway-controller/quickstart-script/hack/quickstart-setup.sh | bash
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 13300  100 13300    0     0   112k      0 --:--:-- --:--:-- --:--:--  118k
Verifying that pre-requisite binaries are present on the system.
Error: kustomize version 4.5.4 or greater is required, but found 4.5.4. Please review the pre-requisite binary versions in the documentation against your local version of customize
```

Something seems to be up with one of the pre-req checks

eoinfennessy:
> Nice!
> 
> ```shell
> curl https://raw.githubusercontent.com/laurafitzgerald/multicluster-gateway-controller/quickstart-script/hack/quickstart-setup.sh | bash
>   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
>                                  Dload  Upload   Total   Spent    Left  Speed
> 100 13300  100 13300    0     0   112k      0 --:--:-- --:--:-- --:--:--  118k
> Verifying that pre-requisite binaries are present on the system.
> Error: kustomize version 4.5.4 or greater is required, but found 4.5.4. Please review the pre-requisite binary versions in the documentation against your local version of customize
> ```
> 
> Something seems to be up with one of the pre-req checks

Hmm... Will try to figure out what's going on here... These checks were a late addition and haven't been extensively tested.

jasonmadigan:
No worries - will check out again later

eoinfennessy:
> No worries - will check out again later

@jasonmadigan I think this has been fixed

jasonmadigan:
Looking better:

```
kind version is compatible: 0.20.0
helm version is compatible: 3.10.0
yq version is compatible: 4.34.2
Error: operator-sdk not found. Please review the pre-requisite versions in the documentation against your local version of operator-sdk
```

Q: do I really _need_ to have the operator-sdk installed?

hack/quickstart-setup.sh
```diff
@@ -0,0 +1,394 @@
...
+
+# Check if a binary is installed and has the expected version
+check_binary_version() {
+    local binary_name="$1"
+    local expected_version="$2"
+    local semver_pattern="$3"
+
+    # Check if the binary is installed
+    if ! command -v "$binary_name" &>/dev/null; then
+        echo "Error: $binary_name not found. Please review the pre-requisite versions in the documentation against your local version of $binary_name"
```
jasonmadigan, line 264: Where are those docs?

jasonmadigan:
```
Error: istioctl version 1.28.0 or greater is required, but found 1.18.1. Please review the pre-requisite binary versions in the documentation against your local version of istioctl
```

for macOS, all of the pre-requisites are installable via brew. `istioctl`, however, is only at 1.18.1 in brew by default. I expect this will work OK, so the version reqs could maybe do with some relaxing.

eoinfennessy:
I should probably also mention that it's possible the binary version numbers required may be a little restrictive. If the version of a binary you've been using successfully when running through ocm walkthrough falls below what is required here, let us know and we can lower the required version.

eoinfennessy:
> 

Just saw your comment after adding a note about this about above. Yep, will lower the version required for istioctl.

eoinfennessy:
> ```
> Error: istioctl version 1.28.0 or greater is required, but found 1.18.1. Please review the pre-requisite binary versions in the documentation against your local version of istioctl
> ```
> 
> for macOS, all of the pre-requisites are installable via brew. `istioctl`, however, is only at 1.18.1 in brew by default. I expect this will work OK, so the version reqs could maybe do with some relaxing.

Just saw your comment after adding a note about this about above. Yep, will lower the version required for istioctl.

laurafitzgerald:
@jasonmadigan @eoinfennessy the versions i used came from https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/hack/make/dependencies.make#L27-L36 If we update them for this case we should probably update them there as well.

jasonmadigan:
> I should probably also mention that it's possible the binary version numbers required may be a little restrictive. If the version of a binary you've been using successfully when running through ocm walkthrough falls below what is required here, let us know and we can lower the required version.

Yeah, I think I hit every single rock here when it came to the version checks ;)

I've upgraded most stuff now, but when I started I was running:

kind: 0.14.0
yq: 4.27.2
istioctl: 1.18.1 (version dep for this already dropped)

jasonmadigan:
OK, much further:

```
clusterrolebinding.rbac.authorization.k8s.io/mgc-cert-manager-controller-certificates created
clusterrolebinding.rbac.authorization.k8s.io/mgc-cert-manager-controller-certificatesigningrequests created
clusterrolebinding.rbac.authorization.k8s.io/mgc-cert-manager-controller-challenges created
clusterrolebinding.rbac.authorization.k8s.io/mgc-cert-manager-controller-clusterissuers created
clusterrolebinding.rbac.authorization.k8s.io/mgc-cert-manager-controller-ingress-shim created
clusterrolebinding.rbac.authorization.k8s.io/mgc-cert-manager-controller-issuers created
clusterrolebinding.rbac.authorization.k8s.io/mgc-cert-manager-controller-orders created
clusterrolebinding.rbac.authorization.k8s.io/mgc-cert-manager-webhook:subjectaccessreviews created
configmap/mgc-cert-manager-webhook created
service/mgc-cert-manager created
service/mgc-cert-manager-webhook created
deployment.apps/mgc-cert-manager created
deployment.apps/mgc-cert-manager-webhook created
mutatingwebhookconfiguration.admissionregistration.k8s.io/mgc-cert-manager-webhook created
validatingwebhookconfiguration.admissionregistration.k8s.io/mgc-cert-manager-webhook created
Waiting for Cert Manager deployments to be ready...
deployment.apps/mgc-cert-manager condition met
deployment.apps/mgc-cert-manager-webhook condition met
validatingwebhookconfiguration.admissionregistration.k8s.io "mgc-cert-manager-webhook" deleted
mutatingwebhookconfiguration.admissionregistration.k8s.io "mgc-cert-manager-webhook" deleted
clusterissuer.cert-manager.io/glbc-ca created
Switched to context "kind-mgc-control-plane".
Initialize local dev setup for the controller on mgc-control-plane
customresourcedefinition.apiextensions.k8s.io/dnspolicies.kuadrant.io created
customresourcedefinition.apiextensions.k8s.io/dnsrecords.kuadrant.io created
customresourcedefinition.apiextensions.k8s.io/managedzones.kuadrant.io created
namespace/multi-cluster-gateways created
secret/mgc-aws-credentials created
configmap/mgc-controller-config created
The ManagedZone "mgc-dev-mz" is invalid: spec.domainName: Required value
```

Not quite sure what that is, likely something that landed recently.

hack/quickstart-setup.sh
laurafitzgerald: This will be the doc mentioned in the associated issue. The user would have come from there to run the command. We're aiming for something similar to https://open-cluster-management.io/getting-started/quick-start/ so the prereq section should have prepared the user for this.

eoinfennessy:
> configmap/mgc-controller-config created
> The ManagedZone "mgc-dev-mz" is invalid: spec.domainName: Required value
> ```
> 
> Not quite sure what that is, likely something that landed recently.

This is unusual. What do you get for `ZONE_ROOT_DOMAIN` if you run 'k describe configmap mgc-controller-config -n multi-cluster-gateways'?

hack/quickstart-setup.sh
jasonmadigan: Sounds good

jasonmadigan:
> > ```
> > configmap/mgc-controller-config created
> > The ManagedZone "mgc-dev-mz" is invalid: spec.domainName: Required value
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Not quite sure what that is, likely something that landed recently.
> 
> This is unusual. What do you get for `ZONE_ROOT_DOMAIN` if you run `k describe configmap mgc-controller-config -n multi-cluster-gateways`?

```
Name:         mgc-controller-config
Namespace:    multi-cluster-gateways
Labels:       <none>
Annotations:  <none>

Data
====
AWS_DNS_PUBLIC_ZONE_ID:
----
fi
LOG_LEVEL:
----
1
ZONE_ROOT_DOMAIN:
----


BinaryData
====

Events:  <none>
```

Given the `fi`, perhaps something might be up with the reading of the config from the .env files

hack/quickstart-setup.sh
eoinfennessy: I don't think these exist yet. The pre-requisite version numbers should probably be listed in the quickstart guide on the website.

laurafitzgerald: I guess we want to include it somewhere under https://docs.kuadrant.io/multicluster-gateway-controller/ ? It's likely a small change so we could do this as part of this issue if folks are happy with how the script is running here.

jasonmadigan:
> > 
> 
> > Given the `fi`, perhaps something might be up with the reading of the config from the .env files
> 
> This script doesn't read anything from .env files. You either get prompted to enter the required values, or you can set these values as environment variables. Did you enter these when prompted?

No prompt, I've got these exported elsewhere - so the script continues without prompting.

I've got valid values for:

```
AWS_ACCESS_KEY_ID
AWS_SECRET_ACCESS_KEY
AWS_DNS_PUBLIC_ZONE_ID

```

I don't have these set:

```
AWS_REGION
ZONE_ROOT_DOMAIN
```

but I'm not prompted (I'm guessing this is where the issues stems from). I don't see anything obviously wrong about the check...

jasonmadigan:
D'oh, OK - we're running this non-interactively, because it's piped.

hack/quickstart-setup.sh
```diff
@@ -0,0 +1,394 @@
...
+if [[ -z "${AWS_ACCESS_KEY_ID}" ]]; then
+  echo "Enter your AWS access key ID:"
+  read AWS_ACCESS_KEY_ID
+fi
+if [[ -z "${AWS_SECRET_ACCESS_KEY}" ]]; then
+  echo "Enter your AWS secret access key:"
+  read AWS_SECRET_ACCESS_KEY
+fi
+if [[ -z "${AWS_REGION}" ]]; then
+  echo "Enter an AWS region (e.g. eu-west-1):"
```
jasonmadigan, line 313: small request: might nice to add mention of the specific env-var looked for here as output, so people know what they should export for next time if they don't want to be prompted.

laurafitzgerald: Addressed in https://github.com/Kuadrant/multicluster-gateway-controller/pull/362/commits/48694239df92160c846a7fea22cf73da59ae96c6

laurafitzgerald:
> > > 
> > 
> > 
> > > Given the `fi`, perhaps something might be up with the reading of the config from the .env files
> > 
> > 
> > This script doesn't read anything from .env files. You either get prompted to enter the required values, or you can set these values as environment variables. Did you enter these when prompted?
> 
> No prompt, I've got these exported elsewhere - so the script continues without prompting.
> 
> I've got valid values for:
> 
> ```
> AWS_ACCESS_KEY_ID
> AWS_SECRET_ACCESS_KEY
> AWS_DNS_PUBLIC_ZONE_ID
> ```
> 
> I don't have these set:
> 
> ```
> AWS_REGION
> ZONE_ROOT_DOMAIN
> ```
> 
> but I'm not prompted (I'm guessing this is where the issues stems from). I don't see anything obviously wrong about the check...

Hi @jasonmadigan can you try this again in a new terminal window. I also saw similar behaviour but a new terminal window resolved it using the command `curl -Lo mgc-quickstart.sh https://raw.githubusercontent.com/Kuadrant/multicluster-gateway-controller/d34e8e74d48731a18f6d28ab65b1ec067366611e/hack/quickstart-setup.sh && chmod +x mgc-quickstart.sh && bash mgc-quickstart.sh`

hack/quickstart-setup.sh
laurafitzgerald: Captured this in the description.

jasonmadigan:
> > > > 
> > > 
> > > 
> > > > Given the `fi`, perhaps something might be up with the reading of the config from the .env files
> > > 
> > > 
> > > This script doesn't read anything from .env files. You either get prompted to enter the required values, or you can set these values as environment variables. Did you enter these when prompted?
> > 
> > 
> > No prompt, I've got these exported elsewhere - so the script continues without prompting.
> > I've got valid values for:
> > ```
> > AWS_ACCESS_KEY_ID
> > AWS_SECRET_ACCESS_KEY
> > AWS_DNS_PUBLIC_ZONE_ID
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I don't have these set:
> > ```
> > AWS_REGION
> > ZONE_ROOT_DOMAIN
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > but I'm not prompted (I'm guessing this is where the issues stems from). I don't see anything obviously wrong about the check...
> 
> Hi @jasonmadigan can you try this again in a new terminal window. I also saw similar behaviour but a new terminal window resolved it using the command `curl -Lo mgc-quickstart.sh https://raw.githubusercontent.com/Kuadrant/multicluster-gateway-controller/d34e8e74d48731a18f6d28ab65b1ec067366611e/hack/quickstart-setup.sh && chmod +x mgc-quickstart.sh && bash mgc-quickstart.sh`

Slightly different issue for me - was down to the previous quickstart pipe command, which ran non-interactively. Running it like you've got above works (can interactively provide input that way). All good now I think.

hack/quickstart-setup.sh
```diff
@@ -16,12 +16,19 @@
 # limitations under the License.
 #
 
+BIN_DIR=$(pwd)/bin
```
laurafitzgerald, line 19: i wonder if we should put this as `/tmp/mgc/bin` as we don't know where the command will be run from. if a cleanup is run seperately and potenetially from a different dir we could end up not able to clean up things.

hack/quickstart-setup.sh
```diff
@@ -0,0 +1,362 @@
...
+
+export KIND_BIN=kind
+export YQ_BIN=yq
+
+source /dev/stdin <<< "$(curl -s https://raw.githubusercontent.com/Kuadrant/multicluster-gateway-controller/main/hack/.kindUtils)"
+source /dev/stdin <<< "$(curl -s https://raw.githubusercontent.com/Kuadrant/multicluster-gateway-controller/main/hack/.clusterUtils)"
+source /dev/stdin <<< "$(curl -s https://raw.githubusercontent.com/Kuadrant/multicluster-gateway-controller/main/hack/.cleanupUtils)"
+# TODO: Uncomment when otherwise ready to merge
+# source /dev/stdin <<< "$(curl -s https://raw.githubusercontent.com/Kuadrant/multicluster-gateway-controller/main/hack/.dependenciesUtils.sh)"
+source ./.dependenciesUtils.sh
```
laurafitzgerald, line 31: For testing can we make this the below as I'm running it outside of the MGC repo in order to ensure we're not getting any local dependencies

```suggestion
# TODO: Update to Kuadrant main before merging.
source /dev/stdin <<< "$(curl -s https://raw.githubusercontent.com/laurafitzgerald/multicluster-gateway-controller/quick-start/hack/.dependenciesUtils.sh)"
```

maleck13:
Running through this and making some notes

1) I think downloading the binaries to a tmp dir (at least some of the less common ones such as yq, istioctl and operator-sdk ) would be good. Running from scratch, it was a bit frustrating the number of deps I had to get in place. Also if we are not going to install a dep we should offer a link to the github repo?
2) I think when asking for AWS credentials in the prompt we should call out what they are for? I guess this will also be in doc.
3) minor thing but when it ran it said `deleting existing kind clusters` prob should be clear it is only removing exiting kuadrant kind clusters
4) when install finished we get `Your managed cluster kind-mgc-workload-1 has joined the Hub successfully. Visit https://open-cluster-management.io/scenarios or https://github.com/open-cluster-management-io/OCM/tree/main/solutions for next steps. ` which makes it look like I should go there. But really I think we want to link them to our guides for Gateways

While this is a batteries included install, it did get me thinking that perhaps we want some options to remove some dependencies. (not part of this but for future consideration) IE I am interested in DNSPolicy only and so there is no need for istio and cert-manager. I think potentially this will become clearer as we look at what a single cluster install should look like other than its current form of being a full OCM install

Dockerfile.quickstart
```diff
@@ -0,0 +1,34 @@
+FROM ubuntu:latest
```
jasonmadigan, line 1: can we switch this to the rhel UBI?

laurafitzgerald: we surely can. 👍

hack/quickstart-setup.sh
```diff
@@ -0,0 +1,322 @@
...
+    kubectl -n metallb-system wait --for=condition=ready pod --selector=app=metallb --timeout=300s
+}
+
+configureControlCluster() {
+
+    clusterName=${1}
+    # Ensure the current context points to the control plane cluster
+    kubectl config use-context kind-${clusterName}
+    kubectl label managedcluster kind-mgc-control-plane ingress-cluster=true
+    cat <<EOF | kubectl apply -f -
```
laurafitzgerald, line 225: @mikenairn I tried to include this under the config/quickstarthub kustomization but I was getting the error. `error: resource mapping not found for name: "kuadrant-multi-cluster-gateway-instance-per-cluster" namespace: "" from "STDIN": no matches for kind "GatewayClass" in version "gateway.networking.k8s.io/v1beta1"
ensure CRDs are installed first` even though I included it as the last item in the list of resources for that kustomization. And when I checked the cluster there was the usual istio gateway class present on the cluster. Any ideas on how I could fix this. It would be nice to include it in the kustomization rather than in the script like the others.

hack/quickstart-setup.sh
```diff
@@ -0,0 +1,322 @@
...
+apiVersion: metallb.io/v1beta1
+kind: L2Advertisement
+metadata:
+  name: empty
+  namespace: metallb-system
+EOF
+}
+
+
+deployIstio() {
```
laurafitzgerald, line 61: I'd like to remove this function because the directory now has kustomization.yaml present but it's called twice. Once for the hub and once for the spoke. 
When moving it to the relevant quickstart kustomization i saw various errors.
For Hub I faced the error `error: resource mapping not found for name: "istiocontrolplane" namespace: "istio-system" from "STDIN": no matches for kind "IstioOperator" in version "install.istio.io/v1alpha1"
ensure CRDs are installed first` similar to the gateway class
And for spoke it's called from inside the ocm functions.

hack/quickstart-setup.sh
```diff
@@ -0,0 +1,322 @@
...
+# Default config
+if [[ -z "${LOG_LEVEL}" ]]; then
+  LOG_LEVEL=1
+fi
+if [[ -z "${OCM_SINGLE}" ]]; then
+  OCM_SINGLE=true
+fi
+if [[ -z "${MGC_WORKLOAD_CLUSTERS_COUNT}" ]]; then
+  MGC_WORKLOAD_CLUSTERS_COUNT=1
+fi
```
laurafitzgerald, line 276: For this quickstart, can we be opinionated about these settings rather than exposing them?

docs/how-to/ocm-control-plane-walkthrough.md
```diff
@@ -77,89 +51,20 @@ Open two windows, which we'll refer to throughout this walkthrough as:
...
-    ```bash
-    make build-controller kind-load-controller deploy-controller
-    ```
-
-    *Alternatively, in a new window we'll call `T3`, you can run the controller locally instead by running:*
-
-    ```bash
-    (export $(cat ./controller-config.env | xargs) && export $(cat ./aws-credentials.env | xargs) && make build-controller install run-controller)
-    ```
+
```
eoinfennessy, line 67: Nice! Probably not within the scope of this PR, but would it be helpful to have a summary of the clusters/resources that have been created up to this point here? Maybe a diagram too? Both of these could potentially be reused on the website.

docs/how-to/ocm-control-plane-walkthrough.md
```diff
@@ -15,50 +15,24 @@ We will start with a single cluster and move to multiple clusters to illustrate
...
-  export MGC_SUB_DOMAIN=myapp.jbloggs.hcpapps.net
-  ```
+ | Env Var                      | Example Value               | Description                                                    |
+|------------------------------|-----------------------------|----------------------------------------------------------------|
+  | `MGC_ZONE_ROOT_DOMAIN`       | `jbloggs.hcpapps.net`       | Hostname for the root Domain                                   |
+  | `MGC_AWS_DNS_PUBLIC_ZONE_ID` | `Z01234567US0IQE3YLO00`     | AWS Route 53 Zone ID for specified `MGC_ZONE_ROOT_DOMAIN`      | | 
+  | `MGC_AWS_ACCESS_KEY_ID`      | `AKIA1234567890000000`      | Access Key ID, with access to resources in Route 53            |
+  | `MGC_AWS_SECRET_ACCESS_KEY`  | `Z01234567US0000000`        | Access Secret Access Key, with access to resources in Route 53 |
+  | `MGC_AWS_REGION`             | `eu-west-1`                 | AWS Region                                                     |
+  | `MGC_SUB_DOMAIN`             | `myapp.jbloggs.hcpapps.net` | AWS Region                                                     |
```
eoinfennessy, line 29: I wonder should this section be moved to after the explanation of the use of two terminal windows?

laurafitzgerald:
@maleck13 replies inline to your comments from above.

> Running through this and making some notes
> 
> 1. I think downloading the binaries to a tmp dir (at least some of the less common ones such as yq, istioctl and operator-sdk ) would be good. Running from scratch, it was a bit frustrating the number of deps I had to get in place. Also if we are not going to install a dep we should offer a link to the github repo?

We'll be covered here by the follow on issue to have the dependencies managed by a Dockerfile. Shout if there is anything further for this one.


> 2. I think when asking for AWS credentials in the prompt we should call out what they are for? I guess this will also be in doc.

Addressed in the output. And also we made an update to the ocm-walkthrough document where we have already good information about the use of those env vars and can add the same information to the associated document for the quickstart.

> 3. minor thing but when it ran it said `deleting existing kind clusters` prob should be clear it is only removing exiting kuadrant kind clusters

addressed 👍 

> 4. when install finished we get `Your managed cluster kind-mgc-workload-1 has joined the Hub successfully. Visit https://open-cluster-management.io/scenarios or https://github.com/open-cluster-management-io/OCM/tree/main/solutions for next steps. ` which makes it look like I should go there. But really I think we want to link them to our guides for Gateways

Do you have a preference for which guide the user should go to next? I don't see a gateway specific one currently available but I could be just missing it?

> 5. We should finish with kuadrant installed (IE MGC and data plane) so that the next step is to create a gateway, place it and protect it

addressed 👍 

> 
> Discussing the dependencies with @mikenairn we asked whether a docker image with our install deps in it would make sense. That way there is no need for figuring out which OS etc and your dependencies reduce to docker, kind and kubectl. Other than installation there is no use for istioctl, operator-sdk or ocm really.

+1 issue coming soon
> 
> While this is a batteries included install, it did get me thinking that perhaps we want some options to remove some dependencies. (not part of this but for future consideration) IE I am interested in DNSPolicy only and so there is no need for istio and cert-manager. I think potentially this will become clearer as we look at what a single cluster install should look like other than its current form of being a full OCM install

In the current implementation the install does have one spoke and one hub cluster. is the intention for the quickstart to contain a one cluster setup?

config/quickstarthub/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+resources:
+  - ../ingress-nginx
+  - ../metallb
+  - ../cert-manager
+  - ../external-dns
```
david-martin, line 5:
```suggestion
```

Pretty sure external-dns is not used for anything.
Probably added during a spike for replacing the dns provider logic in mgc with external-dns

hack/quickstart-setup.sh
```diff
@@ -0,0 +1,322 @@
...
+    clusterName=${1}
+    kubectl config use-context kind-${clusterName}
+    echo "Initialize quickstart setup on ${clusterName}"
+    kustomize build ${QUICK_START_HUB_KUSTOMIZATION} --enable-helm --helm-command helm | kubectl apply -f -
+    echo "Waiting for ingress-nginx deployments to be ready"
+    kubectl -n ingress-nginx wait --timeout=600s --for=condition=Available deployments --all
+    echo "Waiting for metallb-system deployments to be ready"
+    kubectl -n metallb-system wait --for=condition=ready pod --selector=app=metallb --timeout=300s
+    echo "Waiting for external-dns deployments to be ready"
+    kubectl -n external-dns wait --timeout=300s --for=condition=Available deployments --all
```
david-martin, line 203:
```suggestion
```

Not needed. See comment above

config/quickstarthub/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+resources:
+  - ../ingress-nginx
```
david-martin, line 2:
```suggestion
```
I'd argue this isn't needed for the quickstart as no Ingresses are created.
Would need to verify that.
If follow on guides need it (like something with ArgoCD or Thanos/Grafana), it can be added.
That being said, a Gateway should be running so a HttpRoute can be used instead of an Ingress then.

config/quickstartspoke/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+resources:
+  - ../istio
+  - ../gateway-api
+  - ../ingress-nginx
```
david-martin, line 4:
```suggestion
```
Not needed. See comment above

hack/quickstart-setup.sh
```diff
@@ -0,0 +1,322 @@
...
+EOF
+}
+
+deployQuickStartHub() {
+    clusterName=${1}
+    kubectl config use-context kind-${clusterName}
+    echo "Initialize quickstart setup on ${clusterName}"
+    kustomize build ${QUICK_START_HUB_KUSTOMIZATION} --enable-helm --helm-command helm | kubectl apply -f -
+    echo "Waiting for ingress-nginx deployments to be ready"
+    kubectl -n ingress-nginx wait --timeout=600s --for=condition=Available deployments --all
```
david-martin, line 199:
```suggestion
```
Not needed.
See comment above

hack/quickstart-setup.sh
```diff
@@ -0,0 +1,322 @@
...
+    kubectl -n cert-manager wait --timeout=300s --for=condition=Available deployments --all
+}
+
+deployQuickStartSpoke() {
+   clusterName=${1}
+    kubectl config use-context kind-${clusterName}
+    echo "Initialize quickstart setup on ${clusterName}"
+    kustomize build ${QUICK_START_SPOKE_KUSTOMIZATION} --enable-helm --helm-command helm | kubectl apply -f -
+    echo "Waiting for ingress-nginx deployments to be ready"
+    kubectl -n ingress-nginx wait --timeout=600s --for=condition=Available deployments --all
```
david-martin, line 214:
```suggestion
```
Not needed. See comment above

docs/how-to/ocm-control-plane-walkthrough.md
```diff
@@ -6,7 +6,10 @@ You will also deploy a simple application that uses that gateway for ingress and
 We will start with a single cluster and move to multiple clusters to illustrate how a single gateway definition can be used across multiple clusters and highlight the automatic TLS integration and also the automatic DNS load balancing between gateway instances.
 
 ## Requirements
-- [Kind](https://kind.sigs.k8s.io/)
+- [kind](https://kind.sigs.k8s.io/)
+- [operator-sdk](https://sdk.operatorframework.io/docs/installation/)
```
david-martin, line 10: Is operator-sdk still needed?

hack/quickstart-setup.sh
```diff
@@ -0,0 +1,322 @@
...
+
+cleanup
+
+port80=9090
+port443=8445
+proxyPort=9200
+metalLBSubnetStart=200
+
+# Create network for the clusters
+docker network create -d bridge --subnet 172.32.0.0/16 mgc --gateway 172.32.0.1 \
```
david-martin, line 286:
```suggestion
docker network create -d bridge --subnet 172.31.0.0/16 mgc --gateway 172.31.0.1 \
```
An issue to fix this fell through the cracks of stale automation https://github.com/Kuadrant/multicluster-gateway-controller/issues/119
Can we fix this here before this new file is added?

david-martin (commented):
Did a relatively quick first pass, with some comments.
It's great to have this as an easier way to try things out
/approve

Will hold off on lgtm pending resolution of comments and trying it out locally again.

config/default/kustomization.yaml
```diff
@@ -12,52 +12,25 @@ namePrefix: mgc-
...
-#    version: v1
-#    name: webhook-service
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../crd
+- ../rbac
+- ../manager
+patches:
+- path: manager_auth_proxy_patch.yaml
```
mikenairn, line 36: meh, this is generated by kubebuilder, would normally just leave generated files as they were

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
-  newTag: latest
+  newName: quay.io/kuadrant/multicluster-gateway-controller
+  newTag: main
```
mikenairn, line 8: Not strictly necessary for the quick start script to work since you have another kustomization that sets it elsewhere. There's no issue changing this, probably makes more sense, but one thing that might happen now is this file being changed after running `make deploy-controller`. You might want to look at making that use the same image and tag names.

config/quickstarthub/kustomization.yaml
mikenairn: This is definitely not required. I have removed it a couple of times, but seems to keep appearing back, probably bad through a rebase.

docs/how-to/ocm-control-plane-walkthrough.md
mikenairn: Thought for another day, but we really don't have to deploy the kuadrant operator via OLM or deploy OLM with the operator-sdk. I guess for now, since that's what we do we will have to leave it here, but it would be better if we didn't have to.

hack/quickstart-setup.sh
```diff
@@ -0,0 +1,322 @@
...
+# Add workload clusters if MGC_WORKLOAD_CLUSTERS_COUNT environment variable is set
+if [[ -n "${MGC_WORKLOAD_CLUSTERS_COUNT}" ]]; then
+  for ((i = 1; i <= ${MGC_WORKLOAD_CLUSTERS_COUNT}; i++)); do
+    kindCreateCluster ${KIND_CLUSTER_WORKLOAD}-${i} $((${port80} + ${i})) $((${port443} + ${i})) $((${i} + 1))
+    deployIstio ${KIND_CLUSTER_WORKLOAD}-${i}
+    deployQuickStartSpoke ${KIND_CLUSTER_WORKLOAD}-${i}
+    deployOLM ${KIND_CLUSTER_WORKLOAD}-${i}
+    deployOCMSpoke ${KIND_CLUSTER_WORKLOAD}-${i}
+  done
+fi
```
mikenairn, line 319: You might want to consider disentangling the creation of the clusters from the provisioning of components on them, this was something i was hoping would be done for local-setup as part of https://github.com/Kuadrant/multicluster-gateway-controller/issues/256, but for some reason has been stopped.

Doing it this way might not be a huge issue for you in a quick start, if it fails i guess you will just have to start again anyway, but it's really frustrating when it fails to provision a cluster after already creating and provisioning others, and the only option is to start completely from scratch.

You could iterate and create the kind clusters first see https://github.com/mikenairn/multi-cluster-traffic-controller/blob/56059eb0fe82fcf99247aa2fffcd68f07aef331d/hack/local-setup.sh#L231-L243

And then iterate again to provision each with mgc components see https://github.com/mikenairn/multi-cluster-traffic-controller/blob/56059eb0fe82fcf99247aa2fffcd68f07aef331d/hack/local-setup.sh#L247-L260

This isn't something i would expect to block a merge, but i think separating those two things would help provide a better flow in the quick start 

1. Create my infra (Kind Clusters). This part could for example eventually be updated to have you select existing clusters (Clusters already running on public clouds)
2. Provision stuff on created/selected clusters

hack/quickstart-setup.sh
mikenairn: I _think_ you are adding the gateway-api to the spoke and not the hub https://github.com/Kuadrant/multicluster-gateway-controller/pull/362/files#diff-2e5b5a1893a8b2682793d1a75d6ad60b3a3ebb132d09baee88ff5f93961eadd9R3?

hack/quickstart-setup.sh
```diff
@@ -0,0 +1,322 @@
...
+    kubectl config use-context kind-${clusterName}
+    kubectl label managedcluster kind-mgc-control-plane ingress-cluster=true
+    cat <<EOF | kubectl apply -f -
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: GatewayClass
+metadata:
+  name: kuadrant-multi-cluster-gateway-instance-per-cluster
+spec:
+  controllerName: "kuadrant.io/mgc-gw-controller"
+EOF
```
mikenairn, line 232: This reminds me, you won't be able to add CRDs and use them in the same `kustomize` command, you either need two separate apply commands, one with the CRDS, and one with the stuff that uses it. Or you can do what i did here and just run the command again https://github.com/mikenairn/multi-cluster-traffic-controller/blob/56059eb0fe82fcf99247aa2fffcd68f07aef331d/hack/local-setup.sh#L175-L177. Might not be the nicest for a quick start though.

hack/quickstart-setup.sh
mikenairn: To remove this you will need to remove the need for the `istioctl operator init` command below. That generates a load of stuff and applies it to the cluster. You can instead just check it in though and add it to the kustomization see https://github.com/mikenairn/multi-cluster-traffic-controller/tree/56059eb0fe82fcf99247aa2fffcd68f07aef331d/config/istio, only thing here you will hit the issue described here https://github.com/Kuadrant/multicluster-gateway-controller/pull/362/files#r1277269321

mikenairn: ```./bin/istioctl operator dump```

config/default/kustomization.yaml
laurafitzgerald: I was getting some warnings from kubebuilder so I ran `kustomize edit fix` which produced these changes. I can remove them though if that's preferred.

config/manager/kustomization.yaml
laurafitzgerald: yes that make command uses `controller:latest` I chose not to change that one because it would keep the make command consistent with it's previous behaviour. let me know though if i should go ahead and change that? I think it would be just removing the edit like you had in the example?

hack/quickstart-setup.sh
laurafitzgerald: yes i hit that with GatewayClass and one other that is escaping me memory now. It's a pity. Would be nice to apply the CRD and cr in one go. I'll go ahead and use the mechanism above because my preference was to get all yaml (excluding that that depends on an env var) into one of the two quickstart kustomizes. 👍

config/quickstarthub/kustomization.yaml
laurafitzgerald: we'll get rid of it for good here so. 🗡️

hack/quickstart-setup.sh
laurafitzgerald: @mikenairn i've incorporated some of the changes suggested here and pushed them in the latest pitstop commit. I agree it's a lot cleaner and was able to remove more of the scripting. There's probably some further cleanup that could be done. I refactored the control/workload kustomizes to match what you had. It looks cleaner. 👍

docs/how-to/ocm-control-plane-walkthrough.md
laurafitzgerald: I added a note under the terminal sessions section which hopefully addresses this which states that MGC_SUB_DOMAIN is required in both terminals.

hack/quickstart-setup.sh
laurafitzgerald: @mikenairn i can't get to the comment that you're linking to above here. I checked in the changes from your branch to have kustomize install istio. The script is completing successfully. Is the issue you mentioned above still present? If so could you describe it more?

docs/how-to/ocm-control-plane-walkthrough.md
laurafitzgerald: While it would be nice to address this here as we did with istio it seems a bigger change so perhaps we could create a follow on issue for it? We plan to create a follow on issue for having a Dockerfile manage the dependencies so this might not be so bit a problem after that?

config/default/kustomization.yaml
mikenairn: Yeah, it's fine, i was just saying why it was the way it was. It's really just generated as a start anyway. I'd say these would be better as the first lines of the file though. 
```.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
```

docs/how-to/ocm-control-plane-walkthrough.md
mikenairn: Yes, I'd leave it for now. If we can get a docker image working it doesn't really matter.

docs/how-to/ocm-control-plane-walkthrough.md
```diff
@@ -6,7 +6,9 @@ You will also deploy a simple application that uses that gateway for ingress and
...
+- [yq](https://mikefarah.gitbook.io/yq/v/v3.x/)
```
mikenairn, line 11: The script doesn't appear to install dependencies right? Does that mean you need to list all of them here? ocm clusteradm etc...

config/default/kustomization.yaml
laurafitzgerald: ah, I was reading the comment as being about the patch. I didn't notice those lines added to the file. Will move those to the beginning. make sense. 👍

docs/how-to/ocm-control-plane-walkthrough.md
laurafitzgerald: There is a list of what we are doing with the script added. if there is a suitable diagram somewhere, I'm happy to add that too.

docs/how-to/ocm-control-plane-walkthrough.md
laurafitzgerald: In the current state it doesn't install dependencies no. The docs edit here is also just a suggestion. If we'd prefer to stick with make local-setup for this one it can be reverted. If we're happy to go with the change I'll get clusteradm added in as well. Missed that one. Good spot.

hack/quickstart-setup.sh
laurafitzgerald: @mikenairn @maleck13 wdyt?

maleck13: yes we can

config/manager/kustomization.yaml
mikenairn: We can come back to it if it becomes an issue.

laurafitzgerald:
> @laurafitzgerald other than getting the e2e tests passing, is there anything else needed on this PR now?

@maleck13 there's one or two questions in comments above.

laurafitzgerald:
@maleck13 all questions should be addressed now. I pushed a final update to add the quickstart document itself. so final thing to get this merged is that folks are happy with the docs changes that i've made. it does change the context of the walkthrough itself so wanting to be sure it makes sense to everyone. Some discussion also about it here https://kubernetes.slack.com/archives/C05J0D0V525/p1690878462120169
cc @mikenairn @eoinfennessy

mikenairn:
Changes look good to me. I haven't tried it, but will give it a run through later after it's merged.

/lgtm
/approve

PR #367, opened 2023-07-24
Upgrade cert-manager

roivaz:
* Upgrades cert-manager to latest release
* Removes customizations that were required for kcp

### DIFF
--- a/config/cert-manager/kustomization.yaml
+++ b/config/cert-manager/kustomization.yaml
@@ -14,11 +14,5 @@ helmCharts:
         leaderElection:
           namespace: cert-manager
       installCRDs: true
-      # The Job API is not present
-      startupapicheck:
-        enabled: false
-      # The APIService API is not present
-      cainjector:
-        enabled: false
       # Customize active controllers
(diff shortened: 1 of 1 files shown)


config/cert-manager/kustomization.yaml
```diff
@@ -6,19 +6,13 @@ resources:
...
+    version: v1.12.2
     releaseName: mgc
     namespace: cert-manager
     valuesInline:
       global:
         leaderElection:
           namespace: cert-manager
       installCRDs: true
-      # The Job API is not present
-      startupapicheck:
```
david-martin, line 18: Not having this enabled leaves out a valid check that may not manifest until later?

config/cert-manager/kustomization.yaml
```diff
@@ -6,19 +6,13 @@ resources:
...
-        enabled: false
-      # The APIService API is not present
-      cainjector:
```
david-martin, line 21: I don't believe we're using the ca injector stuff, or intend to,
but no harm in leaving the installation as the default.

david-martin (commented):
No immediate concerns with this.
1 question

Issue #335, opened 2023-07-13
Remove DNS Records as part of make local-cleanup command


pmccarthy:
**What**
At the moment running the make local-cleanup command immediately deletes local kind clusters. If dnsrecords have been created on these clusters the associated records in Route53 will be left in place which can lead to conflicts if attempting to repeat a walkthrough with the same dns configurations.

**How**
Before deleting the kind clusters locally, we should check to see if any dnsrecords exist on these clusters first and if so delete them so that the corresponding records in Route53 will be cleaned up also. Once removed the kind clusters can then be deleted.

A caveat to this is that the controller needs to be running either locally or on cluster in order for the records to be cleaned up but this is still an improvement on what we currently have.

philbrookes:
This issue is stale because it has been open for 30 days with no activity.

philbrookes:
This issue was closed because it has been inactive for 30 days since being marked as stale.

PR #371, opened 2023-07-24
GH-256 separation of local setup logic

makslion:
## What 
To speed up local dev work this PR separates provisioning of the kind clusters from deployment onto those clusters 
## How
This PR contains 3 commits with different purposes each: 
1. Separation of the local setup logic into two make targets 
2. Introduction of the attempt to clean up resources from kind clusters rather than removing kind clusters themselves. Allow for much faster redeployment if needed 
3. Changes in docs to reflect those changes. 
I've left old command in place as they are still could be used and they should be working just like before. Plus I've left the old `make local setup` in the readme to showcase that it is still can be done. The make command, however, will tip that you could use other targets. 
And a style point for prompting user to run controller to make sure MGC cleans up DNS record in case of the `make local-cleanup-mgc`. Relevant to the #335 
## Verification 
Will take a while :laughing:  
1. The "old" scenario:
 `make local setup` followed by `make local-cleanup`  after success 

2. I've messed up the MGC scenario 
You could follow [this](https://github.com/makslion/multicluster-gateway-controller/blob/dc1bacd088bdb7c5adab6e908bc2faa82a62a4ab/docs/how-to/ocm-control-plane-walkthrough.md) how-to doc up to the point of placing gateways. 
Run the `make local-cleanup-mgc MGC_WORKLOAD_CLUSTERS_COUNT=1` 
followed by the `make local-setup-mgc MGC_WORKLOAD_CLUSTERS_COUNT=1` and ensure that gateway can be deployed again to the data plane cluster


makslion:
hm, seems like I was to slow on that front. There are similar changes in #362

makslion:
/hold 
for now. This work might be delayed by the `quick-start` script. I'll unhold PR after the team discussion.

hack/.deployUtils
```diff
@@ -322,86 +311,82 @@ deployPrometheusForFederation() {
...
+    clusterName=${1}
+    kubectl config use-context kind-${clusterName}
+    echo "Initialize quickstart setup on ${clusterName}"
+    wait_for "kustomize --load-restrictor LoadRestrictionsNone build ${QUICK_START_HUB_KUSTOMIZATION} --enable-helm --helm-command helm | kubectl apply -f -" "${QUICK_START_HUB_KUSTOMIZATION} cluster config apply" "1m" "5"
+    echo "Waiting for metallb-system deployments to be ready"
+    kubectl -n metallb-system wait --for=condition=ready pod --selector=app=metallb --timeout=300s
+    echo "Waiting for cert-manager deployments to be ready"
+    kubectl -n cert-manager wait --timeout=300s --for=condition=Available deployments --all
+    echo "Waiting for istio deployments to be ready"
+    kubectl -n cert-manager wait --timeout=300s --for=condition=Available deployments --all
```
laurafitzgerald, line 375:
```suggestion
    kubectl -n istio wait --timeout=300s --for=condition=Available deployments --all
```

docs/how-to/ocm-control-plane-walkthrough.md
```diff
@@ -381,6 +381,22 @@ So now we have a working gateway with DNS and TLS configured. Let place this gat
 ## Watching DNS changes
 If you want you can use ```watch dig $MGC_SUB_DOMAIN``` to see the DNS switching between the two addresses 
 
+## Clean up local environment
+> **Note:** :thought_balloon: The `quickstart-setup.sh` cleans up automatically before the run
+
+In any terminal window target control plane cluster by:
```
laurafitzgerald, line 387: I'm wondering if we need to add this here. Reason being that with the recent quickstart change this wallkthrough doesn't required a clone of the repo. But introducing some make commands does.

hack/.deployUtils
```diff
@@ -1,67 +1,10 @@
...
-  metalLBSubnet=${2}
+# Shared functions between local-setup-mgc and quickstart-setup script
 
-  kubectl config use-context kind-${clusterName}
-  echo "Deploying MetalLB to ${clusterName}"
-  ${KUSTOMIZE_BIN} build ${METALLB_KUSTOMIZATION_DIR} | kubectl apply -f -
-  echo "Waiting for deployments to be ready ..."
-  kubectl -n metallb-system wait --for=condition=ready pod --selector=app=metallb --timeout=300s
+configureMetalLB () {
+  metalLBSubnet=${1}
```
laurafitzgerald, line 7:
```suggestion
  metalLBSubnet=${2}
```

laurafitzgerald, line 7:
```suggestion
  clusterName=${1}
  metalLBSubnet=${2}

  kubectl config use-context kind-${clusterName}
  echo "Deploying MetalLB to ${clusterName}"}
```

docs/how-to/ocm-control-plane-walkthrough.md
makslion: fair point. Removed

hack/.deployUtils
makslion: is there a use for the cluster name? I'd just remove it - one less var to swing around

hack/.deployUtils
```diff
@@ -322,86 +311,82 @@ deployPrometheusForFederation() {
...
-# Deploy Redis
-deployRedis ${KIND_CLUSTER_CONTROL_PLANE}
+configureController() {
+    clusterName=${1}
+    kubectl config use-context kind-${clusterName}
+    echo "Initialize local dev setup for the controller on ${clusterName}"
 
-# Deploy MetalLb
-deployMetalLB ${KIND_CLUSTER_CONTROL_PLANE} ${metalLBSubnetStart}
+    cat <<EOF | kubectl apply -f -
```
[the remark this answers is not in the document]
makslion: This one is still for the quickstart script only. I've moved it all anyway: this script will be sourced anyway because of the shared functions and moving everything here makes it "one place to hold them all". Don't have a strong opinion on where to hold quickstart and local-setup specific things. And not sure I could avoid having this bit at all

hack/.deployUtils
makslion: why isio? Or the better question is why it is a `cert-manager` [here](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/hack/quickstart-setup.sh#L171-L181) ?

hack/.deployUtils
laurafitzgerald: it's used to set the context inside the function. if we don't use it the change should still include metalLBSubent=${2} otherwise the wrong value is used in the function call from the quickstart.

hack/.deployUtils
laurafitzgerald: yeah, it's an error in the previous pr. It should be istio-operator. I can fix it seperately if you prefer.

hack/.deployUtils
```diff
@@ -47,7 +47,7 @@ deployOCMHub(){
     clusterName=kind-${KIND_CLUSTER_CONTROL_PLANE}
     echo "Found single cluster installing hub and spoke on the one cluster (${clusterName})"
     join=$(${CLUSTERADM_BIN} get token --context ${clusterName} |  grep -o  'clusteradm.*--cluster-name')
-    ${join} ${clusterName} --bundle-version='0.11.0' --feature-gates=RawFeedbackJsonString=true --force-internal-endpoint-lookup --context ${clusterName} | grep clusteradm
```
laurafitzgerald: This change may not work for the quickstart as everything should run without the need for binaries installed to a directory.

hack/.deployUtils
```diff
@@ -79,7 +79,7 @@ deployOCMSpoke(){
   kubectl config use-context kind-${KIND_CLUSTER_CONTROL_PLANE}
   join=$(${CLUSTERADM_BIN} get token --context kind-${KIND_CLUSTER_CONTROL_PLANE} |  grep -o  'clusteradm.*--cluster-name')
   kubectl config use-context kind-${clusterName}
-  ${join} kind-${clusterName} --bundle-version='0.11.0' --feature-gates=RawFeedbackJsonString=true --force-internal-endpoint-lookup --context kind-${clusterName} | grep clusteradm
```
laurafitzgerald: same here.

hack/.deployUtils
```diff
@@ -47,7 +47,7 @@ deployOCMHub(){
     clusterName=kind-${KIND_CLUSTER_CONTROL_PLANE}
     echo "Found single cluster installing hub and spoke on the one cluster (${clusterName})"
     join=$(${CLUSTERADM_BIN} get token --context ${clusterName} |  grep -o  'clusteradm.*--cluster-name')
-    ${join} ${clusterName} --bundle-version='0.11.0' --feature-gates=RawFeedbackJsonString=true --force-internal-endpoint-lookup --context ${clusterName} | grep clusteradm
+    ${BIN_DIR}/${join} ${clusterName} --bundle-version='0.11.0' --feature-gates=RawFeedbackJsonString=true --force-internal-endpoint-lookup --context ${clusterName} | grep clusteradm
```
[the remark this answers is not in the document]
makslion: it should work. https://github.com/Kuadrant/multicluster-gateway-controller/pull/371/files#diff-dd9adc823d5cb403f5710f23e6624096dd323f422723e7672b0f76c51385d849R21 looks ugly to me, but can't think of a more elegant solution

hack/.deployUtils
makslion: I'm a bit confused. There is a [deployMetaLB](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/hack/local-setup.sh#L56-L82) that is used only for the local setup. And there is a [configureMetaLB](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/hack/quickstart-setup.sh#L39-L58) that is used in the [quickstart](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/hack/quickstart-setup.sh#L273) and is called from the [deployMetaLB](https://github.com/Kuadrant/multicluster-gateway-controller/blob/fa90bdef31b8234dfc5b27fe366cf3d98b54b84e/hack/.deployUtils#L114) function. The configure one is not using context as in case it is called from the deployMetaLB the contexts is already set [here](https://github.com/Kuadrant/multicluster-gateway-controller/blob/fa90bdef31b8234dfc5b27fe366cf3d98b54b84e/hack/.deployUtils#L109) and if it is called from the quickstart script the function before it [sets](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/hack/quickstart-setup.sh#L126) the context. I see how it can break, so would you suggest to actually use the passed inside clusterName to ensure the contexts was set? :thinking:

hack/.deployUtils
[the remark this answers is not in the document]
laurafitzgerald: What is the BIN_DIR value?

hack/.deployUtils
laurafitzgerald: we moved it to configureMetalLB because we moved the deployment of metalLB to kustomize for the quickstart, you could move the local setup to install by kustomize as well and that would bring the two approaches to the same. even if we don't do that, both calls to the function should call in the same way. e.g. remove the passing of the context value from the quickstart if it's not needed. although i think safer that it is. so pass the context from the call inside deployMetalLB

hack/.deployUtils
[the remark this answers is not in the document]
makslion: in case of the local setup it will be `./../bin` and the call will look like `<path to the hack folder>/../bin/clusteradm` and in case of the quicstartscript it will be an empty srting just calling `clusteradm`

hack/.deployUtils
makslion: Yeah, better to be extra safe. Passed around `clusterName` for extra safety

hack/.deployUtils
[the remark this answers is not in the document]
laurafitzgerald: sorry it's not clear to me why we need to add this because that's already part of the declaration of join as CLUSTERADM_BIN. 🤔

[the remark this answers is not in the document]
makslion: if we don't have it `local-setup` stops working (the script will be fine though). It is a bit funky and it evaluates join and  on the line 50 it starts as `clusteradm` even when we are having the `CLUSTERADM_BIN` pointing to the binary 
Here. This code 
```
test(){
  clusterName=${1}

  echo ${BIN_DIR}
  echo ${CLUSTERADM_BIN}
  join=$(${CLUSTERADM_BIN} get token --context kind-${clusterName} |  grep -o  'clusteradm.*--cluster-name')
  echo ${join}
  echo ${BIN_DIR}/${join}
}
```
prompts 
```
/home/hudeb/repos/multicluster-gateway-controller/hack/../bin
/home/hudeb/repos/multicluster-gateway-controller/hack/../bin/clusteradm
clusteradm join --hub-token ... --hub-apiserver https://127.0.0.1:39139 --cluster-name
/home/hudeb/repos/multicluster-gateway-controller/hack/../bin/clusteradm  join --hub-token ... --hub-apiserver https://127.0.0.1:39139 --cluster-name
```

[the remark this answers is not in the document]
laurafitzgerald: Strange. okay, so we need to set BIN_DIR to empty string in the quickstart script then yes?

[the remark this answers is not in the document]
makslion: actually, it would keep `/` in front of it for the quickstart. But now it won't

laurafitzgerald:
I've tried the lastest copy and verifying everything works as before with the quickstart script. 👍

hack/local-setup-mgc.sh
```diff
@@ -0,0 +1,109 @@
...
+METALLB_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/metallb
+CERT_MANAGER_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/cert-manager
+EXTERNAL_DNS_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/external-dns
+ARGOCD_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/argocd
+ISTIO_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/istio/istio-operator.yaml
+GATEWAY_API_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/gateway-api
+REDIS_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/kuadrant/redis
+THANOS_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/thanos
+PROMETHEUS_FOR_FEDERATION_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/prometheus-for-federation
+GRAFANA_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/grafana
```
mikenairn, line 37: Since you are using local env vars now, are these even used?

hack/local-setup-mgc.sh
```diff
@@ -0,0 +1,109 @@
...
+
+set -e pipefail
+
+cleanupMGC
```
mikenairn, line 41: It would be good if we can get to a point where this script was idempotent and this cleanup would not be required . What makes this required here currently? OLM?

hack/quickstart-setup.sh
```diff
@@ -241,31 +77,13 @@ if [[ -z "${MGC_WORKLOAD_CLUSTERS_COUNT}" ]]; then
...
-  for ((i = 1; i <= ${MGC_WORKLOAD_CLUSTERS_COUNT}; i++)); do
-    kindCreateCluster ${KIND_CLUSTER_WORKLOAD}-${i} $((${port80} + ${i})) $((${port443} + ${i})) $((${i} + 1))
-  done
-fi
+kindSetupMGCClusters ${KIND_CLUSTER_CONTROL_PLANE} ${KIND_CLUSTER_WORKLOAD} ${port80} ${port443} ${MGC_WORKLOAD_CLUSTERS_COUNT}
 
 # Apply Cluster Configurations to Control cluster
 # Deploy OCM hub
-deployOCMHub ${KIND_CLUSTER_CONTROL_PLANE}
+deployOCMHub ${KIND_CLUSTER_CONTROL_PLANE} "minimal"
```
mikenairn, line 86: What is the purpose of the "minimal" option? What are redis and istio being installed for in the hub currently that they aren't needed here?

hack/local-setup-mgc.sh
```diff
@@ -0,0 +1,109 @@
...
+    joinSubmarinerBroker ${KIND_CLUSTER_WORKLOAD}-${i}
+    deployIstio ${KIND_CLUSTER_WORKLOAD}-${i}
+    installGatewayAPI ${KIND_CLUSTER_WORKLOAD}-${i}
+    deployIngressController ${KIND_CLUSTER_WORKLOAD}-${i}
+    deployMetalLB ${KIND_CLUSTER_WORKLOAD}-${i} $((${metalLBSubnetStart} + ${i}))
+    deployOLM ${KIND_CLUSTER_WORKLOAD}-${i}
+    deployDashboard ${KIND_CLUSTER_WORKLOAD}-${i} ${i}
+    argocdAddCluster ${KIND_CLUSTER_CONTROL_PLANE} ${KIND_CLUSTER_WORKLOAD}-${i}
+    deployAgentSecret ${KIND_CLUSTER_WORKLOAD}-${i} "true"
+    deployAgentSecret ${KIND_CLUSTER_WORKLOAD}-${i} "false"
```
mikenairn, line 97: Might be a task for after, but i don't believe these are required any longer. They were used by the old agent that no longer exists.

mikenairn (commented):
This looks good generally. It is a little bit difficult to navigate, but that is probably partly to do with my IDE not recognising sourced files. I don't really see any other way of sharing these things across different scripts though.

There are quite a few things in these scripts that should really be optional i.e. argocd, dashboards, submariner etc.. It would be nice to be able to specify a minimum required set of dependencies that need installed on each cluster for mgc to work.

We should also look into aligning the local-setup and quick-start script so they use the same method of building the run list for components that need installed.

These would all be follow on tasks to this though.

hack/local-setup-mgc.sh
makslion: The cleanup also gives an option to "uninstall" our stuff. Like deleting Gateway and asking the user to run the reconciler so the DNS records will be cleaned up and nuking namespaces. The idea being that when I'm playing with stuff locally the `make local-setup-mgc` would be "bring me back to the point when everything was working". 
In the context of this script only OLM really needs to either be cleaned up or the install logic should be adjusted. What would be the benefits of not needing a cleanup before the installation?

hack/.deployUtils
```diff
@@ -94,10 +141,11 @@ deployCertManager() {
   kubectl delete validatingWebhookConfiguration mgc-cert-manager-webhook
   kubectl delete mutatingWebhookConfiguration mgc-cert-manager-webhook
   # Apply the default glbc-ca issuer
-  kubectl create -n cert-manager -f ./config/default/issuer.yaml
+  kubectl apply -n cert-manager -f ./config/default/issuer.yaml
 }
 
 deployExternalDNS() {
+  local EXTERNAL_DNS_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/external-dns
```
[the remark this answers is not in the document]
makslion: motivation for `local` is that they aren't used anywhere else - with so much things happening here and all the vars floating around it seemed like a good idea to highlight that these are just for a function they reside in. The ones that aren't local are used in more than one code segment.

hack/.deployUtils
```diff
@@ -81,7 +27,108 @@ metadata:
...
+    deployOLM ${KIND_CLUSTER_CONTROL_PLANE}
+
+    if ! [[ -n "${minimal}" ]]; then
+        deployIstio ${KIND_CLUSTER_CONTROL_PLANE}
+    fi
+  fi
+  if ! [[ -n "${minimal}" ]]; then
+      echo "Installing Redis in kind-mgc-control-plane"
+      ${KUSTOMIZE_BIN} build ${REDIS_KUSTOMIZATION_DIR} | kubectl apply -f -
+  fi
```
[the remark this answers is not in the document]
mikenairn: Since this was existing behaviour we do not need to look into this as part of this PR, but we should make sure we have follow up issues to look into why this is here.

hack/.deployUtils
```diff
@@ -81,7 +27,108 @@ metadata:
...
+      sleep 10
+      [[ counter -eq $max_retry ]] && echo "Failed!" && exit 1
+      echo "Trying again. Try #$counter"
+      ((++counter))
+    done
+    deployOLM ${KIND_CLUSTER_CONTROL_PLANE}
+
+    if ! [[ -n "${minimal}" ]]; then
+        deployIstio ${KIND_CLUSTER_CONTROL_PLANE}
+    fi
```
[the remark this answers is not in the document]
mikenairn: Since this was existing behaviour we do not need to look into this as part of this PR, but we should make sure we have follow up issues to look into why this is here.

hack/.deployUtils
```diff
@@ -81,7 +27,108 @@ metadata:
...
+    max_retry=18
+    counter=0
+    until ${CLUSTERADM_BIN} accept --clusters ${clusterName}
+    do
+      sleep 10
+      [[ counter -eq $max_retry ]] && echo "Failed!" && exit 1
+      echo "Trying again. Try #$counter"
+      ((++counter))
+    done
+    deployOLM ${KIND_CLUSTER_CONTROL_PLANE}
```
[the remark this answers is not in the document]
mikenairn: Since this was existing behaviour we do not need to look into this as part of this PR, but we should make sure we have follow up issues to look into this.

hack/local-setup-mgc.sh
mikenairn: Since this was existing behaviour we do not need to look into this as part of this PR, but we should make sure we have follow up issues to look into removing this.

hack/.deployUtils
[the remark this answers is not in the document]
mikenairn: I think this is fine :+1:

hack/local-setup-mgc.sh
mikenairn: If someone really wanted to cleanup first they could just call the cleanup script, or make target, before calling this script.

Reason to make it cleanup would be to bring it inline with how any configuration management tooling would work i.e. ArgoCD, by just reapplying config to a known set of clusters. If for instance it fails half way through for some networking issues (Happens to me often), then i don't really need it to clean everything up before it applies stuff again. I just need to keep trying until it completes and brings my clusters into the expected state.

This is not something you need to change in this PR, just something to consider when looking at follow on tasks for enhancements to these scripts.

hack/.deployUtils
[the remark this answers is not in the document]
makslion: #413

hack/.deployUtils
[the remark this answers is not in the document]
makslion: #413

hack/local-setup-mgc.sh
makslion: created #413 to revisit this part

hack/quickstart-setup.sh
makslion: created #413 to revisit this part

hack/local-setup-mgc.sh
makslion: created #413 to revisit this part

makslion:
>There are quite a few things in these scripts that should really be optional i.e. argocd, dashboards, submariner etc.. It would be nice to be able to specify a minimum required set of dependencies that need installed on each cluster for mgc to work.

The #414 will address the "we are installing it all" situation for the local setup

makslion:
tide/merge-method-squash

PR #373, opened 2023-07-25
Add DNSPolicy docs

### LINKED ISSUE #372: Add new dnspolicy doc to MGC repo

mikenairn:
/closes #372


docs/dns-policy.md
```diff
@@ -0,0 +1,497 @@
...
+metadata:
+  name: apps.hcpapps.net
+  namespace: multi-cluster-gateways
+spec:
+  domainName: apps.hcpapps.net
+  description: "apps.hcpapps.net managed domain"
+  dnsProviderSecretRef:
+    name: my-aws-credentials
+    namespace: multi-cluster-gateways
+    type: AWS
```
david-martin, line 69: A comment on the ManagedZone recourse here, not the example.
Is the `type` field the type of the Secret or the type of the dns provider?
Maybe it's both and doesn't matter?

docs/dns-policy.md
```diff
@@ -0,0 +1,497 @@
...
+![aws-recordset-list-1](./images/dns-policy/aws-recordset-list-1.png)
+
+The listener hostname is now be resolvable through dns:
+
+```bash
+dig echo.apps.hcpapps.net +short
+lb-2903yb.echo.apps.hcpapps.net.
+default.lb-2903yb.echo.apps.hcpapps.net.
+lrnse3.lb-2903yb.echo.apps.hcpapps.net.
+172.32.201.1
```
david-martin, line 230: There are a lot of records created.
Is this a good place to explain why that's the case e.g. to future proof so we can do geo later if you want
It can be a little overwhelming trying to make sense of them all.

david-martin (commented):
Few comments, nothing major.
Great to have this detailed doc.

docs/dns-policy.md
mikenairn: yeah, I'll have a go at trying to explain the structure and the reason for it somewhere in here.

docs/dns-policy.md
mikenairn: Good question, i don't believe this is actually required, and probably should just be removed from the spec. /cc @R-Lawton

docs/dns-policy.md
```diff
@@ -0,0 +1,517 @@
+# DNS Policy
+
+The DNSPolicy as a [GatewayAPI](https://gateway-api.sigs.k8s.io/) policy that uses `Direct Policy Attachment` as defined in the [policy attachment mechanism](https://gateway-api.sigs.k8s.io/v1alpha2/references/policy-attachment/) standard.
```
maleck13, line 3:
```suggestion
The DNSPolicy is a [GatewayAPI](https://gateway-api.sigs.k8s.io/) policy that uses `Direct Policy Attachment` as defined in the [policy attachment mechanism](https://gateway-api.sigs.k8s.io/v1alpha2/references/policy-attachment/) standard.
```

docs/dns-policy.md
```diff
@@ -0,0 +1,517 @@
...
+apiVersion: v1
+data:
+  AWS_ACCESS_KEY_ID: <AWS_ACCESS_KEY_ID>
+  AWS_REGION: <AWS_REGION>
+  AWS_SECRET_ACCESS_KEY: <AWS_SECRET_ACCESS_KEY>
+kind: Secret
+metadata:
+  name: my-aws-credentials
+  namespace: multi-cluster-gateways
+type: kuadrant.io/aw
```
maleck13, line 53:
```suggestion
type: kuadrant.io/aws
```

docs/dns-policy.md
maleck13: Yeah not sure this is needed. But it is a separate issue. Initially we had a different DNSProvider API that specified the provider but we have since moved to a secret with the type in it, so I don't think we need the type in the MZ. In additon we may shortly drop the MZ altogether see https://github.com/orgs/Kuadrant/discussions/7

docs/dns-policy.md
maleck13: maybe we link to the DNS structure doc?

maleck13 (commented):
minor things looking pretty good

docs/dns-policy.md
mikenairn: Yeah , I'm going to link to the proposal doc, it's got everything in it. We might want to consider going back and moving that to an actual doc rather than a proposal, but will link to that for now.

docs/dns-policy.md
R-Lawton: Its the type of dnsprovider. It was added to make it easier for users to see what provider is being used without having to go digging but it can be removed

maleck13 (approved):

mikenairn:
@maleck13 @david-martin This look OK now? Just needs an lgtm whenever you are happy with it.

PR #377, opened 2023-07-27
DNSPolicy geo and weight validation updates

mikenairn:
Removes the validation on the geo and weight fields in the DNSPolicy spec in order to more easily support providers with more limited APIs. The values for these fields are now provider specific, and as such any invalid values will result in a failure in the API request resulting in a DNSRecord reconcile error. The Route53 API requires the setting of different fields in the API request depending on whether it's an ISO3166 Alpha-2 country or continent code, as such the validation of the country code is now done in the AWS provider itself, and anything that isn't a valid Alpha-2 code is assumed to be a continent.

Make the loadBalancing field in the DNSPolicy optional. If it's not set, the default weight of 120 will be applied to all weighted records. (Result of feedback after demo).

### DIFF
--- a/pkg/apis/v1alpha1/dnspolicy_types.go
+++ b/pkg/apis/v1alpha1/dnspolicy_types.go
@@ -34,20 +34,18 @@ type DNSPolicySpec struct {
 	// +optional
 	HealthCheck *HealthCheckSpec `json:"healthCheck,omitempty"`
 
-	// +kubebuilder:validation:Required
-	// +required
+	// +optional
 	LoadBalancing *LoadBalancingSpec `json:"loadBalancing"`
 }
 
 type LoadBalancingSpec struct {
-	// +required
+	// +optional
 	Weighted *LoadBalancingWeighted `json:"weighted,omitempty"`
 	// +optional
 	Geo *LoadBalancingGeo `json:"geo,omitempty"`
 }
 
 // +kubebuilder:validation:Minimum=0
-// +kubebuilder:validation:Maximum=255
 type Weight int
 
 type CustomWeight struct {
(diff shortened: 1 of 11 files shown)


pkg/apis/v1alpha1/dnspolicy_types.go
```diff
@@ -34,20 +34,18 @@ type DNSPolicySpec struct {
...
-	// +kubebuilder:validation:Required
-	// +required
+	// +optional
 	LoadBalancing *LoadBalancingSpec `json:"loadBalancing"`
 }
 
 type LoadBalancingSpec struct {
-	// +required
+	// +optional
 	Weighted *LoadBalancingWeighted `json:"weighted,omitempty"`
```
mikenairn, line 43: This is no longer mandatory, and the only thing required in the policy spec is the targetRef, minimum policy required now looks like:

```.yaml
kind: DNSPolicy
metadata:
  name: prod-web
spec:
  targetRef:
    name: prod-web
    kind: Gateway
```

This came about as feedback during the demo, there really is no reason it needs to be required. If you want to change the default value (120), you can still change it by updating the spec but it's really meaningless unless you are applying custom weights.

pkg/apis/v1alpha1/dnspolicy_types.go
```diff
@@ -34,20 +34,18 @@ type DNSPolicySpec struct {
...
 	// +optional
 	Geo *LoadBalancingGeo `json:"geo,omitempty"`
 }
 
 // +kubebuilder:validation:Minimum=0
-// +kubebuilder:validation:Maximum=255
 type Weight int
```
mikenairn, line 49: Removed validation on upper value for weights. Since we will now accept anything for geo, seemed to make sense that we don't limit what you can set for weight either. Providers have different values they will accept:

- Route53: 0 - 255
- Google: 0 - 10000
- Azure:  1- 1000

Like Geo, adding an invalid value for a provider will result in a reconcile error. Error returned from the provider is in the message of the condition.

mikenairn:
Since we now accept provider specific values for geo, be aware that the generated geo lb CNAMES will use that in the grouping name. In the case of google this might lead to quite long dns names like `northamerica-northeast1.lb-4ej5le.myapp.mn.hcpapps.net`. I don't really think this is an issue, just pointing it out, probably makes it clearer to people that use those platforms what it is anyway.

maleck13 (commented):
changes look good any changes need to the docs?

mikenairn:
> changes look good any changes need to the docs?

Docs were updated here https://github.com/Kuadrant/multicluster-gateway-controller/pull/373/commits/41aa41e0ec9a711ace25180e619393fbd031d635

PR #379, opened 2023-07-27
DNSPolicy improve default behaviour

### LINKED ISSUE #359: DNSPolicy improved default behaviour
**What**

We no longer want to create a default DNSPolicy. 

Desired behaviour

- we don't create a DNSPolicy be default
- If there is no DNSPolicy we don't do anything with DNS
- If  a DNSPolicy is created DNSPolicy controller controls the creation and deletion of DNSrecords based on the target
- If the target is removed, DNSPolicy controller removed DNSrecords, if the policy is deleted the records are also removed
- remove the gateway owner ref from the DNSRecords

mikenairn:
/closes #359

Removes the default DNSPolicy creation from the gateway controller.

Update labelling of resources created by the DNSPolicy. Brings it in-line with how the other kuadrant polices work, and makes it easier to find resources created by a policy.

```bash
kubectl get dnsrecords -A --show-labels
NAMESPACE                NAME                   READY   LABELS
multi-cluster-gateways   myapp.mn.hcpapps.net   True    gateway-namespace=multi-cluster-gateways,gateway=prod-web,kuadrant.io/dnspolicy-namespace=multi-cluster-gateways,kuadrant.io/dnspolicy=prod-web
```

### DIFF
--- a/test/integration/dnspolicy_controller_test.go
+++ b/test/integration/dnspolicy_controller_test.go
@@ -15,6 +15,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	gatewayapiv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 
 	"github.com/Kuadrant/multicluster-gateway-controller/pkg/_internal/conditions"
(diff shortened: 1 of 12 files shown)


test/integration/dnspolicy_controller_test.go
```diff
@@ -261,13 +290,13 @@ var _ = Describe("DNSPolicy", Ordered, func() {
 
 			It("should set gateway back reference", func() {
 				existingGateway := &gatewayv1beta1.Gateway{}
-				policyBackRefValue := testNamespace + "/" + TestPlacedGatewayName
-				refs, _ := json.Marshal([]client.ObjectKey{{Name: TestPlacedGatewayName, Namespace: testNamespace}})
+				policyBackRefValue := testNamespace + "/" + dnsPolicy.Name
+				refs, _ := json.Marshal([]client.ObjectKey{{Name: dnsPolicy.Name, Namespace: testNamespace}})
```
mikenairn, line 294: I'm actually not sure how these test were ever passing

test/integration/dnspolicy_controller_test.go
```diff
@@ -171,21 +202,19 @@ var _ = Describe("DNSPolicy", Ordered, func() {
...
 
 			BeforeEach(func() {
-
-				Eventually(func() bool {
-					if err := k8sClient.Get(ctx, client.ObjectKey{testNamespace, TestPlacedGatewayName}, dnsPolicy); err != nil {
-						return false
-					}
-					testModifyPolicyAddHealthCheckAndLoadBalancing(dnsPolicy)
-					if err := k8sClient.Update(ctx, dnsPolicy); err != nil {
+				dnsPolicy = testBuildDNSPolicyWithHealthCheck("test-dns-policy", TestPlacedGatewayName, testNamespace)
```
mikenairn, line 208: @philbrookes I get what you were trying to do here, i had originally done something similar, but for whatever reason running these tests using `envtest` doesn't always appear to cause updates to happen as you would expect leading to flaky tests see https://github.com/Kuadrant/multicluster-gateway-controller/pull/321

maleck13:
tested locally. Works as expected.

PR #384, opened 2023-07-28
DNSRecord AWS API Continuous loop

### LINKED ISSUE #310: DNSRecord reconciler can fall into a continuous loop on AWS API errors
**What**

When an error is returned from the AWS API during a DNSRecord update the reconciler appears to continually try and make the same request forever. 

There also seems to be some issue with the status update of the DNSRecord where the endpoints list is incorrectly updated in certain cases, such as when there is an error returned.

To reproduce create a DNSRecord with an invalid spec and apply it.

- [x]  add status condition to reflect error
- [x]  Configure exponential back off for retry ensuring status reflects success or failure on retry

We should be able to to configure the retry and backoff via the controller runtime
https://github.com/kubernetes-sigs/controller-runtime/blob/16bf3ad036b908d897543c415fcc0bafc5cec711/pkg/controller/controller.go#L117

Ygnas:
Closes #310

### DIFF
--- a/pkg/controllers/dnsrecord/dnsrecord_controller.go
+++ b/pkg/controllers/dnsrecord/dnsrecord_controller.go
@@ -103,7 +103,7 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	if err != nil {
 		status = metav1.ConditionFalse
 		reason = "ProviderError"
-		message = fmt.Sprintf("The DNS provider failed to ensure the record: %v", err)
+		message = fmt.Sprintf("The DNS provider failed to ensure the record: %v", dns.SanitizeError(err))
 	} else {
 		dnsRecord.Status.ObservedGeneration = dnsRecord.Generation
 		dnsRecord.Status.Endpoints = dnsRecord.Spec.Endpoints
@@ -121,7 +121,7 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		}
 	}
 
-	return ctrl.Result{}, nil
+	return ctrl.Result{}, err
 }
 
 // SetupWithManager sets up the controller with the Manager.
--- a/pkg/dns/dns.go
+++ b/pkg/dns/dns.go
@@ -18,6 +18,8 @@ package dns
 
 import (
 	"context"
+	"errors"
+	"regexp"
 
 	"github.com/Kuadrant/multicluster-gateway-controller/pkg/apis/v1alpha1"
 )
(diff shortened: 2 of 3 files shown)


pkg/controllers/dnsrecord/dnsrecord_controller.go
```diff
@@ -103,6 +103,7 @@ func (r *DNSRecordReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	if err != nil {
 		status = metav1.ConditionFalse
 		reason = "ProviderError"
+		err = dns.SanitizeErrormessage(err)
 		message = fmt.Sprintf("The DNS provider failed to ensure the record: %v", err)
```
mikenairn, line 107:
```suggestion
		message = fmt.Sprintf("The DNS provider failed to ensure the record: %v", dns.SanitizeErrormessage(err))
```

pkg/dns/dns.go
```diff
@@ -87,3 +89,9 @@ func (*FakeProvider) ProviderSpecific() ProviderSpecificLabels {
 		HealthCheckID: "fake/health-check-id",
 	}
 }
+
+func SanitizeErrormessage(err error) error {
+	regexp := regexp.MustCompile(`request id: [^\s]+`)
+	sanitizedErr := regexp.ReplaceAllString(err.Error(), "")
+	return errors.New(sanitizedErr)
+}
```
mikenairn, line 97: I'd rename this to `SanitizeErrorMessage` or maybe even `SanitizeError`.

You might also want to consider adding a basic test for it, nothing major, just a couple of examples of what we expect in and the expected output.  I'd also add a [godoc](https://go.dev/blog/godoc) comment to the function explaining what it does, for example:

```
// SanitizeErrormessage removes request specific data from error messages in order to make them consistent across multiple similar requests to the provider.  e.g AWS SDK Request ids `request id: 051c860b-9b30-4c19-be1a-1280c3e9fdc4`
```

mikenairn (changes requested):
Looks good, some small comments.

This will get past the looping, but wonder if we need to make it return an error in the case of an error when publishing the records `ctrl.Result{}, publishErr`. Currently it will try once, maybe twice after the status update, but it should maybe throw an error which would i assume kick in the rate limiting on retries. It could fail to publish for any reason, network issues etc.., so it would be reasonable for it to retry a few times before stopping.

mikenairn (approved):

mikenairn:
Looks good. If the default `ExponentialFailureRateLimiter` looks to be a problem (Re-trying too quickly initially) we can reconfigure it later https://github.com/kubernetes/client-go/blob/20732a1bc198ab57de644af498fa75e73fa44c08/util/workqueue/default_rate_limiters.go#L39-L45, but it will slow down and ultimately stop making requests on repeated errors which is the main thing.

/lgtm
/approve

PR #390, opened 2023-07-31
#314 Metrics for health check probe failures

### LINKED ISSUE #314: Expose a counter metrics from the health check agent
**What**

When health checks fail, we should increment a counter that is exposed via the a prometheus endpoint. This counter should have dimensions relating to the gateway and listener the health check failure is for. 
This will allow users to scrape and alert on this metric. In the future could consider the health check having an option to not remove the record but instead just trigger the metric increment.

sergioifg94:
## Description

> Closes #314 

Create and expose a metric that counts the number of health check failures per probe

## Verification steps

1. Run the local-setup
    ```sh
    make local-setup OCM_SINGLE=true MCTC_WORKLOAD_CLUSTERS_COUNT=1
    ```
3. Run the controller
    ```sh
    ```
5. Create a health check probe that will fail
    ```sh
    kubectl apply -f - <<EOF                                                                                                                                                                                                          
    kind: DNSHealthCheckProbe
    metadata:
      name: example-probe
    spec:
      address: "127.0.0.1"
      host: localhost
      port: 3333
      interval: 30s
    EOF
    ```
7. Check the metric is updated when the health check fails
    ```sh
    curl http://localhost:8080/metrics | grep mgc_dns
    
    mgc_dns_health_check_failures_total{name="example-probe",namespace="default"} 1
    ```

### DIFF
--- a/pkg/controllers/dnshealthcheckprobe/dnshealthcheckprobe_controller.go
+++ b/pkg/controllers/dnshealthcheckprobe/dnshealthcheckprobe_controller.go
@@ -5,12 +5,18 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/go-logr/logr"
+
 	v1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	gatewayapiv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 
+	"github.com/Kuadrant/multicluster-gateway-controller/pkg/_internal/slice"
 	"github.com/Kuadrant/multicluster-gateway-controller/pkg/apis/v1alpha1"
 	"github.com/Kuadrant/multicluster-gateway-controller/pkg/health"
 )
@@ -104,7 +110,11 @@ func (r *DNSHealthCheckProbeReconciler) Reconcile(ctx context.Context, req ctrl.
 			p.AllowInsecureCertificate = probeObj.Spec.AllowInsecureCertificate
 		})
 	} else {
-		notifier := NewStatusUpdateProbeNotifier(r.Client, previous)
+		notifier, err := r.newProbeNotifierFor(ctx, logger, previous)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+
 		r.HealthMonitor.AddProbeQueuer(&health.ProbeQueuer{
 			ID:                       probeId,
 			Interval:                 interval,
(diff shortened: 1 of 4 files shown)


david-martin:
@sergioifg94 looking at the description in https://github.com/Kuadrant/multicluster-gateway-controller/issues/314#issue-1793701416,
in particular this 

> This counter should have dimensions relating to the gateway and listener the health check failure is for

Is it feasible to have those dimensions as labels?


Also, I'm wondering if the failure metric is sufficient on its own, or if a metric with the total number of attempts is needed too for comparison when querying/alerting/visualising?
See https://prometheus.io/docs/practices/instrumentation/#failures

sergioifg94:
@david-martin I think it would be feasable, and it does make sense as we can see the aggregated total of failures per gateway. I'll have a look at making the changes

maleck13:
@sergioifg94 do you want another review of this or are you still updating it?

sergioifg94:
@maleck13 Still updating it, I'll push the changes to the PR soon

sergioifg94:
@david-martin Updated the PR with the suggestions, so the metrics now look like
```
mgc_dns_health_check_attempts_total{gatewayName="prod-web",gatewayNamespace="multi-cluster-gateways",listener="api"} 2
mgc_dns_health_check_failures_total{gatewayName="prod-web",gatewayNamespace="multi-cluster-gateways",listener="api"} 2
```

david-martin:
@sergioifg94 Excellent! This will allow more insightful queries.

Sorry to be a stickler here, but can we make a change to the label names to use snake case, as per the [naming conventions](https://prometheus.io/docs/instrumenting/writing_exporters/#naming).

> Prometheus metrics and label names are written in snake_case. Converting camelCase to snake_case is desirable, though doing so automatically doesn’t always produce nice results for things like myTCPExample or isNaN so sometimes it’s best to leave them as-is.

So `gateway_name` and `gateway_namespace`

sergioifg94:
Thanks @david-martin, good call about the label names, I just pushed the change

david-martin:
Getting metrics as expected, though all probes failing due to local environment being used.

```
curl http://localhost:8080/metrics | grep mgc|grep -v aws
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 68070    0 68070    0     0  7565k      0 --:--:-- --:--:-- --:--:-- 12.9M
# HELP mgc_dns_health_check_attempts_total MGC DNS Health Check Probe total number of attempts
# TYPE mgc_dns_health_check_attempts_total counter
mgc_dns_health_check_attempts_total{gateway_name="prod-web",gateway_namespace="multi-cluster-gateways",listener="api"} 3
# HELP mgc_dns_health_check_failures_total MGC DNS Health Check Probe total number of failures
# TYPE mgc_dns_health_check_failures_total counter
mgc_dns_health_check_failures_total{gateway_name="prod-web",gateway_namespace="multi-cluster-gateways",listener="api"} 3
```

As discussed on slack, I set up a DNSPolicy for a Gateway to ensure the metrics are generated for the associated Gateway.
If a DNSHealthCheckProbe is created manually (i.e. completely separate from a Gateway & HTTPRoute), no metrics will be generated.
This is expected behaviour. Comment from Sergio:

> It is possible to create probes manually like this but the expected behaviour is that theyre created when attaching a dnspolicy to a gateway

/lgtm
/approve

PR #395, opened 2023-08-02
docs and metalb update

laurafitzgerald:
fixing link for curl to run the script. and removing the need to set context because it's set by the script at the end.
Also the metallb ip address wasn't updated.
I'm seeing some issues with the setup however which i'm investigating further
cc @philbrookes


philbrookes (approved):
/lgtm

PR #397, opened 2023-08-02
Health Probe fixes for TLS

philbrookes:
Fix health probes for TLS endpoints and allow ignoring bad certificates.

### DIFF
--- a/pkg/controllers/dnshealthcheckprobe/dnshealthcheckprobe_controller.go
+++ b/pkg/controllers/dnshealthcheckprobe/dnshealthcheckprobe_controller.go
@@ -87,21 +87,23 @@ func (r *DNSHealthCheckProbeReconciler) Reconcile(ctx context.Context, req ctrl.
 			p.Protocol = protocol
 			p.AdditionalHeaders = previous.Spec.AdditionalHeaders
 			p.ExpectedReponses = previous.Spec.ExpectedReponses
+			p.AllowInsecureCertificate = previous.Spec.AllowInsecureCertificate
 		})
 	} else {
 		notifier := NewStatusUpdateProbeNotifier(r.Client, previous)
 		r.HealthMonitor.AddProbeQueuer(&health.ProbeQueuer{
-			ID:                probeId,
-			Interval:          interval,
-			Host:              previous.Spec.Host,
-			Path:              previous.Spec.Path,
-			Port:              previous.Spec.Port,
-			Protocol:          protocol,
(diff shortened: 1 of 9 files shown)


pkg/controllers/dnshealthcheckprobe/dnshealthcheckprobe_controller.go
```diff
@@ -87,6 +87,7 @@ func (r *DNSHealthCheckProbeReconciler) Reconcile(ctx context.Context, req ctrl.
 			p.Protocol = protocol
 			p.AdditionalHeaders = previous.Spec.AdditionalHeaders
 			p.ExpectedReponses = previous.Spec.ExpectedReponses
+			p.AllowInsecureCertificate = previous.Spec.AllowInsecureCertificate
```
sergioifg94, line 90: This is setting the flag only when an existing probe is updated, but not when it's newly created. I'd add this also to the `else` block below, line 103

pkg/health/queuedProbeWorker.go
```diff
@@ -181,3 +187,27 @@ func checkResponse(response int, expected []int) bool {
 	}
 	return false
 }
+
+// TransportWithDNSResponse creates a new transport which overrides hostnames.
+func TransportWithDNSResponse(overrides map[string]string) http.RoundTripper {
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	dialer := &net.Dialer{
+		Timeout:   30 * time.Second,
+		KeepAlive: 30 * time.Second,
```
sergioifg94, line 196: I wonder if it makes sense to have these as part of the spec of the health checks. Not necessarily for this PR but might be interesting

sergioifg94 (commented):
Mostly looks good, just the comments below

pkg/health/queuedProbeWorker.go
philbrookes: yeah, that's an interesting idea. Let's track that in a Jira?

pkg/controllers/dnshealthcheckprobe/dnshealthcheckprobe_controller.go
philbrookes: nice, thanks, added now.

pkg/health/queuedProbeWorker.go
sergioifg94: Sounds good, will do. You mean Jira or a GH issue?

PR #340, opened 2023-07-17
Add Google DNS provider

mikenairn:
closes #258 

ToDo

- [x] Load google credentials from secret
- [x] Implement provider.Ensure (DNSRecord create/update)
  ~~- [x] Add mapping for geo location codes to google regions (NOT needed anymore)~~
- [x] Implement provider.Delete(DNSRecord delete)
- [x] Implement providerEnsureManagedZone
- [x] Implement provider.DeleteManagedZone
- [x] Add documentation on how to work with the google provider
~~- [ ] Update e2e tests to include a google managed zone~~ Moving to separate pr

# Verification 
PREREQS (Youll need GCP access)

to be able to curl use ZONE_ROOT_DOMAIN_GCP=google.hcpapps.net
1. Run through the gcp provider setup in the provider doc https://github.com/Kuadrant/multicluster-gateway-controller/pull/340/files#diff-b2417526f682b86f015e23a3ea7a387ed4c95171f0958d8498c5cdce229307caR78
2. run the following to bring up a cluster with gcp: 
make local-setup OCM_SINGLE=true GCP=true
3.  Run through https://github.com/Kuadrant/multicluster-gateway-controller/blob/943e1753342ff106b4471f4ff7884c2532ac7f43/docs/how-to/ocm-control-plane-walkthrough.md (Note quick start doesn't use gcp)
4. Records should be created in GCP with whatever changes eg geo or weighted
 Screenshot 2023-08-02 at 15 50 42
5. Deleting the DNS record cr will delete the records in gcp

PR #401, opened 2023-08-03
Gh 258 gcp

R-Lawton:


R-Lawton:
Up for review for functionality working on e2e tests

mikenairn:
/hold

The review here still applies https://github.com/Kuadrant/multicluster-gateway-controller/pull/340

pkg/dns/google/google.go
```diff
@@ -0,0 +1,497 @@
...
+			}
+		} else if geoCode {
+			record.RoutingPolicy = &dnsv1.RRSetRoutingPolicy{
+				Geo: &dnsv1.RRSetRoutingPolicyGeoPolicy{},
+			}
+		}
+
+		for _, ep := range endpoints {
+			geoCodeProp, _ := ep.GetProviderSpecificProperty(dns.ProviderSpecificGeoCode)
+			if geoCodeProp.Value != "*" {
```
mikenairn, line 454: @R-Lawton Something is off with this logic. In the case where don't specify a geo-code (weighted only) we will only ever have the single wildcard entry and removing it causes the dns chain of records to break as it's going to be missing any records for the gateway lb.

I had a go and changing it [here](https://github.com/Kuadrant/multicluster-gateway-controller/commit/55ff95795d94099bca4bc46c1a63bc83af829dbd#diff-eea59a6fb9851390f6210a03f57ba88ea1628c32d20d526998600135a1642bed), feels a bit hacky, but might get us to a point of being able to merge this and look into a better solution after.

test/e2e/gateway_single_spoke_gcp_test.go
```diff
@@ -0,0 +1,295 @@
...
+					defer resp.Body.Close()
+					Expect(resp.StatusCode).To(Equal(http.StatusOK))
+				}
+			})
+
+		})
+
+	})
+
+})
```
mikenairn, line 295: This test is out of date with the equivalent one on main. I think however we can just remove it see https://github.com/Kuadrant/multicluster-gateway-controller/commit/55ff95795d94099bca4bc46c1a63bc83af829dbd#r124848829

test/e2e/gateway_single_spoke_test.go
```diff
@@ -293,17 +294,13 @@ var _ = Describe("Gateway single target cluster", func() {
...
 
 					// speed up things by using the authoritative nameserver
-					nameservers, err := net.LookupNS(tconfig.ManagedZone())
+					hname := strings.Replace(tconfig.ManagedZone(), "-", ".", -1)
+
+					nameservers, err := net.LookupNS(hname)
 					Expect(err).ToNot(HaveOccurred())
 					GinkgoWriter.Printf("[debug] authoritative nameserver used for DNS record resolution: %s\n", nameservers[0].Host)
 
-					authoritativeResolver := &net.Resolver{
```
R-Lawton, line 300: @roivaz why did we create our own resolver ? our custom one causes gcp to fail but the default works for both was and gcp

mikenairn: @R-Lawton It will be updated quicker on the authoritative ns, which is why it was being used here. This should work for gcp as well btw, and was locally when i was trying it. What's the issue you were seeing?

.github/workflows/ci-e2e.yaml
```diff
@@ -58,14 +65,29 @@ jobs:
...
           export OCM_SINGLE=1
           export TEST_HUB_NAMESPACE=multi-cluster-gateways
           export TEST_HUB_KUBE_CONTEXT=kind-mgc-control-plane
           export TEST_MANAGED_ZONE=${{ secrets.E2E_AWS_DNS_PUBLIC_ZONE_NAME }}
           make test-e2e
-      - name: Dump MGC logs
+      - name: Dump MGC logs AWS
+        if: ${{ failure() }}
+        run: |
+          kubectl --context kind-mgc-control-plane logs --all-containers --ignore-errors deployment/mgc-controller-manager -n multicluster-gateway-controller-system
```
mikenairn, line 82: You can probably drop the extra `Dump MGC logs x` and just have it once as the last step. It only needs to run once on failure and it will skip all other steps anyway.

test/e2e/gateway_single_spoke_test.go
```diff
@@ -293,7 +294,9 @@ var _ = Describe("Gateway single target cluster", func() {
 					By("ensuring the authoritative nameserver resolves the hostname")
 
 					// speed up things by using the authoritative nameserver
-					nameservers, err := net.LookupNS(tconfig.ManagedZone())
+					hname := strings.Replace(tconfig.ManagedZone(), "-", ".", -1)
+
+					nameservers, err := net.LookupNS(hname)
 					Expect(err).ToNot(HaveOccurred())
```
mikenairn, line 300: @R-Lawton Are any of these changes required?

R-Lawton: no thats my mistake i was getting mixed up with the env vars

pkg/dns/google/google.go
```diff
@@ -0,0 +1,500 @@
...
+				record.RoutingPolicy.Wrr.Items = append(record.RoutingPolicy.Wrr.Items, item)
+			}
+			if geoCode {
+				geoCodeProp, _ := ep.GetProviderSpecificProperty(dns.ProviderSpecificGeoCode)
+				targetIsDefaultGroup := strings.HasPrefix(ep.Targets[0], string(dns.DefaultGeo))
+				if geoCodeProp.Value == "*" && !targetIsDefaultGroup {
+					continue
+				}
+				item := &dnsv1.RRSetRoutingPolicyGeoPolicyGeoPolicyItem{
+					Location: defaultGeo,
```
mikenairn, line 482: Is this not going to always set the location value to `defaultGeo`?

pkg/dns/google/google.go
```diff
@@ -0,0 +1,500 @@
...
+					Weight:  weight,
+				}
+				record.RoutingPolicy.Wrr.Items = append(record.RoutingPolicy.Wrr.Items, item)
+			}
+			if geoCode {
+				geoCodeProp, _ := ep.GetProviderSpecificProperty(dns.ProviderSpecificGeoCode)
+				targetIsDefaultGroup := strings.HasPrefix(ep.Targets[0], string(dns.DefaultGeo))
+				if geoCodeProp.Value == "*" && !targetIsDefaultGroup {
+					continue
+				}
```
mikenairn, line 480: You should probably add this [comment](https://github.com/Kuadrant/multicluster-gateway-controller/commit/55ff95795d94099bca4bc46c1a63bc83af829dbd#diff-eea59a6fb9851390f6210a03f57ba88ea1628c32d20d526998600135a1642bedR479-R481), or something similar in your own words, explaining why we are having to do this as it's probably not immediately obvious.

pkg/dns/google/google.go
R-Lawton: yes your right

pkg/controllers/dnsrecord/dnsrecord_controller.go
```diff
@@ -158,7 +158,7 @@ func (r *DNSRecordReconciler) deleteRecord(ctx context.Context, dnsRecord *v1alp
 
 	err = dnsProvider.Delete(dnsRecord, managedZone)
 	if err != nil {
-		if strings.Contains(err.Error(), "was not found") {
+		if strings.Contains(err.Error(), "was not found") || strings.Contains(err.Error(), "notFound") {
```
mikenairn, line 161: Created follow on task for this https://github.com/Kuadrant/multicluster-gateway-controller/issues/458

pkg/controllers/managedzone/managedzone_controller.go
```diff
@@ -197,7 +197,7 @@ func (r *ManagedZoneReconciler) deleteManagedZone(ctx context.Context, managedZo
 	}
 	err = dnsProvider.DeleteManagedZone(managedZone)
 	if err != nil {
-		if strings.Contains(err.Error(), "was not found") {
+		if strings.Contains(err.Error(), "was not found") || strings.Contains(err.Error(), "notFound") {
```
mikenairn, line 200: Same here https://github.com/Kuadrant/multicluster-gateway-controller/issues/458

pkg/dns/google/google.go
```diff
@@ -0,0 +1,507 @@
...
+					Rrdatas: targets,
+					Weight:  weight,
+				}
+				record.RoutingPolicy.Wrr.Items = append(record.RoutingPolicy.Wrr.Items, item)
+			}
+			if geoCode {
+				geoCodeProp, _ := ep.GetProviderSpecificProperty(dns.ProviderSpecificGeoCode)
+				geoCodeValue := geoCodeProp.Value
+				targetIsDefaultGroup := strings.HasPrefix(ep.Targets[0], string(dns.DefaultGeo))
+				// GCP doesnt except * as value for default geolocations like AWS does.
```
mikenairn, line 479:
```suggestion
				// GCP doesn't accept * as value for default geolocations like AWS does.
```

test/e2e/gateway_single_spoke_test.go
```diff
@@ -292,8 +292,8 @@ var _ = Describe("Gateway single target cluster", func() {
 
 					By("ensuring the authoritative nameserver resolves the hostname")
 
-					// speed up things by using the authoritative nameserver
```
mikenairn, line 295: Can the chnages in this file just be reverted?

.github/workflows/ci-e2e.yaml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - GH-258-GCP
```
mikenairn, line 7: Can be removed now

mikenairn:
I haven't tried it locally this time, but the e2e test is working now so think we can merge.

/lgtm
/approve

Issue #324, opened 2023-07-11
Add PolicyAffected status condition for gateway


maleck13:
With DNSPolicy targeting gateways, it is import we update the gateway status with a condition to let the owner know the gateway is affected by a given DNSPolicy resource. This will also bring us into line with the latest proposed changes around policy discovery

https://github.com/kubernetes-sigs/gateway-api/pull/2128/files#diff-afe84021d0647e83f420f99f5d18b392abe5ec82d68f03156c7534de9f19a30aR888-R899

https://gateway-api.sigs.k8s.io/geps/gep-713/#conditions

Issue #374, opened 2023-07-25
Implement TLSPolicy


jasonmadigan:
**What**

Move all TLS logic into a new TLSPolicy controller, with distinct reconciliation.

Currently, the gateway controller reconciliation will create and update TLS certificates. The proposed TLSPolicy targets a Gateway and its reconciliation will ensure certificates are in place for a given Gateway. This is more in line with our newly adopted DNSPolicy, which manages all DNS resources for a target Gateway.

**How**

- [x] #378
- [x] Look to relocate & re-use `cluster_eventmapper.go` & `gateway_eventmapper.go`, as they'll be useful for TLSPolicy as well.
- [x] Remove old certificate management from the existing gateway controller (e.g. `reconcileTLS`)
- [ ] Relocate certificate management services/structs that exist in the gateway controller, to the new TLSPolicy controller (e.g. `CertificateService`)
- [x] Handle upstream Gateway removal and cleanup of associated resources (e.g. `Certificates.CleanupCertificates`)
- [ ] For now, create a default TLSPolicy if none exist, with a default issuer (see `glbc-ca` default, self-signed issuer currently). Note: This could do with some discussion as to how this is created - this is mostly useful for devs, although it also has some usefulness for those users who are getting started with Kuadrant. It may be more appropriate to create this via our setup scripts outside of the running process.
- [x] TLSPolicy will use a similar policy attachment pattern as other resources such as R`ateLimitPolicy`, `AuthPolicy` and `DNSPolicy`.
- [x] A standalone description document for TLSPolicy should be created, similar to https://github.com/Kuadrant/multicluster-gateway-controller/pull/373
- [x] If we choose to not-create TLSPolicy by default, we should alter the steps present in the OCM setup walkthrough to create a working default self-signed issuer.
- [x] Kuadrant policies retain an annotation on targetted Gateways, which will be useful to build lists of "managed" Gateways

At a high level, reconciliation should cover:
- [x] TLSPolicy creation. Triggers reconcile, create certificates & keys for Gateway listeners, similar to how these are created currently (but now centralised in the TLSPolicy controller).
- [x] Gateway deletion - triggers reconcile and delete associated resources owned by TLSPolicy (certificates and secrets).
- [x] TLSPolicy deletion: deletes resources owned by TLSPolicy. Remove TLSPolicy annotation from targetted Gateways.
- [x] TLSPolicy target ref change - cleanup existing resources, recreate afresh for new target gateway next loop.
[...]

PR #402, opened 2023-08-03
378 Add TLSPolicy

### LINKED ISSUE #378: Create a new TLSPolicy CR and reconciler
**What**

Add a new CRD for TLSPolicy it should contain a target ref that targets a gateway and be similar in form the DNSPolicy reconciler. The goal here is to create a basic policy the decides the issuer of the cert and to refactor the TLS logic to be controlled by a new TLSPolicy controller

Add one property to this CR for now `IssuerName` (we will add more later)

- [x]  Move all the TLS logic from the tls service into the TLS controller
- [x] Change the logic in the controller to only setup TLS certificates using the chosen issuer when a TLSPolicy is present
- [x] TLS policy ownes the certificate resources
- [x] Update and add new tests
- [x] Add basic documentation for TLSPolicy

mikenairn:
TLSPolcy implementation

Implements the TLSPolicy as a Gateway API Policy using Direct Policy attachment.

The TLSPolicy copies the approach taken by cert manager for securing gateways, and is essentially a policy version of what is described here https://cert-manager.io/docs/usage/gateway/#supported-annotations.
Note: This behaviour differs from what we used to do, and is not a direct translation of what the gateway controller did previously. It is expected that the gateway owner(admin) setup the tls section as required and the TLSPolicy processes it, it does not add the tls section to the gateway. Also TLS certs/secrets can be shared amongst similar hosts, up to the gateway admin to decide this.

Completely separates the TLS management using cert manager from OCM gateway placement. The gateway controller no longer cares about how tls secrets are created, instead it will only try and sync secrets that exist regardless of how they were created. This approach will allow for hosts not currently managed in the configured tls provider account to still be added and synced to the downstream clusters.

/closes #378 

Note: This does not add the policy affected status condition to the target gateway re. https://github.com/Kuadrant/multicluster-gateway-controller/issues/324.


config/crd/patches/cainjection_in_tlspolicies.yaml
```diff
@@ -0,0 +1,7 @@
+# The following patch adds a directive for certmanager to inject CA into the CRD
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    cert-manager.io/inject-ca-from: $(CERTIFICATE_NAMESPACE)/$(CERTIFICATE_NAME)
```
maleck13, line 6: what does this actually do?

pkg/controllers/tlspolicy/tlspolicy_controller.go
```diff
@@ -0,0 +1,231 @@
...
+			return err
+		}
+	}
+
+	// update annotation of policies affecting the gateway
+	return r.ReconcileGatewayPolicyReferences(ctx, tlsPolicy, gatewayDiffObj)
+}
+
+func (r *TLSPolicyReconciler) calculateStatus(tlsPolicy *v1alpha1.TLSPolicy, specErr error) *v1alpha1.TLSPolicyStatus {
+	newStatus := tlsPolicy.Status.DeepCopy()
```
david-martin, line 195: Some thoughts on TLSPolicy status in https://github.com/Kuadrant/multicluster-gateway-controller/issues/374#issuecomment-1650172511 if you haven't already seen those comments.

config/crd/bases/kuadrant.io_tlspolicies.yaml
```diff
@@ -0,0 +1,295 @@
...
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: TLSPolicySpec defines the desired state of TLSPolicy
+            properties:
+              certManager:
```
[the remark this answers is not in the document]
mikenairn: Went back and forth on this one, we could inline it `CertManagerSpec json:",inline"`, but this would allow you to add specs for others as unlikely as that may be. I'm easy either way tbh, this just makes it a bit more obvious it's certManager we/users are dealing with.

config/crd/patches/cainjection_in_tlspolicies.yaml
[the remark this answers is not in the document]
mikenairn: It's not, see above re. generated by kubebuilder. Will be removed before merge.

hack/local-setup.sh
```diff
@@ -368,33 +368,33 @@ initController ${KIND_CLUSTER_CONTROL_PLANE}
...
-    deployAgentSecret ${KIND_CLUSTER_WORKLOAD}-${i} "true"
-    deployAgentSecret ${KIND_CLUSTER_WORKLOAD}-${i} "false"
+#    deployOLM ${KIND_CLUSTER_WORKLOAD}-${i}
+#    deployDashboard ${KIND_CLUSTER_WORKLOAD}-${i} ${i}
+#    argocdAddCluster ${KIND_CLUSTER_CONTROL_PLANE} ${KIND_CLUSTER_WORKLOAD}-${i}
+#    deployAgentSecret ${KIND_CLUSTER_WORKLOAD}-${i} "true"
+#    deployAgentSecret ${KIND_CLUSTER_WORKLOAD}-${i} "false"
     deployOCMSpoke ${KIND_CLUSTER_WORKLOAD}-${i}
-    deployPrometheusForFederation ${KIND_CLUSTER_WORKLOAD}-${i}
+#    deployPrometheusForFederation ${KIND_CLUSTER_WORKLOAD}-${i}
```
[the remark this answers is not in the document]
mikenairn: Yes, can be ignored, didn't intend to commit these changes.

config/samples/tlspolicy.yaml
```diff
@@ -0,0 +1,22 @@
...
+spec:
+  targetRef:
+    name: prod-web
+    group: ""
+    kind: Gateway
+  issuerRef:
+    group: cert-manager.io
+    kind: ClusterIssuer
+    name: glbc-ca
+  secretName: todo-host
```
maleck13, line 14: hmm how does this work with multiple listeners?

maleck13: so this can be used independent of DNS as the issuer will have the credentials, nice

maleck13 (commented):
This is looking good to me. Little bit unsure about the certmanger property but prob ok as first pass alpha API. Also tried this out locally and works as expected.

@mikenairn whats left before this can be moved from a WIP?

pkg/controllers/tlspolicy/tlspolicy_controller.go
maleck13: Yeah at least a ready condition.

mikenairn: There is a ready condition that will be set to false on spec reconcile failure, with a reason of `ReconciliationError` and message containing the error. I guess if we want more tailored messages based on types of errors we can look into that, but this is a bit of a catch all currently.

.github/workflows/ci-e2e.yaml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - 378_add-tls-policy
```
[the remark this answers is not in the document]
mikenairn: Since these changes can't be tested on this PR, it's running here https://github.com/Kuadrant/multicluster-gateway-controller/actions/runs/5832446191/job/15817886112

mikenairn:
/hold

Couple of small things to fix before this can be merged, but is good for a review.

pkg/controllers/tlspolicy/tlspolicy_controller.go
```diff
@@ -0,0 +1,271 @@
...
+				return ctrl.Result{}, err
+			}
+		}
+		return ctrl.Result{}, nil
+	}
+
+	// add finalizer to the tlsPolicy
+	if !controllerutil.ContainsFinalizer(tlsPolicy, TLSPolicyFinalizer) {
+		if err := r.AddFinalizer(ctx, tlsPolicy, TLSPolicyFinalizer); client.IgnoreNotFound(err) != nil {
+			return ctrl.Result{Requeue: true}, err
```
maleck13, line 120: if you return an err as I understand it, it will auto get requeued? So do we need the Requeue true and the err?

maleck13:
took another look though. All looks pretty reasonable. @david-martin you might give it another review when you are back

config/crd/kustomization.yaml
```diff
@@ -6,15 +6,9 @@ resources:
...
 - bases/kuadrant.io_dnshealthcheckprobes.yaml
+- bases/kuadrant.io_tlspolicies.yaml
 #+kubebuilder:scaffold:crdkustomizeresource
 
-patchesStrategicMerge:
-# [CERTMANAGER] To enable cert-manager, uncomment all the sections with [CERTMANAGER] prefix.
-# patches here are for enabling the CA injection for each CRD
-#- patches/cainjection_in_dnsrecords.yaml
-#- patches/cainjection_in_managedzones.yaml
-#+kubebuilder:scaffold:crdkustomizecainjectionpatch
```
mikenairn, line 16: Just removed the lot, we will never use this

pkg/controllers/tlspolicy/tlspolicy_controller.go
mikenairn: I think we should try and keep ourselves aligned with other kuadrant operators and look to do whatever ever is agreed here https://github.com/Kuadrant/architecture/pull/9

pkg/controllers/tlspolicy/tlspolicy_controller.go
```diff
@@ -0,0 +1,269 @@
...
+
+func (r *TLSPolicyReconciler) reconcileResources(ctx context.Context, tlsPolicy *v1alpha1.TLSPolicy, targetNetworkObject client.Object) error {
+	// validate
+	err := tlsPolicy.Validate()
+	if err != nil {
+		return err
+	}
+
+	// reconcile based on gateway diffs
+	gatewayDiffObj, err := r.ComputeGatewayDiffs(ctx, tlsPolicy, targetNetworkObject, &TLSPolicyRefsConfig{})
```
david-martin, line 155: Does this work for targeted HTTPRoutes as well?

pkg/controllers/tlspolicy/tlspolicy_controller.go
```diff
@@ -0,0 +1,269 @@
...
+			&source.Kind{Type: &gatewayapiv1beta1.Gateway{}},
+			handler.EnqueueRequestsFromMapFunc(gatewayEventMapper.MapToPolicy),
+		).
+		Complete(r)
+}
+
+// The following methods are here temporarily and copied from the kuadrant-operator https://github.com/Kuadrant/kuadrant-operator/blob/main/pkg/reconcilers/targetref_reconciler.go#L45
+// FetchValidTargetRef and FetchValidGateway currently expect that the gateway should have a ready condition, but in the
+// case of the TLSPolicy it might not be ready because there is an invalid tls section that this policy would resolve.
+// ToDo mnairn: Create issue in kuadrant-operator and link it here
```
david-martin, line 238: Issue been created yet?
Any reason they can't be referenced from the kuadrant-operator?

test/e2e/gateway_single_spoke_test.go
```diff
@@ -190,103 +204,144 @@ var _ = Describe("Gateway single target cluster", func() {
...
+					err := tconfig.SpokeClient(0).Delete(ctx, dnsPolicy)
+					Expect(err).ToNot(HaveOccurred())
+					err = tconfig.SpokeClient(0).Delete(ctx, tlsPolicy)
+					Expect(err).ToNot(HaveOccurred())
+				})
+
+				It("makes available a hostname that can be resolved and reachable through HTTPS", func(ctx SpecContext) {
+
+					// Wait for the DNSrecord to exists: this shouldn't be necessary but I have found out that AWS dns servers
+					// cache the "no such host" response for a period of aprox 10-15 minutes, which makes the test run for a
```
david-martin, line 269: Is this dns negative caching, or an AWS API thing a result of using the same host each test run, or something else?

david-martin (commented):
Some comments & questions.
I'll run through the updated ocm walkthrough before adding lgtm

david-martin (commented):
Verified locally, following steps from ocm walkthrough to include a tls block in the Gateway, and a TLSPolicy resource targeting the Gateway.

The TLSPolicy status looks good.

```
kubectl get tlspolicy -A -o yaml
apiVersion: v1
items:
- apiVersion: kuadrant.io/v1alpha1
  kind: TLSPolicy
  metadata:
    annotations:
      kubectl.kubernetes.io/last-applied-configuration: |
        {"apiVersion":"kuadrant.io/v1alpha1","kind":"TLSPolicy","metadata":{"annotations":{},"name":"prod-web","namespace":"multi-cluster-gateways"},"spec":{"issuerRef":{"group":"cert-manager.io","kind":"ClusterIssuer","name":"glbc-ca"},"targetRef":{"group":"gateway.networking.k8s.io","kind":"Gateway","name":"prod-web"}}}
    creationTimestamp: "2023-08-16T10:50:30Z"
    finalizers:
    - kuadrant.io/tls-policy
    generation: 1
    name: prod-web
    resourceVersion: "7655"
    uid: 9d385ff0-0f6a-4c8f-8c5e-bf738c7c612f
  spec:
    issuerRef:
      group: cert-manager.io
      kind: ClusterIssuer
      name: glbc-ca
    targetRef:
      kind: Gateway
      name: prod-web
  status:
    conditions:
    - lastTransitionTime: "2023-08-16T10:58:22Z"
      message: Gateway is TLS Enabled
      reason: GatewayTLSEnabled
      status: "True"
      type: Ready
    observedGeneration: 1
kind: List
metadata:
  resourceVersion: ""
```

Certificate & Secret exist

```
kubectl get certificate,secret -n multi-cluster-gateways
NAME                                           READY   SECRET             AGE
certificate.cert-manager.io/apps-hcpapps-tls   True    apps-hcpapps-tls   11m

NAME                         TYPE                DATA   AGE
secret/apps-hcpapps-tls      kubernetes.io/tls   3      11m
secret/mgc-aws-credentials   kuadrant.io/aws     3      24m
```

Endpoint can be curled as expected.

```
curl -k -I https://$MGC_SUB_DOMAIN
HTTP/2 200
content-type: text/plain
date: Wed, 16 Aug 2023 10:59:47 GMT
content-length: 1610
x-envoy-upstream-service-time: 1
server: istio-envoy
```

docs/dns-policy.md
```diff
@@ -94,8 +94,8 @@ spec:
       - 201
       - 301
     failureThreshold: 5
-    port: 443
-    protocol: https
+    port: 80
+    protocol: http
```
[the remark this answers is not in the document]
mikenairn: > Given this is an example DNSPolicy, is the reason for the change to http to avoid having to reference TLSPolicy as well?

Yes, since this is DNSPolicy doc i didn't want to talk about TLSPolicy in it and you'd need to add a TLSPolicy now to make it work for https. Also i don't see it needs to, DNS doesn't care if your endpoints are secure or not.

pkg/controllers/tlspolicy/tlspolicy_controller.go
mikenairn: Followed the approach we took with DNSPolicy and it only supports targeting a Gateway currently.

The controller itself follows the same pattern that all kuadrant policies use, i believe these can handle HTTPRoutes as targets, but afaik we don't currently.

pkg/controllers/tlspolicy/tlspolicy_controller.go
mikenairn: > Issue been created yet? 

No, will do it now and update.

> Any reason they can't be referenced from the kuadrant-operator?

You mean why is this not using the kuadrant operator functions? Reason is this check https://github.com/Kuadrant/kuadrant-operator/blob/main/pkg/reconcilers/targetref_reconciler.go#L55-L57. In the case of TLS we don't want it to do that since it could be not ready because of TLS and we want this to continue to create the TLS certificate and secrets that might bring the gateway into a programmed state.

PR #409, opened 2023-08-08
Additional headers as secret

philbrookes:
Change additional headers to a secret ref, so that users can store auth keys, etc. in it.

Update DNS Policy document with new fields. 

## verification
- create a workload and see DNSRecords
- Create a secret with a key / value pair, where the value ends with "\n" (e.g. "test\n": `dGVzdAo=`)
- Create a health check spec that references this secret
- See the health check failing due to the invalid header (may take upto a minute)
- Change the secret value to no longer end with "\n" (e.g. "test": `dGVzdA==`)
- See the health check begin passing (may take upto a minute)


config/crd/bases/kuadrant.io_dnshealthcheckprobes.yaml
```diff
@@ -45,22 +45,23 @@ spec:
...
-                    name:
-                      type: string
-                    value:
-                      type: string
-                  required:
-                  - name
-                  - value
-                  type: object
-                type: array
+                properties:
```
maleck13, line 49: this is a bit confusing. Am I refrencing a secret?

philbrookes: yes, this is following @roivaz suggestion, in case the headers contained an authorization token or similar.

pkg/controllers/dnshealthcheckprobe/dnshealthcheckprobe_controller.go
```diff
@@ -77,6 +78,21 @@ func (r *DNSHealthCheckProbeReconciler) Reconcile(ctx context.Context, req ctrl.
 
 	probeId := probeId(previous)
 
+	additionalHeaders := v1alpha1.AdditionalHeaders{}
+
+	if previous.Spec.AdditionalHeaders != nil {
+		secretKey := client.ObjectKey{Name: previous.Spec.AdditionalHeaders.Name, Namespace: previous.Spec.AdditionalHeaders.Namespace}
```
roivaz, line 84: a namespaced resource shouldn't reference secrets in a different namespace, maybe we should just drop the `namespace` field and use the same namespace where the dnshealthcheckprobe lives?

config/crd/bases/kuadrant.io_dnshealthcheckprobes.yaml
maleck13: ok I see, I think then it should be consistent with other refs
IE for tls it is
```certificateRefs```
so for headers should prob be 
```additionalHeadersRef``` ?

File: docs/dns-policy.md [hunk 3]
```diff
@@ -93,6 +107,17 @@ spec:
 - `Name` is the name of the target resource.
 - `Namespace` is the namespace of the referent. Currently only local objects can be referred so value is ignored.
 
+### Health Check
+The health check section is optional, the following fields are available:
+
+`allowInsecureCertificates`: Added for development environments, allows health probes to not fail when finding an invalid (e.g. self-signed) certificate.
+`additionalHeaders`: A reference to a secret which contains additional headers such as an authentication token
```
maleck13, line 114: also this

pkg/controllers/dnshealthcheckprobe/dnshealthcheckprobe_controller.go
```diff
@@ -77,6 +78,21 @@ func (r *DNSHealthCheckProbeReconciler) Reconcile(ctx context.Context, req ctrl.
...
+
+	if previous.Spec.AdditionalHeadersRef != nil {
+		secretKey := client.ObjectKey{Name: previous.Spec.AdditionalHeadersRef.Name, Namespace: previous.Namespace}
+		additionalHeadersSecret := &v1.Secret{}
+		if err := r.Client.Get(ctx, secretKey, additionalHeadersSecret); err != nil {
+			return ctrl.Result{}, fmt.Errorf("error retrieving additional headers secret %v/%v: %w", secretKey.Namespace, secretKey.Name, err)
+		}
+		for k, v := range additionalHeadersSecret.Data {
+			additionalHeaders = append(additionalHeaders, v1alpha1.AdditionalHeader{
+				Name:  k,
```
maleck13, line 91: any validation we should do here?

maleck13 (commented):
should there be a unit test for the additional headers work?

pkg/controllers/dnshealthcheckprobe/dnshealthcheckprobe_controller.go
philbrookes: I don't see what we could realistically validate here, a little searching hasn't led me to any sort of generic header validation. Did you have anything in mind, or were there some concerns about this, that I'm missing?

Considering that these headers can only be sent in requests to clusters that have had a gateway deployed to them from the same MGC - allowing them to define any headers they like seems fine?

philbrookes:
> should there be a unit test for the additional headers work?

I'm not seeing a meaningful place to put unit tests for this, but I think I can see how to write some tests around them in the envtest suite, so I'll do that now.

pkg/controllers/dnshealthcheckprobe/dnshealthcheckprobe_controller.go
maleck13: well trimming space. failing on spaces in the key? the values prob not though

maleck13:
I'm not seeing a meaningful place to put unit tests for this, but I think I can see how to write some tests around them in the envtest suite, so I'll do that now.

2 things come to mind now
1) so currently no unit tests around the health checks?
2) could we not isolate the additonal header logic in a single function that could be tested as a unit?

maleck13:
code changes looking reasonable will do a quick validation locally

docs/dns-policy.md
```diff
@@ -84,6 +84,18 @@ spec:
     name: prod-web
     group: gateway.networking.k8s.io
     kind: Gateway
+  healthCheck:
+    allowInsecureCertificates: true
```
maleck13, line 88: this still looks wrong? in the CRD is `allowInsecureCertificate` https://github.com/Kuadrant/multicluster-gateway-controller/pull/409/files#diff-72079cb887312d7bc89d463be464b6882d63251f812ef1efbef663182e4a6d4dR34 but here it is `allowInsecureCertificates`

maleck13 (approved):
tried locally. Couldn't connet to the app but that is an issue locally otherwise looks good to merge now /lgtm

Issue #272, opened 2023-06-21
Remove unhealthy DNS Records from DNS provider


sergioifg94:
## What

In order to support the ability of endpoints to be taken down when the associated health check fails, we must implement the logic in the DNSRecord controller to watch for status updates in the associated DNSHealthCheckProbe, and disable the endpoint if the endpoint is unhealthy and the number of consecutive failures is greater than the failure threshold configured in the DNSPolicy, as well as re-enable it when it becomes healthy again.

This removal of a DNS Record will need to cascade through the DNS Record to ensure no dead-ends are created where a CNAME has no records to resolve to.

## DNS Record Structure Diagram:
https://lucid.app/lucidchart/2f95c9c9-8ddf-4609-af37-48145c02ef7f/edit?viewport_loc=-188%2C-61%2C2400%2C1183%2C0_0&invitationId=inv_d5f35eb7-16a9-40ec-b568-38556de9b568

## How

When a DNS health check probe is failing, it will update the DNS Record CR with a custom field on that endpoint to mark it as failing.

There are then 3 scenarios which we need to consider:
1 - All endpoints are healthy
2 - All endpoints are unhealthy
3 - Some endpoints are healthy and some are unhealthy.

In the cases 1 and 2, the result should be the same: All records are published to the DNS Provider.

When scenario 3 is encountered the following process should be followed:
- For each gateway IP or CNAME: this should be omitted if unhealthy.
- For each managed gateway CNAME: This should be omitted if all child records are unhealthy.
- For each GEO CNAME: This should be omitted if all the managed gateway CNAMEs have been omitted.
- Load balancer CNAME: This should never be omitted.

If we consider the DNS record to be a hierarchy of parents and children, then whenever any parent has no healthy children that parent is also considered unhealthy. No unhealthy elements are to be included in the DNS Record.

maleck13:
@philbrookes this is done in the DNS policy controller now under reconcileGatewayDNSRecords

PR #412, opened 2023-08-08
Delete Endpoint from DNSRecord when endpoint it's related probe is marked with status healthy: false

laurafitzgerald:
### **What**

Closes https://github.com/Kuadrant/multicluster-gateway-controller/issues/272

### **Changes**

- [x] Trigger a reconcile of the DNSPolicy if there is an update to any DNSHealthProbeCheck.
- [x] Remove enpoint record if it's related probe is unhealthy.
- [x] traverse the chain of records and remove any that have only unhealthy children to prevent NXDOMAIN responses as per https://github.com/Kuadrant/multicluster-gateway-controller/pull/431
- [x] unit test updates
- [x] integration tests

### **Verification**

follow setup as per the https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/docs/how-to/ocm-control-plane-walkthrough.md

NOTE: there are a few adjustments to the steps mentioned below

Before adding any of the resources to the clusters scale down the copy of the controller running from the quickstart using `kc edit deployment mgc-controller-manager -n multicluster-gateway-controller-system` and run the copy of this branch against the cluster using `(export $(cat ./controller-config.env | xargs) && export $(cat ./aws-credentials.env | xargs) && make build-controller install run-controller)`

Create secret `kubectl create secret generic probe-headers --from-literal=test=dGVzdAo= -n multi-cluster-gateways`

The following DNSPolicy can be used to verify instead of the one recommended in the walkthrough. Note: the intentionally low threshold of 1 for testing.

```
kubectl apply -f - <<EOF
kind: DNSPolicy
metadata:
  name: prod-web
spec:
  targetRef:
    name: prod-web
    kind: Gateway
  healthCheck:
    additionalHeadersRef:
      name: probe-headers
    endpoint: /
    expectedResponses:
      - 200
      - 201
      - 301
    failureThreshold: 1
    port: 443
    protocol: https
EOF
```

## Verify that the correct probes are in place.

Check the Gateway Policy for the number of addresses . 

`kubectl get gateway prod-web -n multi-cluster-gateways -o yaml | yq -r '.status.addresses'`

You should see something like: 
```
  value: kind-mgc-control-plane/172.31.200.0
  value: kind-mgc-workload-1/172.31.201.0
```

Verify the probes created match that. 
`kc get dnshealthcheckprobes -A`
e.g. 
```
kc get dnshealthcheckprobes -A        
NAMESPACE                NAME                    HEALTHY   LAST CHECKED
multi-cluster-gateways   172.31.200.0-prod-web   true     48s
multi-cluster-gateways   172.31.201.0-prod-web   true      48s
```

Verify endpoints are present in the DNSRecord
`kubectl get dnsrecord prod-web-api -n multi-cluster-gateways -o yaml | yq -r '.spec.endpoints'`

example output
```
- dnsName: 2c71gf.lb-4ej5le.test.lf.hcpapps.net
  recordTTL: 60
  recordType: A
  targets:
    - 172.31.200.0
- dnsName: default.lb-4ej5le.test.lf.hcpapps.net
  providerSpecific:
    - name: weight
      value: "120"
  recordTTL: 60
  recordType: CNAME
  setIdentifier: 2c71gf.lb-4ej5le.test.lf.hcpapps.net
  targets:
    - 2c71gf.lb-4ej5le.test.lf.hcpapps.net
- dnsName: default.lb-4ej5le.test.lf.hcpapps.net
  providerSpecific:
    - name: weight
      value: "120"
  recordTTL: 60
  recordType: CNAME
  setIdentifier: lrnse3.lb-4ej5le.test.lf.hcpapps.net
  targets:
    - lrnse3.lb-4ej5le.test.lf.hcpapps.net
- dnsName: lb-4ej5le.test.lf.hcpapps.net
  providerSpecific:
    - name: geo-code
      value: '*'
  recordTTL: 300
  recordType: CNAME
  setIdentifier: default
  targets:
    - default.lb-4ej5le.test.lf.hcpapps.net
- dnsName: lrnse3.lb-4ej5le.test.lf.hcpapps.net
  recordTTL: 60
  recordType: A
  targets:
    - 172.31.201.0
- dnsName: test.lf.hcpapps.net
  recordTTL: 300
  recordType: CNAME
  targets:
    - lb-4ej5le.test.lf.hcpapps.net
```

Scale down a service to produce an unhealthy probe
`kubectl edit deployment -n default` - update the replicas to 0

Wait for one probe to go unhealthy 
`watch -n1 "kubectl get dnshealthcheckprobes -A"`

Verify that the related endpoint is removed from the DNSRecord.
`kubectl get dnsrecord prod-web-api -n multi-cluster-gateways -o yaml | yq -r '.spec.endpoints'`

example output, Note that no endpoint has a target that doesn't also have a endpoint of the same name, aka no dead records
```
- dnsName: default.lb-4ej5le.test.lf.hcpapps.net
  providerSpecific:
    - name: weight
      value: "120"
  recordTTL: 60
  recordType: CNAME
  setIdentifier: lrnse3.lb-4ej5le.test.lf.hcpapps.net
  targets:
    - lrnse3.lb-4ej5le.test.lf.hcpapps.net
- dnsName: lb-4ej5le.test.lf.hcpapps.net
  providerSpecific:
    - name: geo-code
      value: '*'
  recordTTL: 300
  recordType: CNAME
  setIdentifier: default
  targets:
    - default.lb-4ej5le.test.lf.hcpapps.net
- dnsName: lrnse3.lb-4ej5le.test.lf.hcpapps.net
  recordTTL: 60
  recordType: A
  targets:
    - 172.31.201.0
- dnsName: test.lf.hcpapps.net
  recordTTL: 300
  recordType: CNAME
  targets:
    - lb-4ej5le.test.lf.hcpapps.net
```

`watch dig $MGC_SUB_DOMAIN` should never return NXDOMAIN error. 

Log into the workload cluster

Scale down a service to produce an unhealthy probe
`kubectl edit deployment -n default` - update the replicas to 0

There now should be no health endpoints for the gateway as neither pod service is running.

There should now be no endpoints showing as ready. 
`kubectl get dnsrecord prod-web-api -n multi-cluster-gateways -o yaml | yq -r '.spec.endpoints'`
Output
`null`

```
curl -k https://$MGC_SUB_DOMAIN
no healthy upstream%   
 ```

Scale back up one of the pods for the echo api/

Verify that endpoints are returned related to the service

`kubectl get dnsrecord prod-web-api -n multi-cluster-gateways -o yaml | yq -r '.spec.endpoints'`

```
- dnsName: 2c71gf.lb-4ej5le.test.lf.hcpapps.net
  recordTTL: 60
  recordType: A
  targets:
    - 172.31.200.0
- dnsName: default.lb-4ej5le.test.lf.hcpapps.net
  providerSpecific:
    - name: weight
      value: "120"
  recordTTL: 60
  recordType: CNAME
  setIdentifier: 2c71gf.lb-4ej5le.test.lf.hcpapps.net
  targets:
    - 2c71gf.lb-4ej5le.test.lf.hcpapps.net
- dnsName: lb-4ej5le.test.lf.hcpapps.net
  providerSpecific:
    - name: geo-code
      value: '*'
  recordTTL: 300
  recordType: CNAME
  setIdentifier: default
  targets:
    - default.lb-4ej5le.test.lf.hcpapps.net
- dnsName: test.lf.hcpapps.net
  recordTTL: 300
  recordType: CNAME
  targets:
    - lb-4ej5le.test.lf.hcpapps.net
```

### CODE REVIEW

**Review by maleck13 — COMMENTED** [review]
@philbrookes @sergioifg94 are we removing the records or setting the weight to 0 I had thought we would try to avoid removing records when unhealthy as this will lead to client issues that still reference that record?

**Comment by philbrookes** [comment]
> @philbrookes @sergioifg94 are we removing the records or setting the weight to 0 I had thought we would try to avoid removing records when unhealthy as this will lead to client issues that still reference that record?

I also think we should be removing these using  weight '0' when that is an option, something may have gotten lost in translation here. 

However, I am not convinced that this will address the concerns of a client referencing a non-existing record as I suspect that AWS considers a weight 0 record and a deleted record as equally "not to be published".

**Comment by laurafitzgerald** [comment]
> > @philbrookes @sergioifg94 are we removing the records or setting the weight to 0 I had thought we would try to avoid removing records when unhealthy as this will lead to client issues that still reference that record?
> 
> I also think we should be removing these using weight '0' when that is an option, something may have gotten lost in translation here.
> 
> However, I am not convinced that this will address the concerns of a client referencing a non-existing record as I suspect that AWS considers a weight 0 record and a deleted record as equally "not to be published".

Hey guys. sorry if there was a misunderstanding here. To clarify then what I think we expect.
If a endpoint is deemed unhealthy
The weight remains the same in the cr.
A value is set in the providerSpecific values to somehow represent that the endpoint is marked as unhealthy.
The weight in the actual provider is set to 0. note this can be done for AWS and GCP but not for Azure (Azure begins at 1)

**Comment by philbrookes** [comment]
> 

Yes, this sounds good to me. The DNS Record should reflect the ideal state of the DNS, assuming everything was working, and the weight is set to zero in the provider specific package.

**Comment by mikenairn** [comment]
Is there any reason we need to do this in the provider package? I'd have expected this to be done higher up during the decision making for the weight when it's building up the list of endpoints. This is just logic we will have to replicate in all providers as we add new ones.

If it's the zero vs delete thing, you can just make it that we set the weight to 0 for unhealthy endpoints and pass that to the provider. It would be the exception to have to do extra processing at the provider level to deal with it (Azure not supporting a weight of 0). This would work for google and aws without changing the provider implementations.

**Comment by laurafitzgerald** [comment]
> Is there any reason we need to do this in the provider package? I'd have expected this to be done higher up during the decision making for the weight when it's building up the list of endpoints. This is just logic we will have to replicate in all providers as we add new ones.
> 
> If it's the zero vs delete thing, you can just make it that we set the weight to 0 for unhealthy endpoints and pass that to the provider. It would be the exception to have to do extra processing at the provider level to deal with it (Azure not supporting a weight of 0). This would work for google and aws without changing the provider implementations.

The idea aiui is that the weight in the cr remains the same but we set the value in the provider to 0.

**Comment by laurafitzgerald** [comment]
> 

I have no objection to setting it to 0. I think we decided not to set it to 0 but rather use a flag because there may be some specific requirements from other providers how to unpublish (e.g. Azure doesn't allows 0) So if we do go with just setting it to 0 we may have other providers which will need maintain some intelligent code inside. So seems like we can't avoid that altogether. I think this is why we decided on the flag.

**Comment by laurafitzgerald** [comment]
> > The idea aiui is that the weight in the cr remains the same but we set the value in the provider to 0.
> 
> The DNSRecord API is currently fairly simple and IMO it should remain that way. It just reconciles the list of endpoints into the provider with no magic involved, if the endpoint is in the spec it will be represented in the provider after it's reconciled. Any reason we are trying to avoid just updating the DNSRecord? If the weight is 0, regardless of the reason, it's 0 and seems logical that it be reflected in the DNSRecord spec.

@mikenairn I have no objection to setting it to 0. I think we decided not to set it to 0 but rather use a flag because there may be some specific requirements from other providers how to unpublish (e.g. Azure doesn't allows 0) So if we do go with just setting it to 0 we may have other providers which will need maintain some intelligent code inside. So seems like we can't avoid that altogether. I think this is why we decided on the flag.

**Comment by philbrookes** [comment]
> > > The idea aiui is that the weight in the cr remains the same but we set the value in the provider to 0.
> > 
> > 
> > The DNSRecord API is currently fairly simple and IMO it should remain that way. It just reconciles the list of endpoints into the provider with no magic involved, if the endpoint is in the spec it will be represented in the provider after it's reconciled. Any reason we are trying to avoid just updating the DNSRecord? If the weight is 0, regardless of the reason, it's 0 and seems logical that it be reflected in the DNSRecord spec.
> 
> @mikenairn I have no objection to setting it to 0. I think we decided not to set it to 0 but rather use a flag because there may be some specific requirements from other providers how to unpublish (e.g. Azure doesn't allows 0) So if we do go with just setting it to 0 we may have other providers which will need maintain some intelligent code inside. So seems like we can't avoid that altogether. I think this is why we decided on the flag.

Just putting it in writing here, that we discussed this on a call, and have agreed to proceed for now with deletion of unhealthy DNS Records from the provider and the DNS Record spec.

File: pkg/controllers/dnshealthcheckprobe/dnshealthcheckprobe_controller.go [hunk 1]
```diff
@@ -149,14 +149,14 @@ func getAdditionalHeaders(ctx context.Context, clt client.Client, probeObj *v1al
 			return additionalHeaders, fmt.Errorf("error retrieving additional headers secret %v/%v: %w", secretKey.Namespace, secretKey.Name, err)
 		} else if err != nil {
 			probeError := fmt.Errorf("error retrieving additional headers secret %v/%v: %w", secretKey.Namespace, secretKey.Name, err)
-			probeObj.Status.Healthy = false
+			*probeObj.Status.Healthy = false
```
philbrookes, line 152: this should be nil

pkg/controllers/dnshealthcheckprobe/dnshealthcheckprobe_controller.go
```diff
@@ -149,14 +149,14 @@ func getAdditionalHeaders(ctx context.Context, clt client.Client, probeObj *v1al
...
 			probeObj.Status.ConsecutiveFailures = 0
 			probeObj.Status.Reason = "additional headers secret not found"
 			return additionalHeaders, probeError
 		}
 		for k, v := range additionalHeadersSecret.Data {
 			if strings.ContainsAny(strings.TrimSpace(k), " \t") {
-				probeObj.Status.Healthy = false
+				*probeObj.Status.Healthy = false
```
philbrookes, line 159: this should be nil too

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
```diff
@@ -101,58 +104,46 @@ func (r *DNSPolicyReconciler) expectedProbesForGateway(ctx context.Context, gw c
...
+				Name:      fmt.Sprintf("%s-%s", matches[0], dnsPolicy.Name),
+				Namespace: gw.Namespace,
+				Labels:    commonDNSRecordLabels(client.ObjectKeyFromObject(gw), client.ObjectKeyFromObject(dnsPolicy)),
+				Annotations: map[string]string{
+					AnnotationHealthCheckProbeEndpointName: dnsPolicy.Name,
+				},
+			},
+			Spec: v1alpha1.DNSHealthCheckProbeSpec{
+				Port:                     *port,
+				Host:                     string(*gw.Spec.Listeners[0].Hostname),
```
philbrookes, line 134: should we not iterate over the listeners here, rather than hardcoding to only read the first one?

pkg/controllers/dnspolicy/dns_helper.go
```diff
@@ -239,8 +241,31 @@ func (dh *dnsHelper) setEndpoints(ctx context.Context, mcgTarget *dns.MultiClust
...
+		if probe != nil {
+			probeHealthy := true
+			if probe.Status.Healthy != nil {
+				probeHealthy = *probe.Status.Healthy
+			}
+			if !probeHealthy || probe.Status.ConsecutiveFailures <= *probe.Spec.FailureThreshold {
+				newEndpoints = append(newEndpoints[:i], newEndpoints[i+1:]...)
+				i--
+			}
+		}
```
sergioifg94, line 261:
```suggestion
		if probe == nil {
				continue
		}
		
		probeHealthy := true
		if probe.Status.Healthy != nil {
			probeHealthy = *probe.Status.Healthy
		}
		if !probeHealthy || probe.Status.ConsecutiveFailures <= *probe.Spec.FailureThreshold {
			newEndpoints = append(newEndpoints[:i], newEndpoints[i+1:]...)
			i--
		}		
```

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
laurafitzgerald: Yes, sorry meant to make a note that this wasn't the final but just where I wanted to pitstop yesterday. 👍

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
```diff
@@ -101,58 +100,43 @@ func (r *DNSPolicyReconciler) expectedProbesForGateway(ctx context.Context, gw c
...
-						Path:                     dnsPolicy.Spec.HealthCheck.Endpoint,
-						Protocol:                 *dnsPolicy.Spec.HealthCheck.Protocol,
-						Interval:                 metav1.Duration{Duration: 60 * time.Second},
-						AdditionalHeadersRef:     dnsPolicy.Spec.HealthCheck.AdditionalHeadersRef,
-						FailureThreshold:         dnsPolicy.Spec.HealthCheck.FailureThreshold,
-						ExpectedResponses:        dnsPolicy.Spec.HealthCheck.ExpectedResponses,
-						AllowInsecureCertificate: dnsPolicy.Spec.HealthCheck.AllowInsecureCertificates,
-					},
-				}
-				healthChecks = append(healthChecks, withDNSRecord(dnsRecord, healthCheck))
```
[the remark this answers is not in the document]
laurafitzgerald: We've updated the expectedProbeforGateways to not depend on the dnsrecord. @philbrookes any objection to getting the dns record still in the function but only using it for this call?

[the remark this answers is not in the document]
sergioifg94: I see that now we create the probes by iterating through `.status.addresses` instead of the endpoints. I wonder if there's a way to find the matching listener for each address and set it as an annotation/label of the probe, then it wouldn't be needed to get the DNSRecord at all

laurafitzgerald:
@philbrookes i pushed a wip commit for the unit tests. the tests are passing but need more work to cover this block https://github.com/Kuadrant/multicluster-gateway-controller/pull/412/files#diff-fc4edceb0a1f16301bbcf2eadce3c8aa8b40b88cb9638f6d872190cb6e6e0cadR264-R306

pkg/controllers/dnspolicy/dns_helper.go
```diff
@@ -355,3 +423,26 @@ func (r *dnsHelper) deleteDNSRecordForListener(ctx context.Context, owner metav1
 func isWildCardListener(l gatewayv1beta1.Listener) bool {
 	return strings.HasPrefix(string(*l.Hostname), "*")
 }
+
+func (dh *dnsHelper) getDNSHealthCheckProbes(ctx context.Context, gateway *gatewayv1beta1.Gateway, dnsPolicy *v1alpha1.DNSPolicy) ([]*v1alpha1.DNSHealthCheckProbe, error) {
+	list := &v1alpha1.DNSHealthCheckProbeList{}
+	if err := dh.List(ctx, list, &client.ListOptions{
+		LabelSelector: labels.SelectorFromSet(commonDNSRecordLabels(client.ObjectKeyFromObject(gateway), client.ObjectKeyFromObject(dnsPolicy))),
+		Namespace:     gateway.Namespace,
```
laurafitzgerald, line 431: The namespace here is needed for the tests. the fake client wont return the probe unless there is a Namespace ListOption. @philbrookes would we expect in a running scenario that the probes would be in a different namespace to the gateway or dnspolicy? This could probably be changed to dnsPolicy.Namespace to be honest.

philbrookes: changing thisto dnspolicy.namespace seems fine to me.

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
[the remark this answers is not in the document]
laurafitzgerald: This should be resolved now. Let me know if there are any further updates needed @sergioifg94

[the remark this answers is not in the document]
sergioifg94: Looks good now on that side, thanks @laurafitzgerald I'll mark it as resolved

pkg/controllers/dnspolicy/dns_helper.go
```diff
@@ -355,3 +437,29 @@ func (r *dnsHelper) deleteDNSRecordForListener(ctx context.Context, owner metav1
...
+	retProbes := []*v1alpha1.DNSHealthCheckProbe{}
+	for _, probe := range probes {
+		for _, target := range endpoint.Targets {
+			if strings.Contains(probe.Name, target) {
+				retProbes = append(retProbes, probe)
+			}
+		}
+	}
+	return retProbes
+}
```
mikenairn, line 465: What endpoints do we current create probes for? Is it just the ultimate targets (the gateway address) or all endpoints?

philbrookes: it's for anything that resolves outside of the DNS Record (i.e. CNAMEs to other zones, or A records)

philbrookes: though the more general answer is: yes, the gateway addresses

mikenairn: We create probes for all endpoints, i see: 
```
kubectl get DNSHealthCheckProbe -A
NAMESPACE                NAME                                                  HEALTHY   LAST CHECKED
multi-cluster-gateways   172.31.200.0-prod-web-api                             false     48s
multi-cluster-gateways   172.31.201.0-prod-web-api                             false     15s
multi-cluster-gateways   172.31.202.0-prod-web-api                             false     28s
multi-cluster-gateways   24osuu.lb-4ej5le.myapp.mn.hcpapps.net-prod-web-api    false     25s
multi-cluster-gateways   2c71gf.lb-4ej5le.myapp.mn.hcpapps.net-prod-web-api    false     45s
multi-cluster-gateways   default.lb-4ej5le.myapp.mn.hcpapps.net-prod-web-api   false     38s
multi-cluster-gateways   lb-4ej5le.myapp.mn.hcpapps.net-prod-web-api           false     35s
multi-cluster-gateways   lrnse3.lb-4ej5le.myapp.mn.hcpapps.net-prod-web-api    false     18s
```

So this logic we are adding is going to check the health of the CNAME records e.g. lb-4ej5le.myapp.mn.hcpapps.net-prod-web-api, if that is "unhealthy", remove that and also anything below it.

What scenario are we trying to cover by checking those explicitly? Is a health probe per gateway address (The IP or hostname being returned in the Gateway) not enough? I'm not saying it's wrong, I'm just trying to see what we are gaining by doing it this way.

A simpler solution that i was imagining was that we would use the HealthProbe information during the creation of the `dns.MultiClusterGatewayTarget`, that is passed into `setEndpoints`, rather than building all the endpoints and then removing them again. Essentially remove "unhealthy" cluster/gateway/addresses before it even gets to `setEndpoints`. That way health stuff doesn't need to worry about the record structure and CNAMES  and all the rest, just that the service is accessible on the expected clusters via the expected host.

philbrookes:
Verified this is all working as expected locally.

pkg/controllers/dnspolicy/dns_helper.go
philbrookes: Currently main does create too many probes, but if you run on this branch you should only see probes for A records - or CNAMEs that point outside the zone. For example, I currently have a dns record like this:
```
endpoints:
  - dnsName: 24osuu.lb-4ej5le.myapp.pb.hcpapps.net
    recordTTL: 60
    recordType: A
    targets:
    - 172.31.202.0
  - dnsName: 2c71gf.lb-4ej5le.myapp.pb.hcpapps.net
    recordTTL: 60
    recordType: A
    targets:
    - 172.31.200.0
  - dnsName: default.lb-4ej5le.myapp.pb.hcpapps.net
    providerSpecific:
    - name: weight
      value: "120"
    recordTTL: 60
    recordType: CNAME
    setIdentifier: 24osuu.lb-4ej5le.myapp.pb.hcpapps.net
    targets:
    - 24osuu.lb-4ej5le.myapp.pb.hcpapps.net
  - dnsName: default.lb-4ej5le.myapp.pb.hcpapps.net
    providerSpecific:
    - name: weight
      value: "120"
    recordTTL: 60
    recordType: CNAME
    setIdentifier: 2c71gf.lb-4ej5le.myapp.pb.hcpapps.net
    targets:
    - 2c71gf.lb-4ej5le.myapp.pb.hcpapps.net
  - dnsName: default.lb-4ej5le.myapp.pb.hcpapps.net
    providerSpecific:
    - name: weight
      value: "120"
    recordTTL: 60
    recordType: CNAME
    setIdentifier: lrnse3.lb-4ej5le.myapp.pb.hcpapps.net
    targets:
    - lrnse3.lb-4ej5le.myapp.pb.hcpapps.net
  - dnsName: lb-4ej5le.myapp.pb.hcpapps.net
    providerSpecific:
    - name: geo-code
      value: '*'
    recordTTL: 300
    recordType: CNAME
    setIdentifier: default
    targets:
    - default.lb-4ej5le.myapp.pb.hcpapps.net
  - dnsName: lrnse3.lb-4ej5le.myapp.pb.hcpapps.net
    recordTTL: 60
    recordType: A
    targets:
    - 172.31.201.0
  - dnsName: myapp.pb.hcpapps.net
    recordTTL: 300
    recordType: CNAME
    targets:
    - lb-4ej5le.myapp.pb.hcpapps.net
```
And the following health probes:
```
NAMESPACE                NAME                        HEALTHY   LAST CHECKED
multi-cluster-gateways   172.31.200.0-prod-web-api   true      31s
multi-cluster-gateways   172.31.201.0-prod-web-api   true      26s
multi-cluster-gateways   172.31.202.0-prod-web-api   true      21s
```

mikenairn: Yeah OK, i see. So we are no longer creating the ones for all the CNAME endpoints which is good as i couldn't see why we needed to.

Since that is the case, you could probably have done this more simply as described here https://github.com/Kuadrant/multicluster-gateway-controller/pull/412#discussion_r1315571223. If you think of it as removing unhealthy gateway addresses rather than endpoints in the DNSRecord. From what i can gather from the changes in   `pkg/controllers/dnspolicy/dnspolicy_healthchecks.go` this would fit in better anyway (process the gateway addresses not the DNSRecord). You wouldn't need to figure out what CNAMES you need to remove since it would just never have added it in the first place if the addresses are all empty for a cluster/gateway.

Up to you though, just a suggestion.

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
```diff
@@ -101,58 +103,48 @@ func (r *DNSPolicyReconciler) expectedProbesForGateway(ctx context.Context, gw c
...
-			if errors.IsNotFound(err) {
-				continue
-			} else {
-				return nil, err
-			}
+		matches := re.FindAllString(address.Value, -1)
+		if len(matches) != 1 {
+			log.V(3).Info("Found more or less than 1 ip address")
+			continue
 		}
```
mikenairn, line 119: Why use a regexp here instead of processing the type of Address (IPAddress, kuadrant.io/MultiClusterIPAddress etc..). I assume this doesn't work for Hostname addresses?

pkg/controllers/dnspolicy/probe_eventmapper.go
```diff
@@ -0,0 +1,60 @@
...
+func (p *ProbeEventMapper) mapToPolicyRequest(obj client.Object, _ string, _ common.PolicyRefsConfig) []reconcile.Request {
+	logger := p.Logger.V(3).WithValues("object", client.ObjectKeyFromObject(obj))
+	probe, ok := obj.(*v1alpha1.DNSHealthCheckProbe)
+	if !ok {
+		logger.Info("mapToPolicyRequest:", "error", fmt.Sprintf("%T is not a *v1alpha1.DNSHealthCheckProbe", obj))
+		return []reconcile.Request{}
+	}
+
+	allDNSPolicy := &v1alpha1.DNSPolicyList{}
+	err := p.client.List(context.TODO(), allDNSPolicy)
```
mikenairn, line 39: You shouldn't need to list all the policies, there should be enough info in the annotation to build a request for the correct policy without this additional API call.

laurafitzgerald: Gotcha. This should be simplified now in the latest commit. 👍

laurafitzgerald: ah we actually don't add the same annotations to the probes. we are adding labels. I think it would be better for us to add the annotations and use those. taking a look now.

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
laurafitzgerald: The idea is strip out he cluster prefix from the Value. The value (ip address or hostname) is used in the probe. you're right in saying that this wont work in it's current implementation for type hostname. My next suggestion would be to split out using the `/` given this line https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/controllers/gateway/gateway_controller.go#L216 wdyt?

laurafitzgerald: I went a head an implemented the above. Let me know if the preference is to change this but both should be handled now.

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
```diff
@@ -101,40 +101,44 @@ func (r *DNSPolicyReconciler) expectedProbesForGateway(ctx context.Context, gw c
...
 			port = &defaultPort
 		}
 
-		matches := re.FindAllString(address.Value, -1)
-		if len(matches) != 1 {
-			log.V(3).Info("Found more or less than 1 ip address")
-			continue
+		matches := strings.Split(address.Value, "/")
+		if len(matches) != 2 {
+			return nil, errors.New(fmt.Sprintf("unable to extract address from address %s ", address.Value))
```
laurafitzgerald, line 115: Are we happy to error here rather than log. The log seemed quiet if there was to be a problem with the formatting of the adress.Value cc @philbrookes ?

philbrookes: I wonder was this a log rather than an error, so that any good values still get processed?

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -209,7 +209,7 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 			} else if *address.Type == gatewayv1beta1.HostnameAddressType {
 				addressType = MultiClusterHostnameAddressType
 			} else {
-				break // ignore address type. Unsupported for multi cluster gateway
+				continue // ignore address type gatewayv1beta1.NamedAddressType. Unsupported for multi cluster gateway
```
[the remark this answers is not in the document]
philbrookes: this makes sense to me.

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
```diff
@@ -99,60 +100,54 @@ func (r *DNSPolicyReconciler) expectedProbesForGateway(ctx context.Context, gw c
...
-						FailureThreshold:         dnsPolicy.Spec.HealthCheck.FailureThreshold,
-						ExpectedResponses:        dnsPolicy.Spec.HealthCheck.ExpectedResponses,
-						AllowInsecureCertificate: dnsPolicy.Spec.HealthCheck.AllowInsecureCertificates,
-					},
-				}
-				healthChecks = append(healthChecks, withDNSRecord(dnsRecord, healthCheck))
+			// handle protocol being nil
+			var protocol string
+			if dnsPolicy.Spec.HealthCheck.Protocol == nil {
+				protocol = string(v1alpha1.HttpProtocol)
```
laurafitzgerald, line 124: Does this make sense as a default? if the value isn't set we end up with a panic. @philbrookes

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
laurafitzgerald: I put this back to a log. 👍  removed the return of the error as it's never used.

pkg/controllers/dnspolicy/dns_helper.go
laurafitzgerald: Resolving this here as it will potentially be addressed in a seperate pr.

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
philbrookes: hmm, I think HTTPS probably makes more sense for the default  value here, I imagine that is much more common.

laurafitzgerald: Cool. I'll update it. 👍

pkg/apis/v1alpha1/dnsrecord_types.go
```diff
@@ -107,7 +107,7 @@ type DNSRecordSpec struct {
 	ManagedZoneRef *ManagedZoneReference `json:"managedZone,omitempty"`
 	// +kubebuilder:validation:MinItems=1
 	// +optional
-	Endpoints []*Endpoint `json:"endpoints"`
+	Endpoints []*Endpoint `json:"endpoints,omitempty"`
```
mikenairn, line 110: Out of interest, why was this needed? I have no issue with it, other than it probably doesn't make sense with that validations checking for 1 item. We either allow it to be empty or we don't.

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
mikenairn: Could the protocol not be determined by the listener?

mikenairn:
I think these changes look fine. One thing that this changes that is not really pointed out is that we are creating probes for gateway addresses now rather that DNSRecord.Endpoints. This is a good change IMO, and makes these health probes for a gateway instead of a DNSRecord allowing them to potentially be used by anything that wants to know about the status of the services exposed by a gateway and not limiting their use to DNS creation related use cases. 

Since we have made this change, i do believe changing how we are manipulating the dns records that we are creating could be done differently to avoid health checks becoming involved in the manipulation of the DNSRecord itself and I'd probably create a follow on issue to change that if we don't intend to address it here re. https://github.com/Kuadrant/multicluster-gateway-controller/pull/412/files#r1314733762

pkg/apis/v1alpha1/dnsrecord_types.go
laurafitzgerald: I faced a nil dereference at one point when this was empty. That was at a point where I hadn't implemented the use case of including all endpoints when the checks resulted in there being no endpoints present. So I think it's unlikely that we'd hit that now but i kept in in case that case came up again in order to prevent any crashes in the code.

laurafitzgerald:
> I think these changes look fine. One thing that this changes that is not really pointed out is that we are creating probes for gateway addresses now rather that DNSRecord.Endpoints. This is a good change IMO, and makes these health probes for a gateway instead of a DNSRecord allowing them to potentially be used by anything that wants to know about the status of the services exposed by a gateway and not limiting their use to DNS creation related use cases.
> 
> Since we have made this change, i do believe changing how we are manipulating the dns records that we are creating could be done differently to avoid health checks becoming involved in the manipulation of the DNSRecord itself and I'd probably create a follow on issue to change that if we don't intend to address it here re. https://github.com/Kuadrant/multicluster-gateway-controller/pull/412/files#r1314733762

thanks @mikenairn i'm currently looking at a potential change to address this. If you're okay with it I'll address it in a follow on pr so we can get this one merged. Would we need an issue for that?

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
laurafitzgerald: Yes, true. I updated that and also for the default port, we can take that from the listener as well if that makes sense?

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
```diff
@@ -94,70 +88,67 @@ func (r *DNSPolicyReconciler) deleteUnexpectedGatewayProbes(ctx context.Context,
...
-	}
+	for _, address := range gw.Status.Addresses {
+		matches := strings.Split(address.Value, "/")
+		if len(matches) != 2 {
+			log.V(3).Info(fmt.Sprintf("%s cannot be processed. expected <clsutername>/<hostname/ipaddress>", address.Value))
+			return nil
+		}
 
-	for _, listener := range gw.Spec.Listeners {
+		interval := metav1.Duration{Duration: 60 * time.Second}
```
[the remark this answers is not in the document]
sergioifg94: It looks good, but I'd just move that block that defines `interval` outside of the loop, otherwise we're redefining the variable in every iteration. Wouldn't be surprised that the compiler is smart enough to optimise it so it doesn't make any difference but just in case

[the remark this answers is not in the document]
laurafitzgerald: Done cc @mikenairn

pkg/apis/v1alpha1/dnsrecord_types.go
mikenairn: We should probably remove https://github.com/Kuadrant/multicluster-gateway-controller/pull/412/files#diff-2c474b44d62073e226cd94ece69ad23db20801a324cec059d85f1a9128acebd0L108. Can be done later.

PR #415, opened 2023-08-09
Containerise Dependencies for Quickstart Script

### LINKED ISSUE #387: Package dependencies for quickstart via Docker image
**What**
Package the dependencies required to run the quickstart script into a Dockerfile so that the user doesn't need to complete a large set of pre-requisite steps.

Once dependencies are managed by the Dockerfile remove the reference to pre-requisites handled by the Dockerfile from https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/docs/how-to/ocm-control-plane-walkthrough.md#requirements and https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/docs/getting-started.md#pre-requisites

**Alternative**
Scriptify the `make dependencies` command so that the script will handle installation of dependencies

**How**
NOTE: the how is not currently solved but here is an example that was tried.

Sample Dockerfile
```
FROM registry.access.redhat.com/ubi8/ubi:8.1

WORKDIR /app

RUN dnf install -y curl tar

RUN curl -Lo /usr/local/bin/kind https://github.com/kubernetes-sigs/kind/releases/download/v0.11.0/kind-linux-amd64 \
    && chmod +x /usr/local/bin/kind
```
Usage example
```
mkdir -p /tmp/mgc_bin
docker run --rm -v /tmp/mgc_bin:/usr/local/bin mgc-quickstart:test

chmod +x /tmp/mgc_bin/kind
export KIND_BIN=/tmp/kind_bin/kind
```

Error output is 
```[exec /usr/local/bin/kind: exec format error
/dev/stdin: line 5: /tmp/kind_bin/kind: cannot execute binary file](docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "/usr/local/bin/kind": stat /usr/local/bin/kind: no such file or directory: unknown.)```

OR 
```chmod: /tmp/mgc_bin/kind: No such file or directory
/dev/stdin: line 5: /tmp/kind_bin/kind: cannot execute binary file```

### LINKED ISSUE #418: OCM Control Plane Walkthrough not working
While running through the [OCM Control Plane walkthrough](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/docs/how-to/ocm-control-plane-walkthrough.md) I encountered an issue with the deployment of a gateway on the second cluster where no IP address was assigned, see below:
```
kubectl get gateways -A
Set kubectl context to "kind-mgc-workload-1"
NAMESPACE                         NAME       CLASS   ADDRESS   PROGRAMMED   AGE
kuadrant-multi-cluster-gateways   prod-web   istio                          5m20s
```
Looking at the gateway resource it seems that the address cannot be assigned:
```
kubectl get gateway prod-web -n kuadrant-multi-cluster-gateways -o yaml
...
  - lastTransitionTime: "2023-08-10T09:48:29Z"
    message: 'Assigned to service(s) prod-web-istio.kuadrant-multi-cluster-gateways.svc.cluster.local:443,
      but failed to assign to all requested addresses: address pending for hostname
      "prod-web-istio.kuadrant-multi-cluster-gateways.svc.cluster.local"'
```
Having spoken with others in the team it appears that they are also hitting this issue

**Steps to reproduce**

- Following the [OCM Control Plane walkthrough](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/docs/how-to/ocm-control-plane-walkthrough.md)
- When running through the steps in the [Introduce the second cluster section](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/docs/how-to/ocm-control-plane-walkthrough.md#introducing-the-second-cluster) you'll notice that the deployed gateway does not have an assigned IP address. Based on the walkthrough the IP should be `172.32.201.0 `

### PR DESCRIPTION
**Pull request by eoinfennessy**
/closes #387 
/closes #418

## What

- [x] Adds docker image containing binaries needed for `hack/quickstart-setup.sh`
- [x] `hack/quickstart-setup.sh` changed to use binaries in above image
- [x] Adds GitHub action to build and push tools image to quay.io
- [x] Update docs: change requirements in `ocm-control-plane-walkthrough.md` and `getting-started.md`
- [x] Fixes bug in quickstart script where no IPAddressPool was created for MetalLB in spoke clusters

## Verification

EDIT: Local verification of the quickstart script is not currently possible because changes have been made that are necessary for merging. Ping me if you would like to do additional verification so I can push up a temporary change.

Run `hack/quickstart-setup.sh` locally and verify a successful install. Note: The `docker build` command currently runs in this script and should be removed before merging or when the tools image is available on quay.io.

Verify that `make local-setup` continues to work as expected.

### CODE REVIEW

**Comment by maleck13** [comment]
@eoinfennessy do you want me to give this try locally?

**Comment by eoinfennessy** [comment]
> @eoinfennessy do you want me to give this try locally?

Probably not much need to try it out locally yet... I'll try to get a few more of the binaries working first.

File: hack/quickstart-setup.sh [hunk 1]
```diff
@@ -16,18 +16,24 @@
...
+export TOOLS_IMAGE=quay.io/kuadrant/mgc-tools:latest
+export TMP_DIR=/tmp/mgc
 
 operatorSdkInDocker() {
   docker run -u $UID -v "${TMP_DIR}:${TMP_DIR}:z" --network mgc -e KUBECONFIG="${TMP_DIR}/kubeconfig" --rm ${TOOLS_IMAGE} -c "operator-sdk $*"
 }
 
+yqInDocker() {
+  docker run -u $UID -v "${TMP_DIR}:${TMP_DIR}:z" --rm ${TOOLS_IMAGE} -c "yq $1 '$2' $3"
+}
```
eoinfennessy, line 28: @mikenairn @laurafitzgerald This is far from ideal, but works for the two lines that require `yq` that are used in this script (these commands are in the `kindGenExternalKubeconfig` function in `hack/.kindUtils`). The way args are handled here is because of some issues with quotes used in the `yq` commands being lost when passed to this function. I tried to refactor the commands to work for the binary and a version of this function that used `$*`, but couldn't find a way that worked for both. Any ideas how this could be improved?

mikenairn: Id try and changing the entrypoint to the container i.e. `docker run --rm -it -u 1000 -v /tmp/mgc:/tmp/mgc:z --network mgc -e KUBECONFIG=/tmp/mgc/kubeconfig --entrypoint=yq quay.io/kuadrant/mgc-tools:latest` 

``` #!/bin/bash

TOOLS_IMAGE=quay.io/kuadrant/mgc-tools:latest

dockerBinCmd() {
  echo "docker run --rm -it -u $UID -v /tmp/mgc:/tmp/mgc:z --network mgc -e KUBECONFIG=/tmp/mgc/kubeconfig --entrypoint=$1 $TOOLS_IMAGE"
}

KUSTOMIZE_BIN=$(dockerBinCmd "kustomize")
YQ_BIN=$(dockerBinCmd "yq")
OPERATOR_SDK_BIN=$(dockerBinCmd "operator-sdk")

${KUSTOMIZE_BIN} version
${YQ_BIN} --version
${OPERATOR_SDK_BIN} version
${OPERATOR_SDK_BIN} olm status 
```

laurafitzgerald:
I haven't verified but docs updates look great. 👍

.github/workflows/tools-image.yaml
```diff
@@ -0,0 +1,63 @@
...
+      - main
+      - "release-*"
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+"
+
+env:
+  IMG_REGISTRY_HOST: quay.io
+  IMG_REGISTRY_ORG: kuadrant
+  IMG_REGISTRY_REPO: mgc-tools
+  MAIN_BRANCH_NAME: main
```
mikenairn, line 15: Is this repo in place?

Dockerfile.tools
```diff
@@ -0,0 +1,25 @@
...
+
+COPY hack/make/dependencies.make hack/make/dependencies.make
+COPY Makefile Makefile
+
+# run make
+RUN make kustomize operator-sdk kind helm yq clusteradm LOCALBIN=/usr/local/bin
+
+ENV HOME /home
+
+ENTRYPOINT [ "/bin/bash" ]
```
mikenairn, line 25: We can likely clean this up (remove yum cache, Makefile etc..) to make the image smaller, but ... probably not that big a deal either. Can be looked into later.

hack/.kindUtils
```diff
@@ -56,10 +56,11 @@ nodes:
...
 EOF
-mkdir -p ./tmp/kubeconfigs
-${KIND_BIN} get kubeconfig --name ${cluster} > ./tmp/kubeconfigs/${cluster}.kubeconfig
-${KIND_BIN} export kubeconfig --name ${cluster} --kubeconfig ./tmp/kubeconfigs/internal/${cluster}.kubeconfig --internal
-kindGenExternalKubeconfig
+
+  mkdir -p ${TMP_DIR}/kubeconfigs
+  ${KIND_BIN} get kubeconfig --name ${cluster} > ${TMP_DIR}/kubeconfigs/${cluster}.kubeconfig
+  ${KIND_BIN} export kubeconfig --name ${cluster} --kubeconfig ${TMP_DIR}/kubeconfigs/internal/${cluster}.kubeconfig --internal
+  kindGenExternalKubeconfig
```
mikenairn, line 63: Would be interested to know where these kubeconfigs are being used currently, and if we even need them all. Ideally we'd just export the combined internal kubeconfig in the one place, the one you are creating in the quick start script, and here ideally.

hack/quickstart-setup.sh
```diff
@@ -77,10 +89,19 @@ if [[ -z "${MGC_WORKLOAD_CLUSTERS_COUNT}" ]]; then
...
+
+# Make temporary directory for kubeconfig
+mkdir -p ${TMP_DIR}
+
 cleanupKind
 
 kindSetupMGCClusters ${KIND_CLUSTER_CONTROL_PLANE} ${KIND_CLUSTER_WORKLOAD} ${port80} ${port443} ${MGC_WORKLOAD_CLUSTERS_COUNT}
 
+# Export kubeconfig for control plane cluster
+kind export kubeconfig -n ${KIND_CLUSTER_CONTROL_PLANE} --kubeconfig ${TMP_DIR}/kubeconfig --internal
```
mikenairn, line 103: This could be done in `kindCreateCluster` where the other kubeconfigs are generated, would work in local-setup and quick start then if we decided to just always use the docker tools image.

mikenairn:
This seems fine, but i haven't tried it. Do you want to undo the changes in the quick start script that need removed before it can be merged?

.github/workflows/tools-image.yaml
eoinfennessy: No, not yet. Does it need to be created manually before this action runs?

jasonmadigan:
@eoinfennessy I'm guessing the `mgc-tools` image is stiill private?

```bash
Unable to find image 'quay.io/kuadrant/mgc-tools:latest' locally
docker: Error response from daemon: manifest for quay.io/kuadrant/mgc-tools:latest not found: manifest unknown: manifest unknown.
```

sergioifg94:
Verified it locally. Just as a note, I had some issues building the tools image due to my local docker install so it might be good to add to the docs the minimum docker version required, as well as the fact that [Buildx](https://github.com/docker/buildx) is required to suport heredoc when adding the kubernetes repo

jasonmadigan:
Verified HEAD~1 there and worked good.

> Verified it locally. Just as a note, I had some issues building the tools image due to my local docker install so it might be good to add to the docs the minimum docker version required, as well as the fact that [Buildx](https://github.com/docker/buildx) is required to suport heredoc when adding the kubernetes repo

Yeah, I think `buildx` has been included in engine since Docker 19. Off-topic, we should look to see Podman works OK at some point too.

Ygnas:
> Verified HEAD~1 there and worked good.
> 
> > Verified it locally. Just as a note, I had some issues building the tools image due to my local docker install so it might be good to add to the docs the minimum docker version required, as well as the fact that [Buildx](https://github.com/docker/buildx) is required to suport heredoc when adding the kubernetes repo
> 
> Yeah, I think `buildx` has been included in engine since Docker 19. Off-topic, we should look to see Podman works OK at some point too.

It does not work with **Podman**, as there are some issues with heredoc.

sergioifg94:
If heredoc is causing trouble, maybe we can have a base image that has the repository added with a simple `COPY`, and build the tools image on top of it?

mikenairn:
> Verified it locally. Just as a note, I had some issues building the tools image due to my local docker install so it might be good to add to the docs the minimum docker version required, as well as the fact that [Buildx](https://github.com/docker/buildx) is required to suport heredoc when adding the kubernetes repo

People really shouldn't be needing to build this image themselves, it's just because nobody has pushed it and the GH workflow isn't in place yet. So i don't really see this is a big issue. That said, kubectl isn't even being used inside the container from what i can tell, so you could just remove that whole bit from the Dockerfile.

> It does not work with **Podman**, as there are some issues with heredoc. Not using it in Dockerfile works fine.

For the same reasons i wouldn't be too concerned about being able to build the image with podman, but it would be interesting to know if you could swap out docker for podman when running the tools via the container i.e `podman run --rm -i --entrypoint="kustomize" quay.io/kuadrant/mgc-tools:latest build ....`

@eoinfennessy I built the image as is and pushed it to `quay.io/kuadrant/mgc-tools:latest`, but it's far too big. We will need to reduce the size of the image by quite a bit. We could look into a multi stage build maybe and copy the binaries from a builder to the final image.

hack/.kindUtils
eoinfennessy: These are only being used in the submariner deployment and PoC docs. I think it should now be possible to remove the creation of these files and rework `kindGenExternalKubeconfig` so that it does not need to use these additional files.

eoinfennessy:
> That said, kubectl isn't even being used inside the container from what i can tell, so you could just remove that whole bit from the Dockerfile.
Yeah, we're not using kubectl in the container (unless any of the other tools depend on it for some reason)? I'll remove this now.

> @eoinfennessy I built the image as is and pushed it to `quay.io/kuadrant/mgc-tools:latest`, but it's far too big. We will need to reduce the size of the image by quite a bit. We could look into a multi stage build maybe and copy the binaries from a builder to the final image.
Will look into this. Is there a lightweight base image you'd recommend for the second stage?

jasonmadigan:
> > That said, kubectl isn't even being used inside the container from what i can tell, so you could just remove that whole bit from the Dockerfile.
> 
> Yeah, we're not using kubectl in the container (unless any of the other tools depend on it for some reason)? I'll remove this now.
> 
> > @eoinfennessy I built the image as is and pushed it to `quay.io/kuadrant/mgc-tools:latest`, but it's far too big. We will need to reduce the size of the image by quite a bit. We could look into a multi stage build maybe and copy the binaries from a builder to the final image.
> 
> Will look into this. Is there a lightweight base image you'd recommend for the second stage?

Small change, but could switch to ubi-minimal: https://catalog.redhat.com/software/containers/ubi8/ubi-minimal/5c359a62bed8bd75a2c3fba8

70MB->~35MB.

Multi-stage build like Mike suggested should shave a significant bit off

eoinfennessy:
> Small change, but could switch to ubi-minimal: https://catalog.redhat.com/software/containers/ubi8/ubi-minimal/5c359a62bed8bd75a2c3fba8

Thanks @jasonmadigan!

Latest `Dockerfile.tools` uses `ubi8-minimal`. It may be possible in future to further reduce the size of this tools image by using a `ubi8-micro` image, but I needed to use `microdnf` to install git, as copying over the binary from the build stage caused an error in the script that I haven't been able to resolve:
```
kustomize --load-restrictor LoadRestrictionsNone build github.com/kuadrant/multicluster-gateway-controller.git/config/quick-start/control-cluster
Error: git cmd = '/usr/bin/git fetch --depth=1 origin HEAD': exit status 128
```

Dockerfile.tools
```diff
@@ -0,0 +1,25 @@
...
+RUN make kustomize operator-sdk helm yq clusteradm LOCALBIN=/usr/local/bin
+
+
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.8 AS tools
+
+RUN microdnf install -y git && microdnf clean all
+
+COPY --from=build /usr/local/bin /usr/local/bin
+COPY --from=build /usr/bin/clusteradm /usr/bin/clusteradm
+COPY --from=build /usr/bin/clusteradm /usr/bin/clusteradm
```
mikenairn, line 21: Is this meant to be here twice? Why is it not in `/usr/local/bin`?

eoinfennessy: Whoops! Will remove the duplicate now. Not sure why `make` installed `clusteradm` to this directory. Do you think I should copy it into the `/usr/local/bin` directory?

eoinfennessy: Seems the `make` target for `clusteradm` doesn't use the `LOCALBIN` variable for some reason.
```
.PHONY: clusteradm
clusteradm: $(CLUSTERADM)
$(CLUSTERADM):
	test -s $(CLUSTERADM)|| curl -sL https://raw.githubusercontent.com/open-cluster-management-io/clusteradm/main/install.sh | INSTALL_DIR=bin bash -s -- $(CLUSTERADM_VERSION)
```

mikenairn: OK, odd, not something we need to fix here i guess. Remove the dup line and i think we can just merge it and see how it goes.

Ygnas (approved):
I have verified all the steps and can confirm that everything is functioning smoothly on my end. Additionally, issue #418 appears to have been resolved, second cluster IP address was assigned.

Issue #413, opened 2023-08-08
Reevaluate local setup functions


maksymvavilov:
Follow up on #256 

There is a number of functions, invoked during the local setup and quickstart, with decisions made a while ago. Some of those decisions might not be optimal or might not make sense anymore. 
It might be worth reevaluating what and how we do. For example, the deployment of Redis and Istio as part of the OCM hub installation or `clusteradm accept` calls. If some of the components that are being deployed are not used any longer we should consider removing them (like the `deployAgentSecret` function). 
Additionally, it was suggested to revisit the deployment logic to allow the installation to continue once interrupted - running the install script brings components to a healthy state or creates them if they are missing. While currently, it is a mixture of that approach and "cleanup first and then install again"

eoinfennessy:
A bug somewhat related to this: No clusters are deployed with OCM if `make local-setup` is run without the `OCM_SINGLE` variable set.

Replicate this by running `make local-setup` - No `managedClusters` will be present after the setup.

philbrookes:
This issue is stale because it has been open for 60 days with no activity.

philbrookes:
This issue was closed because it has been inactive for 30 days since being marked as stale.

philbrookes:
This issue is stale because it has been open for 60 days with no activity.

philbrookes:
This issue was closed because it has been inactive for 30 days since being marked as stale.

philbrookes:
This issue is stale because it has been open for 60 days with no activity.

philbrookes:
This issue was closed because it has been inactive for 30 days since being marked as stale.

PR #416, opened 2023-08-10
waiting for deployments

laurafitzgerald:
Small pr to add wait for deployments


makslion:
/lgtm

Looks good and after a bit of digging around don't see reasons/ways to make the `local-setup` path wait for those. I'll tag the #413 and [link](https://kubernetes.slack.com/archives/C05J0D0V525/p1691661539443489) the slack thread here so the person revisiting this logic would have an easier time finding out the reasoning behind the change.

laurafitzgerald:
> /lgtm
> 
> Looks good and after a bit of digging around don't see reasons/ways to make the `local-setup` path wait for those. I'll tag the #413 and [link](https://kubernetes.slack.com/archives/C05J0D0V525/p1691661539443489) the slack thread here so the person revisiting this logic would have an easier time finding out the reasoning behind the change.

The wait is based on the components that are called by kustomize. so the kustomize we use is the quick-start ones. so there shouldn't be any variance on the waits between the two.

makslion (approved):

maleck13 (approved):

PR #417, opened 2023-08-10
Add mgc metric dashboard

### LINKED ISSUE #308: Create an example Grafana dashboard that shows SRE focused metrics for the controller
*What*

Building on #305, design a set of dashboard panels that show important metrics for monitoring the multicluster-gateway-controller component from an SRE viewpoint.
This should focus on the [4 Golden Signals](https://sre.google/sre-book/monitoring-distributed-systems/#xref_monitoring_golden-signals).
To gather the required metrics, additional ServiceMonitors or federated metrics may be required.

Some example metrics & queries to use can be seen in the kcp-glbc dashboard [here](https://gist.github.com/david-martin/3332a5d89717bebdfe8ce1bb887dafd8).

In particular:
- kube_pod_status_phase
- kube_pod_container_resource_limits
- kube_pod_container_resource_requests
- container_cpu_usage_seconds_total
- kube_pod_container_status_restarts_total
- container_memory_rss
- go_goroutines
- workqueue_adds_total (if using workqueue)
- workqueue_depth
- workqueue_queue_duration_seconds_bucket

Ygnas:
Closes #308 

## What

- [x] Adds new dashboard for mgc metrics

## Verification
Run `make local-setup OCM_SINGLE=true METRICS_FEDERATION=true MGC_WORKLOAD_CLUSTERS_COUNT=1` and `make build-controller kind-load-controller deploy-controller METRICS=true`
The dashboard can be found on the Default dashboard page named: `MGC SRE Dashboard`

image

## Any suggestions welcome:
- Are there any specific visualizations that should be prioritized on the dashboard? (e.g., graphs, tables, gauges, heatmaps)
- Are there additional metrics or data sources that should be included in this dashboard?
- Would it be beneficial to incorporate more variables to allow users to filter and customize the dashboard's data dynamically?
- Are there specific variables that should be used?

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -33,4 +33,4 @@ resources:
 - ../rbac
 - ../manager
 patches:
-- path: manager_auth_proxy_patch.yaml
+- path: manager_metrics_patch.yaml
--- a/config/default/manager_metrics_patch.yaml
+++ b/config/default/manager_metrics_patch.yaml
@@ -1,0 +1,21 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: controller-manager
+  namespace: system
+spec:
+  template:
+    spec:
+      containers:
+      - name: manager
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+              - "ALL"
+        args:
+        - "--metrics-bind-address=0.0.0.0:8080"
+        - "--leader-elect"
+        ports:
+        - containerPort: 8080
+          name: metrics
--- a/config/prometheus-for-federation/grafana-dashboard-mgc-metrics.json
+++ b/config/prometheus-for-federation/grafana-dashboard-mgc-metrics.json
@@ -24,8 +24,8 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "id": 19,
-  "iteration": 1691663562703,
+  "id": 25,
+  "iteration": 1692087230577,
   "links": [],
   "liveNow": false,
   "panels": [
@@ -272,7 +272,7 @@
           },
           "editorMode": "code",
           "exemplar": false,
-          "expr": "sum by (pod) (rate(kube_pod_container_status_restarts_total[30m]) > 0 )",
+          "expr": "sum by (pod) (rate(kube_pod_container_status_restarts_total{namespace=~\"$namespace\"}[30m]) > 0 )",
           "format": "time_series",
           "instant": true,
           "intervalFactor": 2,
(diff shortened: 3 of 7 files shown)


david-martin:
> * Are there any specific visualizations that should be prioritized on the dashboard? (e.g., graphs, tables, gauges, heatmaps)
> * Are there additional metrics or data sources that should be included in this dashboard?
> * Are there specific variables that should be used?

In general, looks like there's a good spread of visualisations already included.
I can't think of anything offhand at this time.

> * Would it be beneficial to incorporate more variables to allow users to filter and customize the dashboard's data dynamically?

I think less variables would be better initially.
I'm wondering about the usefulness of having the `Namespace` dropdown on an MGC focused dashboard.
By default at the moment, 'All' is selected, which shows aggregated stats (like Pod Statuses) across all namespaces, which is more broad than I would expect on an MGC focused dashboard.
I think it would be better to have the Namespace dropdown auto populated with the `multicluster-gateway-controller-system` namespace (but leave the dropdown in case it's installed in some other namespace).
e.g. 
image


Some other feedback when reviewing the dashboard.

*Missing metrics*

The Goroutines panel, which uses the metric `go_goroutines`, is showing no data for me for the `multicluster-gateway-controller-system` namespace.
image
It looks like metrics from the mgc controller are not being scraped (or being federated back to thanos).
Some of the panels use metrics from other scrape targets/jobs (like apisever for workqueue_adds_total, kubelet for container_cpu_usage_seconds_total, or kube-state-metrics for kube_pod_status_phase) which would explain why they show data for the mgc controller.

I thought this was because the `METRICS=true` flag has to be set when starting the controller to ensure the ServiceMonitor is created. I didn't do this initially, but even after doing this the metrics weren't there.
This may need some debugging of the prometheus job for the mgc controller.


*Doubling up of legend labels*

On the cpu & memory panels, the requests and limits labels are doubling up. Any idea what's causing this?
image

*Legend labels for Work Queue panels*

It would be helpful to have labels shown below the work queue panels to more easily get an overview of what the panel is showing
image

*Work Queue panels not filtering by namespace (and missing metrics)*

Related to the missing metrics problem, and may be partially solved when that is fixed.
Both work queue panels are not filtering by namespace, and show workqueue stats for lots of unrelated controllers.
It would be great if this was filtered by namespace (and thus filtered to the mgc controller) to get a better idea of what CRD queue times are like.

To see all the available metrics that are available from the mgc controller, I used port-forward and curl
e.g.
```
kubectl port-forward -n multicluster-gateway-controller-system mgc-controller-manager-84c947d9b9-rktjv :8080
curl localhost:51560/metrics
```
i.e. this list https://gist.github.com/david-martin/c009d564a3f3c1eaa94f1641e148c972

david-martin:
Re-reviewed.

Dashboard is looking good.

/lgtm
/approve

image

image

The CPU usage is showing constantly at or above the limits (so I suspect it is getting throttled).
A separate issue will be opened by @Ygnas to investigate that i.e. is that expected usage? Should the limit be raised?

PR #431, opened 2023-08-15
more specific removal details

philbrookes:


laurafitzgerald:
/lgtm
document describes the problem and the solution clearly. 👍

PR #433, opened 2023-08-17
GH-324 pollicyaffected condition for gateway

makslion:
## What
Reporting the state of the DNSPolicy to the Gateway affected by that policy 

## Verification 
1. Get the MGC running 
2. Create a DNSPolicy 
```
kubectl apply -f - <<EOF
kind: DNSPolicy
metadata:
  name: prod-web
spec:
  targetRef:
    name: prod-web
    kind: Gateway
  healthCheck:
    endpoint: /
    expectedResponses:
      - 200
      - 201
      - 301
    failureThreshold: 5
    port: 443
    protocol: https
EOF
```
3. Watch conditions on the gateway getting updated.


makslion:
/hold 
I'd like to ensure that this is what we want and how we want. 
The "what" part: 
1. We have an annotation on the Gateway referencing policy. The conditions block looks like a glorified annotation with an error message attached in case something goes wrong. In any case, it might be worth discussing with Kuadrant people since they are the ones to [implement](https://github.com/Kuadrant/kuadrant-operator/blob/1e12d4c61938b9cacd415dd32d7c0b43014fea85/pkg/reconcilers/targetref_reconciler.go#L206-L232) an annotation. 
2. There is still a discussion on the Gateway side about how this should be done. Since all this is still considered to be "experimental" and aimed to land after the GA should we wait for them to [deploy](https://github.com/kubernetes-sigs/gateway-api/issues/738) conditions for the DNSPolicy itself?

pkg/_internal/conditions/conditions.go
```diff
@@ -1,5 +1,13 @@
 package conditions
 
+type ConditionType string
+type ConditionReason string
+
 const (
-	ConditionTypeReady string = "Ready"
+	ConditionTypeReady ConditionType = "Ready"
+
+	DNSPolicyAffected         ConditionType   = "DNSPolicyAffected"
```
maleck13, line 9: hmm I am wondering if this would be better in the dnspolicy package?

pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -225,6 +236,49 @@ func (r *DNSPolicyReconciler) readyCondition(targetNetworkObjectectKind string,
 	return cond
 }
 
+func buildGatewayCondition(reason conditions.ConditionReason, err error) *metav1.Condition {
```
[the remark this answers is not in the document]
maleck13: or we could make it a bit more generic and add it to the conditions package 
```
func buildPolicyAffectedCondition(conditionType conditions.ConditionType, reason conditions.ConditionReason, err error) *metav1.Condition {
	condition := &metav1.Condition{
		Type:    string(conditionType),
		Status:  metav1.ConditionTrue,
		Reason:  string(reason),
		Message: fmt.Sprintf("%s policy affected", conditionType),
	}

	if err != nil {
		condition.Status = metav1.ConditionFalse
		condition.Message = err.Error()
	}

	return condition
}
```

pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -225,6 +236,49 @@ func (r *DNSPolicyReconciler) readyCondition(targetNetworkObjectectKind string,
...
+
+	if err != nil {
+		condition.Status = metav1.ConditionFalse
+		condition.Message = err.Error()
+	}
+
+	return condition
+}
+
+func (r *DNSPolicyReconciler) updateGatewayCondition(ctx context.Context, condition *metav1.Condition, gatewayDiff *reconcilers.GatewayDiff) error {
```
maleck13, line 255: These seem like potentially generic for all policy controllers?

pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -154,19 +157,27 @@ func (r *DNSPolicyReconciler) reconcileResources(ctx context.Context, dnsPolicy
 		return err
 	}
 
-	if err := r.reconcileDNSRecords(ctx, dnsPolicy, gatewayDiffObj); err != nil {
+	if err = r.reconcileDNSRecords(ctx, dnsPolicy, gatewayDiffObj); err != nil {
+		gatewayCondition = buildGatewayCondition(conditions.DNSPolicyReasonInvalid, err)
+		_ = r.updateGatewayCondition(ctx, gatewayCondition, gatewayDiffObj)
```
maleck13, line 162: not sure why we are ignoring the error here?

pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -154,19 +157,27 @@ func (r *DNSPolicyReconciler) reconcileResources(ctx context.Context, dnsPolicy
...
 		return err
 	}
 
-	if err := r.reconcileHealthChecks(ctx, dnsPolicy, gatewayDiffObj); err != nil {
+	if err = r.reconcileHealthChecks(ctx, dnsPolicy, gatewayDiffObj); err != nil {
+		gatewayCondition = buildGatewayCondition(conditions.DNSPolicyReasonInvalid, err)
+		_ = r.updateGatewayCondition(ctx, gatewayCondition, gatewayDiffObj)
```
maleck13, line 168: same here and below. Not much point in returning an error if all we do is ignore it

maleck13:
https://github.com/Kuadrant/multicluster-gateway-controller/pull/436

PR #442, opened 2023-08-22
Run tools image in non-interactive mode

eoinfennessy:
## What

Remove `-i` option from the tools image's `docker run` command to prevent quickstart script exiting early when piping it to bash.

## Verification

Verify that curling the script and piping it to bash successfully creates and configures clusters.

```
curl https://raw.githubusercontent.com/eoinfennessy/multicluster-gateway-controller/fix_quickstart_with_curl/hack/quickstart-setup.sh | bash
```


jasonmadigan (approved):
/lgtm
/approve

PR #449, opened 2023-08-23
Change namespace for MGC in quickstart script

eoinfennessy:
## What

Similar to `make local-setup`, this deploys MGC in the `multicluster-gateway-controller-system` namespace when running the quickstart script.

## Verification

This is not currently in a verifiable state. I've verified that this works locally and following on with the OCM guide works also. Leave a message below if you would like to do further verification and I can push up a temporary change.


Ygnas:
Verified that everything is working locally, OCM guide also works.
/approve

Ygnas (approved):

Issue #427, opened 2023-08-11
How to install the data plane components


maleck13:
**What**

Create a how-to document under the how-to sections of https://docs.kuadrant.io/multicluster-gateway-controller/  that covers how to install the kuadrant data plane components via OCM hub

The usecase here is:  `As a cluster admin , with an existing cluster I wan to use to try kuadrant, how can I install the dataplane components into my spoke cluster. `

steps should be something like:

1) OCM should be installed and you should have an existing OCM hub cluster (we should cover the specifics of that setup in the cp install guide which this should link to) with the control plane components installed (IE follow the CP guide first)
2) Install Istio version x.y as this is the gateway provider and we do not install this (we can provide a ref set of install steps)
3) Install OLM as the service protection components are delivered by OLM currently 
4) Ensure this cluster is added as a managed cluster to your OCM hub (can be same cluster or another)
5) create the `ManagedClusterAddon` resource in the managed cluster namespace to trigger the install of kuadrant into the spoke cluster
```
apiVersion: addon.open-cluster-management.io/v1alpha1
kind: ManagedClusterAddOn
metadata:
 name: kuadrant-addon
 namespace: kind-mgc-workload-1
spec:
 installNamespace: open-cluster-management-agent-addon
```

https://github.com/Kuadrant/multicluster-gateway-controller/tree/main/config/kuadrant/deploy/hub

https://docs.kuadrant.io/multicluster-gateway-controller/docs/how-to/kuadrant-addon-walkthrough/

are resources that should help here

jasonmadigan:
Once in place, look to promote https://docs.kuadrant.io/multicluster-gateway-controller/docs/getting-started/ to the current, top-level getting started guide. And include a jumping off point to reference this data plane install/setup doc.

maleck13:
There will be some pre-requisites needed here. 

1) OLM will need to be installed into the data plane clusters
2) Istio will need to be installed into the data plane clusters

maleck13:
We already do this is in the quick start

jasonmadigan:
@maleck13 given we're so close, I think it might also make sense to also install an RLP, and see it working (since it's only an extra step or two) - turns it into a pretty comprehensive guide, touching on all of the bits in a steel thread?

maleck13:
[...]

PR #452, opened 2023-08-23
data-plane components installation guide

jasonmadigan:
Closes: https://github.com/Kuadrant/multicluster-gateway-controller/issues/427

Added a new Kuadrant data plane components guide, detailing how to install the data-plane components with an existing OCM ManagedCluster.

Use-case: `As a cluster admin , with an existing cluster I want to use to try Kuadrant, how can I install the data-plane components into my workload/ManagedCluster cluster.`

Note: we can likely remove/drop the existing kuadrant-addon-walkthrough doc. We should replace it with a small guide on how to use the data-plane pieces.

### DIFF
--- a/docs/how-to/service-protection-setup.md
+++ b/docs/how-to/service-protection-setup.md
@@ -15,8 +15,7 @@ This walkthrough will show you how to install and setup the Kuadrant Operator in
   * We recommend installing Istio 1.17.0, including Gateway API v0.6.2
   * ```bash
     # On the Hub cluster:
-    kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null || \
-    { kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.6.2" | kubectl apply -f -; }
+    kubectl apply -k "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.6.2"
     ```
   * See also: https://istio.io/v1.17/blog/2022/getting-started-gtwapi/
 
@@ -31,7 +30,7 @@ Alternatively, if you'd like to quickly get started locally, without having to w
 
 To install the Kuadrant Service Protection components into a `ManagedCluster`, target your OCM hub cluster with `kubectl` and run:
 
-`kubectl apply -k "github.com/kuadrant/multicluster-gateway-controller.git/config/service-protection-install-guide" -n <your-managed-cluster>`
+`kubectl apply -k "github.com/kuadrant/multicluster-gateway-controller.git/config/service-protection-install-guide?ref=main" -n <your-managed-cluster>`
 
 The above command will install the `ManagedClusterAddOn` resource needed to install the Kuadrant addon into the specified namespace, and install the Kuadrant data-plane components into the `open-cluster-management-agent-addon` namespace. 
 
@@ -44,7 +43,7 @@ The Kuadrant addon will install:
 For more details, see the Kuadrant components installed by the (kuadrant-operator)[https://github.com/Kuadrant/kuadrant-operator#kuadrant-components]
 
 ### Existing Istio installations and changing the default Istio Operator name
-In the case where you have an existing Istio installation to a cluster you may encounter an issue where the Kuadrant Operator expects Istio's Operator to be named `istiocontrolplane`.
+In the case where you have an existing Istio installation on a cluster, you may encounter an issue where the Kuadrant Operator expects Istio's Operator to be named `istiocontrolplane`.
 
 The `istioctl` command saves the IstioOperator CR that was used to install Istio in a copy of the CR named `installed-state`.
 
@@ -59,7 +58,7 @@ This will propogate down and update the Kuadrant Operator, used by the Kuadrant 
 To verify the Kuadrant OCM addon has installed currently, run:
 
 ```bash
-kubectl wait --timeout=5m -n kuadrant-system deployment/authorino-operator deployment/kuadrant-operator-controller-manager deployment/limitador-operator-controller-manager --for=condition=Available
+kubectl wait --timeout=5m -n kuadrant-system kuadrant/kuadrant-sample --for=condition=Ready
 ```
 
 You should see the namespace `kuadrant-system`, and the following pods come up:
--- a/docs/how-to/service-protection-setup.md
+++ b/docs/how-to/service-protection-setup.md
@@ -1,0 +1,81 @@
+# Installing Kuadrant Service Protection into an existing OCM Managed Cluster
+
+## Introduction
+This walkthrough will show you how to install and setup the Kuadrant Operator into an [OCM](https://open-cluster-management.io/) [Managed Cluster](https://open-cluster-management.io/concepts/managedcluster/).
+
+## Prerequisites
+* Access to an Open Cluster Management (>= v0.11.0) Managed Cluster, which has already been bootstrapped and registered with a hub cluster
+  * We have [a guide](./kuadrant-hub-install.md) which covers this in detail
+  * Also see:
+    * https://open-cluster-management.io/getting-started/quick-start/
+    * https://open-cluster-management.io/concepts/managedcluster/
+* OLM will need to be installed into the ManagedCluster where you want to run the Kuadrant Service Protection components
+  * See https://olm.operatorframework.io/docs/getting-started/
+* Kuadrant uses Istio as a Gateway API provider - this will need to be installed into the data plane clusters
+  * We recommend installing Istio 1.17.0, including Gateway API v0.6.2
+  * ```bash
+    # On the Hub cluster:
+    kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null || \
+    { kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.6.2" | kubectl apply -f -; }
+    ```
+  * See also: https://istio.io/v1.17/blog/2022/getting-started-gtwapi/
+
+
+Alternatively, if you'd like to quickly get started locally, without having to worry to much about the pre-requisites, take a look our [Quickstart Guide](./ocm-control-plane-walkthrough.md). It will get you setup with Kind, OLM, OCM & Kuadrant in a few short steps.
+
+
+## Install the Kuadrant OCM Add-On
+
+
+**Note:** if you've run our [Quickstart Guide](./ocm-control-plane-walkthrough.md), you'll be set to run this command as-is.
+
+To install the Kuadrant Service Protection components into a `ManagedCluster`, target your OCM hub cluster with `kubectl` and run:
+
+`kubectl apply -k "github.com/kuadrant/multicluster-gateway-controller.git/config/service-protection-install-guide" -n <your-managed-cluster>`
+
+The above command will install the `ManagedClusterAddOn` resource needed to install the Kuadrant addon into the specified namespace, and install the Kuadrant data-plane components into the `open-cluster-management-agent-addon` namespace. 
+
+The Kuadrant addon will install:
+
+* the Kuadrant Operator
+* Limitador (and its associated operator)
+* Authorino  (and its associated operator)
+
+For more details, see the Kuadrant components installed by the (kuadrant-operator)[https://github.com/Kuadrant/kuadrant-operator#kuadrant-components]
+
+### Existing Istio installations and changing the default Istio Operator name
+In the case where you have an existing Istio installation to a cluster you may encounter an issue where the Kuadrant Operator expects Istio's Operator to be named `istiocontrolplane`.
+
+The `istioctl` command saves the IstioOperator CR that was used to install Istio in a copy of the CR named `installed-state`.
+
+To let the Kuadrant operator use this existing installation, set the following:
+
+`kubectl annotate managedclusteraddon kuadrant-addon "addon.open-cluster-management.io/values"='{"IstioOperator":"installed-state"}' -n <managed spoke cluster>`
+
+This will propogate down and update the Kuadrant Operator, used by the Kuadrant OCM Addon.
+
+## Verify the Kuadrant addon installation
+
+To verify the Kuadrant OCM addon has installed currently, run:
+
+```bash
+kubectl wait --timeout=5m -n kuadrant-system deployment/authorino-operator deployment/kuadrant-operator-controller-manager deployment/limitador-operator-controller-manager --for=condition=Available
+```
+
+You should see the namespace `kuadrant-system`, and the following pods come up:
+* authorino-*value*
+* authorino-operator-*value*
+* kuadrant-operator-controller-manager-*value*
+* limitador-*value*
+* limitador-operator-controller-manager-*value*
+
+# Further Reading
+With the Kuadrant data plane components installed, here is some further reading material to help you utilise Authorino and Limitador:
+
+[Getting started with Authorino](https://docs.kuadrant.io/authorino/)
+[Getting started With Limitador](https://docs.kuadrant.io/limitador-operator/)
+
+
+
+
+
--- a/docs/how-to/service-protection-setup.md
+++ b/docs/how-to/service-protection-setup.md
@@ -14,7 +14,6 @@ This walkthrough will show you how to install and setup the Kuadrant Operator in
 * Kuadrant uses Istio as a Gateway API provider - this will need to be installed into the data plane clusters
   * We recommend installing Istio 1.17.0, including Gateway API v0.6.2
   * ```bash
-    # On the Hub cluster:
     kubectl apply -k "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.6.2"
     ```
   * See also: https://istio.io/v1.17/blog/2022/getting-started-gtwapi/
--- a/config/kuadrant/redis/limitador/kustomization.yaml
+++ b/config/kuadrant/redis/limitador/kustomization.yaml
@@ -6,4 +6,4 @@ secretGenerator:
     literals:
       -  URL=redis://172.31.0.3:30611
     options:
-      disableNameSuffixHash: true   
+      disableNameSuffixHash: true
--- a/config/service-protection-install-guide/kustomization.yaml
+++ b/config/service-protection-install-guide/kustomization.yaml
@@ -1,3 +1,2 @@
 resources:
-  - ../default
   - managed-cluster-addon.yaml
--- a/config/service-protection-install-guide/kustomization.yaml
+++ b/config/service-protection-install-guide/kustomization.yaml
@@ -1,0 +1,3 @@
+resources:
+  - ../default
+  - managed-cluster-addon.yaml
--- a/config/service-protection-install-guide/managed-cluster-addon.yaml
+++ b/config/service-protection-install-guide/managed-cluster-addon.yaml
@@ -1,0 +1,7 @@
+apiVersion: addon.open-cluster-management.io/v1alpha1
+kind: ManagedClusterAddOn
+metadata:
+ name: kuadrant-addon
+spec:
+ installNamespace: open-cluster-management-agent-addon
+ 
--- a/docs/how-to/kuadrant-hub-install.md
+++ b/docs/how-to/kuadrant-hub-install.md
@@ -163,4 +163,4 @@ clusterissuer.cert-manager.io/mgc-ca condition met
 
 Now that you have MGC installed and configured in your hub cluster, you can now continue with any of these follow-on guides:
 
-- Installing the Kuadrant data-plane pieces [TODO: link to this]
+- Installing the [Kuadrant Service Protection components](./service-protection-setup.md)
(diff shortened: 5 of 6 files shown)


docs/how-to/data-plane-installation.md
```diff
@@ -0,0 +1,74 @@
+# Installing Kuadrant data-plane into an existing OCM Managed Cluster
+
+## Introduction
+This walkthrough will show you how to install and setup the Kuadrant Operator into an [OCM](https://open-cluster-management.io/) [Managed Cluster](https://open-cluster-management.io/concepts/managedcluster/).
```
pmccarthy, line 4: Is there a need to have two separate links for OCM and the Managed Cluster definition? Just the one pointing at https://open-cluster-management.io/concepts/managedcluster would work I think

docs/how-to/data-plane-installation.md
```diff
@@ -0,0 +1,74 @@
...
+
+Steps:
+- OCM should be installed and you should have an existing OCM hub cluster (we should cover the specifics of that setup in the cp install guide which this should link to) with the control plane components installed (IE follow the CP guide first)
+- Install Istio version x.y as this is the gateway provider and we do not install this (we can provide a ref set of install steps)
+- Install OLM as the service protection components are delivered by OLM currently
+- Ensure this cluster is added as a managed cluster to your OCM hub (can be same cluster or another)
+- create the ManagedClusterAddon resource in the managed cluster namespace to trigger the install of kuadrant into the spoke cluster
+
+## Prerequisites
+* Access to an OCM Managed Cluster, which has already been bootstrapped and registered with a hub cluster
```
maleck13, line 14: we should specify the version of OCM to use >=0.6.0

These versions are captured in the dependencies.make file

docs/how-to/data-plane-installation.md
```diff
@@ -0,0 +1,74 @@
+# Installing Kuadrant data-plane into an existing OCM Managed Cluster
+
+## Introduction
+This walkthrough will show you how to install and setup the Kuadrant Operator into an [OCM](https://open-cluster-management.io/) [Managed Cluster](https://open-cluster-management.io/concepts/managedcluster/).
+
+Steps:
+- OCM should be installed and you should have an existing OCM hub cluster (we should cover the specifics of that setup in the cp install guide which this should link to) with the control plane components installed (IE follow the CP guide first)
```
pmccarthy, line 7: Should this be linking out to an OCM install guide? Something along the lines of what's outlined [here] (https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/docs/how-to/ocm-control-plane-walkthrough.md) via the quickstart script?

pmccarthy: Perhaps work to add a CP install guide is already underway, if it is feel free to ignore the above comment

docs/how-to/data-plane-installation.md
```diff
@@ -0,0 +1,74 @@
...
+
+The above command will install the `ManagedClusterAddOn` resource needed to install the Kuadrant addon into the `kind-mgc-workload-1` namespace, and install the Kuadrant data-plane components into the `open-cluster-management-agent-addon` namespace. 
+
+The Kuadrant addon will install:
+
+* the Kuadrant Operator
+* Limitador (and its associated operator)
+* Authorino  (and its associated operator)
+
+For more details, see the Kuadrant components installed by the (kuadrant-operator)[https://github.com/Kuadrant/kuadrant-operator#kuadrant-components]
```
Ygnas, line 48:
```suggestion
For more details, see the Kuadrant components installed by the [kuadrant-operator](https://github.com/Kuadrant/kuadrant-operator#kuadrant-components)
```
I think it should be like this

docs/how-to/data-plane-installation.md
jasonmadigan: A link to a guide would help, although I guess we'd link to upstream OCM - this guide is moreso about installing data-plane pieces onto an existing OCM, so unlikely they would've used our guides to install OCM originally. Although it wouldn't hurt to mention using our quickstart guide would give them a good way to do initial tyre kicking.

docs/how-to/data-plane-installation.md
jasonmadigan: have a dropped a mention of this guide into the pre-requisites now

docs/how-to/data-plane-installation.md
jasonmadigan: OK - added both!

jasonmadigan:
Going to rework this a bit to tie it closer to https://github.com/Kuadrant/multicluster-gateway-controller/pull/460

docs/how-to/data-plane-installation.md
```diff
@@ -0,0 +1,72 @@
+# Installing Kuadrant data-plane into an existing OCM Managed Cluster
+
+## Introduction
+This walkthrough will show you how to install and setup the Kuadrant Operator into an [OCM](https://open-cluster-management.io/) [Managed Cluster](https://open-cluster-management.io/concepts/managedcluster/).
+
+## Prerequisites
```
maleck13, line 6: This is looking pretty good to me but wondering if we want to link to the install guide now as a pre-req?

docs/how-to/data-plane-installation.md
```diff
@@ -0,0 +1,72 @@
+# Installing Kuadrant data-plane into an existing OCM Managed Cluster
```
maleck13, line 1: We might change this to `Service Protection` ?

jasonmadigan: better, yes

docs/how-to/data-plane-installation.md
jasonmadigan: yep, will add a bi-directional link between them

jasonmadigan:
@maleck13 addressed these comments, thanks for review

jasonmadigan:
@maleck13 had the same thought as well while reading the new MGC setup doc tbh. 

WDYT about including a bit on installing Istio GW? Maybe not OLM, but Istio might be worth having a few steps for. Doesn't have to be in this PR, can land it just after.

docs/how-to/service-protection-setup.md
```diff
@@ -0,0 +1,74 @@
+# Installing Kuadrant Service Protection into an existing OCM Managed Cluster
+
+## Introduction
+This walkthrough will show you how to install and setup the Kuadrant Operator into an [OCM](https://open-cluster-management.io/) [Managed Cluster](https://open-cluster-management.io/concepts/managedcluster/).
+
+## Prerequisites
+* Access to an Open Cluster Management (>= 0.6.0) Managed Cluster, which has already been bootstrapped and registered with a hub cluster
```
pehala, line 7: In the https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/docs/how-to/mgc-setup.md guide, there is OCM 0.11.0 required.

jasonmadigan: good spot, fixed. thought I'd made these consistent in an earlier commit but I guess not

laurafitzgerald:
Does this guide remove the need for us to maintain https://github.com/Kuadrant/multicluster-gateway-controller/tree/main/config/kuadrant now?

docs/how-to/service-protection-setup.md
```diff
@@ -0,0 +1,74 @@
...
+kind: ManagedClusterAddOn
+metadata:
+ name: kuadrant-addon
+ namespace: kind-mgc-workload-1
+spec:
+ installNamespace: open-cluster-management-agent-addon
+EOF
+```
+
+**Note:** if you've run our Quickstart Setup guide, you'll  be set to run this command as-is.
```
laurafitzgerald, line 38: Should we include a link for the reference `Quickstart Setup` so it's clear what's being referred to here? What if they haven't run that? are there other steps needed?

laurafitzgerald:
@maleck13 what would you think then about using the kustomize directory here rather than a number of manual steps. We could apply that kustomize?

maleck13:
> 

@laurafitzgerald do you mean to replace? https://github.com/Kuadrant/multicluster-gateway-controller/pull/452/files#diff-fd05af1ee1684ec2aedf2f27cb950c76ec59dbbc66c2574366c4d3c58175d9d6R38

docs/how-to/service-protection-setup.md
```diff
@@ -0,0 +1,74 @@
...
+  * We have [a guide](./kuadrant-hub-install.md) which covers this in detail
+  * Also see:
+    * https://open-cluster-management.io/getting-started/quick-start/
+    * https://open-cluster-management.io/concepts/managedcluster/
+* OLM will need to be installed into the ManagedCluster where you want to run the Kuadrant Service Protection components
+  * See https://olm.operatorframework.io/docs/getting-started/
+* Kuadrant uses Istio as a Gateway API provider - this will need to be installed into the data plane clusters
+  * See https://istio.io/v1.16/blog/2022/getting-started-gtwapi/
+  * We recommend installing Istio 1.17.0
+
```
maleck13, line 17: I think we want a pre-req of Gateway API v0.6.2 as with the hub we can prob provide a command to install

docs/how-to/service-protection-setup.md
```diff
@@ -0,0 +1,74 @@
...
+## Prerequisites
+* Access to an Open Cluster Management (>= v0.11.0) Managed Cluster, which has already been bootstrapped and registered with a hub cluster
+  * We have [a guide](./kuadrant-hub-install.md) which covers this in detail
+  * Also see:
+    * https://open-cluster-management.io/getting-started/quick-start/
+    * https://open-cluster-management.io/concepts/managedcluster/
+* OLM will need to be installed into the ManagedCluster where you want to run the Kuadrant Service Protection components
+  * See https://olm.operatorframework.io/docs/getting-started/
+* Kuadrant uses Istio as a Gateway API provider - this will need to be installed into the data plane clusters
+  * See https://istio.io/v1.16/blog/2022/getting-started-gtwapi/
```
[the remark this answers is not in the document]
maleck13: @eguzki wondering if we have a better resource for installing istio and also would like to confirm the version you are testing against.
@jasonmadigan In that blog it also reference 0.5.2 of GWAPI wondering if we just want to point them at basic install

```
curl -L https://istio.io/downloadIstio | sh -
$ cd istio-<version>
$ ./bin/istioctl install --set profile=minimal -y
```

seems to be the relevant command

docs/how-to/service-protection-setup.md
```diff
@@ -0,0 +1,74 @@
...
+This walkthrough will show you how to install and setup the Kuadrant Operator into an [OCM](https://open-cluster-management.io/) [Managed Cluster](https://open-cluster-management.io/concepts/managedcluster/).
+
+## Prerequisites
+* Access to an Open Cluster Management (>= v0.11.0) Managed Cluster, which has already been bootstrapped and registered with a hub cluster
+  * We have [a guide](./kuadrant-hub-install.md) which covers this in detail
+  * Also see:
+    * https://open-cluster-management.io/getting-started/quick-start/
+    * https://open-cluster-management.io/concepts/managedcluster/
+* OLM will need to be installed into the ManagedCluster where you want to run the Kuadrant Service Protection components
+  * See https://olm.operatorframework.io/docs/getting-started/
```
maleck13, line 13: I wonder is it worth just saying
```
if you have operator-sdk installed at version >=N then you can simply run operator-sdk olm install against each of your managed clusters?

File: docs/how-to/service-protection-setup.md [hunk 12]
```diff
@@ -0,0 +1,74 @@
...
+  * See https://istio.io/v1.16/blog/2022/getting-started-gtwapi/
+  * We recommend installing Istio 1.17.0
+
+
+Alternatively, if you'd like to quickly get started locally, without having to worry to much about the pre-requisites, take a look our [Quickstart Guide](./ocm-control-plane-walkthrough.md). It will get you setup with Kind, OLM, OCM & Kuadrant in a few short steps.
+
+
+## Install the Kuadrant OCM Add-On
+
+To install the Kuadrant Service Protection components into a `ManagedCluster`, target your cluster and run:
```
maleck13, line 24:
```suggestion
To install the Kuadrant Service Protection components into a `ManagedCluster`, target your OCM hub cluster with kubectl and run:
```

docs/how-to/service-protection-setup.md
```diff
@@ -0,0 +1,74 @@
...
+
+```bash
+kubectl apply -f - <<EOF
+apiVersion: addon.open-cluster-management.io/v1alpha1
+kind: ManagedClusterAddOn
+metadata:
+ name: kuadrant-addon
+ namespace: kind-mgc-workload-1
```
maleck13, line 32: this namespace is likely wrong

docs/how-to/service-protection-setup.md
```diff
@@ -0,0 +1,74 @@
...
+spec:
+ installNamespace: open-cluster-management-agent-addon
+EOF
+```
+
+**Note:** if you've run our [Quickstart Guide](./ocm-control-plane-walkthrough.md), you'll  be set to run this command as-is.
+
+The above command will install the `ManagedClusterAddOn` resource needed to install the Kuadrant addon into the `kind-mgc-workload-1` namespace, and install the Kuadrant data-plane components into the `open-cluster-management-agent-addon` namespace.
```
maleck13, line 40: might be better to move this warning to above the command?

docs/how-to/service-protection-setup.md
```diff
@@ -0,0 +1,74 @@
...
+
+* the Kuadrant Operator
+* Limitador (and its associated operator)
+* Authorino  (and its associated operator)
+
+For more details, see the Kuadrant components installed by the (kuadrant-operator)[https://github.com/Kuadrant/kuadrant-operator#kuadrant-components]
+
+## Verify the Kuadrant addon installation
+
+To verify the Kuadrant OCM addon has installed currently, run:
```
maleck13, line 52: maybe a command like 
```kubectl wait --timeout=5m -n kuadrant-system deployment/authorino-operator deployment/kuadrant-operator-controller-manager deployment/limitador-operator-controller-manager --for=condition=Available ```
which is what we use in the hub also

File: docs/how-to/service-protection-setup.md (hunk 10 above)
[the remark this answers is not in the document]
Comment (maleck13) [reply docs/how-to/service-protection-setup.md:15]: hmm ok so using this install method may not be right way currently. As I recall we need to use the `IstioOperator` resource

File: docs/how-to/service-protection-setup.md (hunk 15 above)
Comment (jasonmadigan) [reply docs/how-to/service-protection-setup.md:52]: yeah, I liked this from the other guide

File: docs/how-to/service-protection-setup.md (hunk 10 above)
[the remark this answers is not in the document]
Comment (maleck13) [reply docs/how-to/service-protection-setup.md:15]: also wondering if they already have istio installed is there a way they can point kuadrant to it and is there any specific config we need them to apply to an existing install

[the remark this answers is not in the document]
Comment (maleck13) [reply docs/how-to/service-protection-setup.md:15]: looking at the istio-system namesapce I see there is an existing IstioOperator resource called ```installed-state``` so perhaps we can change find a way to change the defualt Kuadrant looks at. Looks to an env ```ISTIOOPERATOR_NAME ```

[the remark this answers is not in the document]
Comment (maleck13) [reply docs/how-to/service-protection-setup.md:15]: 1.19 is the latest release. Will see does it work with that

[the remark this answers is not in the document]
Comment (eguzki) [reply docs/how-to/service-protection-setup.md:15]: Using currently [ISTIOVERSION = 1.17.2](https://github.com/Kuadrant/kuadrant-operator/blob/7dcbc3dabbe13d2092adeaf018b2aaccb0eee23a/make/istio.mk#L11)

We do not have doc in service protection to install istio. We link to upstream istio installation guide. Istio installation is out of scope of kuadrant (a.k.a. `batteries not included`) currently.

For our dev/testing environment, what we currently do is 
* download `istioctl` CLI tool for the specific version.
* run: `istioctl operator init` to install the Istio controller
* Apply [IstioOperator CR](https://github.com/Kuadrant/kuadrant-operator/blob/7dcbc3dabbe13d2092adeaf018b2aaccb0eee23a/config/dependencies/istio/istio-operator.yaml) with the following customizations:
  * Default ingress gateway enabled 
  * No egree gateway
  * No cni
  * istio namespace: `istio-system`

From their [Install operator guide](https://istio.io/latest/docs/setup/install/operator/), the operator is deprecated. ( on the other hand, from ServiceMesh side I read that there is WIP new operator for Istio to converge OSSM and istio :shrug: ) 
```
Use of the operator for new Istio installations is discouraged in favor of the [Istioctl](https://istio.io/latest/docs/setup/install/istioctl) and [Helm](https://istio.io/latest/docs/setup/install/helm) installation methods. While the operator will continue to be supported, new feature requests will not be prioritized.
```
Thus, we need to update the installation method for the dev/testing env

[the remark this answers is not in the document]
Comment (maleck13) [reply docs/how-to/service-protection-setup.md:15]: yeah although `istioctl install --set profile=minimal -y` still creates it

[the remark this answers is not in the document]
Comment (eguzki) [reply docs/how-to/service-protection-setup.md]: As of today, only used to register Authorino (to enable AuthPolicy), if the Istio installation has some custom values other than default values, yes, they need to be set as env vars
```go
func controlPlaneProviderName() string {                                         
    return env.GetString("ISTIOOPERATOR_NAME", "istiocontrolplane")              
}                                                                                
                                                                                 
func controlPlaneConfigMapName() string {                                        
    return env.GetString("ISTIOCONFIGMAP_NAME", "istio")                         
}                                                                                
                                                                                 
func controlPlaneProviderNamespace() string {                                    
    return env.GetString("ISTIOOPERATOR_NAMESPACE", "istio-system")              
}                                                                                

```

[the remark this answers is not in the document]
maleck13: PR to allow these to be customised https://github.com/Kuadrant/multicluster-gateway-controller/pull/494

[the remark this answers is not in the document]
maleck13: @jasonmadigan once the above merges. We can add a section about ```changing the default istiooperator name``` 

To do that you will now install the add-on as normal. Then in the cluster you want to modify the name / location, you will have to set 

```
k annotate managedclusteraddon kuadrant-addon "addon.open-cluster-management.io/values"='{"IstioOperator":"installed-state"}' -n <managed cluster>
```
this will propigate down and update the Kuadrant operator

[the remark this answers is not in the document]
jasonmadigan: pushed some new words to a new section for bring-your-own-istio scenarios.

docs/how-to/service-protection-setup.md
```diff
@@ -0,0 +1,91 @@
...
+The Kuadrant addon will install:
+
+* the Kuadrant Operator
+* Limitador (and its associated operator)
+* Authorino  (and its associated operator)
+
+For more details, see the Kuadrant components installed by the (kuadrant-operator)[https://github.com/Kuadrant/kuadrant-operator#kuadrant-components]
+
+### Existing Istio installations and changing the default Istio Operator name
+In the case where you have an existing Istio installation to a cluster you may encounter an issue where the Kuadrant Operator expects Istio's Operator to be named `istiocontrolplane`.
```
maleck13, line 57:
```suggestion
In the case where you have an existing Istio installation on a cluster, you may encounter an issue where the Kuadrant Operator expects Istio's Operator to be named `istiocontrolplane`.
```

docs/how-to/service-protection-setup.md
```diff
@@ -0,0 +1,81 @@
...
+  * Also see:
+    * https://open-cluster-management.io/getting-started/quick-start/
+    * https://open-cluster-management.io/concepts/managedcluster/
+* OLM will need to be installed into the ManagedCluster where you want to run the Kuadrant Service Protection components
+  * See https://olm.operatorframework.io/docs/getting-started/
+* Kuadrant uses Istio as a Gateway API provider - this will need to be installed into the data plane clusters
+  * We recommend installing Istio 1.17.0, including Gateway API v0.6.2
+  * ```bash
+    # On the Hub cluster:
+    kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null || \
```
maleck13, line 18: hmm I think we can simplify and unify this with the hub 
```
kubectl apply -k "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.6.2"
```

docs/how-to/service-protection-setup.md
```diff
@@ -0,0 +1,81 @@
...
+`kubectl annotate managedclusteraddon kuadrant-addon "addon.open-cluster-management.io/values"='{"IstioOperator":"installed-state"}' -n <managed spoke cluster>`
+
+This will propogate down and update the Kuadrant Operator, used by the Kuadrant OCM Addon.
+
+## Verify the Kuadrant addon installation
+
+To verify the Kuadrant OCM addon has installed currently, run:
+
+```bash
+kubectl wait --timeout=5m -n kuadrant-system deployment/authorino-operator deployment/kuadrant-operator-controller-manager deployment/limitador-operator-controller-manager --for=condition=Available
```
maleck13, line 62:
```suggestion
kubectl wait --timeout=5m -n kuadrant-system kuadrant/kuadrant-sample  --for=condition=Ready
```

docs/how-to/service-protection-setup.md
```diff
@@ -0,0 +1,80 @@
...
+  * We have [a guide](./kuadrant-hub-install.md) which covers this in detail
+  * Also see:
+    * https://open-cluster-management.io/getting-started/quick-start/
+    * https://open-cluster-management.io/concepts/managedcluster/
+* OLM will need to be installed into the ManagedCluster where you want to run the Kuadrant Service Protection components
+  * See https://olm.operatorframework.io/docs/getting-started/
+* Kuadrant uses Istio as a Gateway API provider - this will need to be installed into the data plane clusters
+  * We recommend installing Istio 1.17.0, including Gateway API v0.6.2
+  * ```bash
+    # On the Hub cluster:
```
maleck13, line 17:
```suggestion
```

maleck13 (approved):

PR #460, opened 2023-08-29
#280: Setting up MGC in Existing OCM Clusters

### LINKED ISSUE #280: Install guide for the multi cluster gateway controller (mgc)
is linked to #427 
**what**

beyond the quick start we should provide an install guide for setting up kuadrant in an OCM installation that already exists.

We should make the following assumptions

1) the hub and spoke clusters may already exists
2) we are being installed by a cluster admin for the hub

Rough Steps:

Pre-req:
Existing OCM installation at version x.y Link off to OCM installation docs

1) Configure OCM with the required feature gate (see comment below)
2) Using kustomize and from our tagged remote release (for now just use master but we will have an actual release tag in the near future) install the MGC, add-on manager, Gateway API CRDS and Cert-Manager in to the multi-cluster-gateways ns in the hub 
3) Configure a ManagedZone with google or AWS credentials
4) Create a ClusterCertificateIssuer (can be self signed like what we use locally show example and link off to cert-manager docs)
5) verify the setup 
          - Check ManagedZone status
          - Check the add-on manager and MGC deployment status

We should also add a set of debugging help

- Getting the logs 

The follow on from this should be the installation of the service protection components in the spoke cluster.

When writing this we can test it against a local kind cluster with OCM installed, however we should also test it against our Hosted cluster

eoinfennessy:
/closes #280 

## What

A how-to guide and kustomization directory for installing MGC in preexisting OCM clusters

## Verification

Run the [base setup gist](https://gist.github.com/eoinfennessy/3954dfc68fcc54ef524d00558dc85d98) and continue with the `mgc-setup.md` walkthrough. There are some verification steps in the walkthrough, but we may want to add more?


docs/how-to/mgc-setup.md
```diff
@@ -0,0 +1,142 @@
+# Setting up MGC in Existing OCM Clusters
+
+This guide will show you how to install and configure the Multi-Cluster Gateway Controller in existng [Open Cluster Management](https://open-cluster-management.io/)-configured clusters.
```
Ygnas, line 3:
```suggestion
This guide will show you how to install and configure the Multi-Cluster Gateway Controller in existing [Open Cluster Management](https://open-cluster-management.io/)-configured clusters.
```

eoinfennessy: Oops! Thanks Ignas!

eoinfennessy:
@ficap @jsmolar, Here's the guide for installing MGC in preexisting OCM-configured clusters.

config/mgc-install-guide/kustomization.yaml
```diff
@@ -0,0 +1,20 @@
+resources:
+  - ../cert-manager
```
maleck13, line 2: is there a way to optionally include this? Some clusters may already have cert manager installed

eoinfennessy: I'll try moving installation of cert-manager to the "Creating a Cert Issuer" step and mark it as optional.

docs/how-to/mgc-setup.md
```diff
@@ -0,0 +1,145 @@
...
+```bash
+kustomize --load-restrictor LoadRestrictionsNone build "github.com/eoinfennessy/multicluster-gateway-controller.git/config/mgc-install-guide?ref=mgc_setup_guide_for_ocm" --enable-helm | kubectl apply -f -
+```
+
+In addition to the MGC, this will also install the Kuadrant add-on manager, Gateway API CRDs, and [cert-manager](https://cert-manager.io/).
+
+After the configuration has been applied, you can verify that the MGC and add-on manager have been installed and are running:
+
+```bash
+kubectl get pods -n multicluster-gateway-controller-system
```
maleck13, line 42: Might be work using ```kubectl wait --timeout=5m -n multicluster-gateway-controller-system deployment/mgc-controller-manager deployment/mgc-kuadrant-add-on-manager --for=condition=Available ```?

docs/how-to/mgc-setup.md
```diff
@@ -0,0 +1,145 @@
...
+  name: mgc-aws-credentials
+  namespace: multicluster-gateway-controller-system
+type: "kuadrant.io/aws"
+stringData:
+  AWS_ACCESS_KEY_ID: ${MGC_AWS_ACCESS_KEY_ID}
+  AWS_SECRET_ACCESS_KEY: ${MGC_AWS_SECRET_ACCESS_KEY}
+  AWS_REGION: ${MGC_AWS_REGION}
+EOF
+
+cat <<EOF | kubectl apply -f -
```
maleck13, line 72: hmm are these actually needed?

maleck13: I don't think this configmap is needed

docs/how-to/mgc-setup.md
```diff
@@ -0,0 +1,145 @@
...
+mgc-kuadrant-add-on-manager-7479cb7f9c-rzk4w   1/1     Running   0          2m
+```
+
+## Creating a ManagedZone
+
+To manage the creation of DNS records, MGC uses [ManagedZone](https://docs.kuadrant.io/multicluster-gateway-controller/docs/how-to/managedZone/) resources. A `ManagedZone` can be configured to use DNS Zones on either AWS (Route53), and GCP. We will now create a ManagedZone on the cluster using AWS credentials.
+
+First, export the [environment variables detailed here](https://docs.kuadrant.io/multicluster-gateway-controller/docs/getting-started/#config) in a terminal session.
+
+Next, create a secret and configMap containing the AWS credentials and config [TODO: is the ConfigMap needed here?]:
```
maleck13, line 56: In my doc around providers, I will list the AWS and Google Perms we need so the user knows what to provide and what we can do

docs/how-to/mgc-setup.md
eoinfennessy: The secret is used for the `dnsProviderSecretRef` in the `ManagedZone` config. I'm not sure, but I think we may be able to do without the configMap as the zone root domain & ID is specified in the `ManagedZone` spec.

docs/how-to/mgc-setup.md
```diff
@@ -0,0 +1,145 @@
...
+kind: ClusterIssuer
+metadata:
+  name: glbc-ca
+  namespace: cert-manager
+spec:
+  selfSigned: {}
+EOF
+```
+
+Verify that the `clusterIssuer` is ready:
```
maleck13, line 131: same here can prob use a similar wait command as above

docs/how-to/mgc-setup.md
```diff
@@ -0,0 +1,145 @@
...
+kubectl get clusterissuers -n cert-manager
+```
+```
+NAME      READY   AGE
+glbc-ca   True    3s
+```
+
+## Next Steps
+
+Now that you have MGC installed and configured in your hub cluster, you can now continue with any of these follow-on guides:
```
maleck13, line 143: I think we are missing the setup of the gatewayclass

maleck13:
This seemed to work pretty well overall. I didn't have a gateway provider on my spokes yet so could only get so far but was able to see the controllers running and setup a MZ and create a gateway.
This should be rebased on master as my perms changes there are needed. 
also I think we should include setting up the default gatewayclass

docs/how-to/mgc-setup.md
maleck13: Yeah I tested without and it worked

docs/how-to/mgc-setup.md
eoinfennessy: I added a section on creating the `GatewayClass`, but just thinking now would it be better to remove this section and apply the `GatewayClass` in the kustomize diectory that deploys MGC, add-on manager,  etc?

eoinfennessy: I ran into an issue doing this because the gateway API CRDs and GatewayClass are applied together and because of this the CRDs aren't ready in time for the GatewayClass creation. Latest commit still separates creation of GatewayClass, but the section is a bit shorter and the resource is remotely applied.

Another option we could possibly take with this would be to:
1. Install gateway API CRDs first
2. Then apply the MGC, add-on manager, GatewayClass, etc. using kustomize

maleck13: I think installing Gateway API at a particular version is a pre-req so installing it first makes sense +1

config/mgc-install-guide/namespace.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    control-plane: controller-manager
+  name: multi-cluster-gateways
```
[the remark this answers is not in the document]
eoinfennessy: I moved the MGC deployment into a `config/deploy/remote` dir in the latest commit. There's quite a few config files currently in `config/quick-start/control-cluster/cluster-config` that may need to be reused in multiple setups. Do you have any opinion where these should be located?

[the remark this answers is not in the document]
eoinfennessy: Removed the `deploy/remote` dir and changed the default namespace in `config/default`

config/default/kustomization.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 # Adds namespace to all resources.
-namespace: multi-cluster-gateways
+namespace: multicluster-gateway-controller-system
```
mikenairn, line 4: Makes sense to me :+1:

config/mgc-install-guide/cert-manager/kustomization.yaml
```diff
@@ -0,0 +1,17 @@
...
+    metadata:
+      name: mgc-cert-manager-webhook
+    $patch: delete
+  - |-
+    apiVersion: admissionregistration.k8s.io/v1
+    kind: MutatingWebhookConfiguration
+    metadata:
+      name: mgc-cert-manager-webhook
+      namespace: cert-manager
+    $patch: delete
```
mikenairn, line 17: We have this duplicated in two places (here and quickstart), and local-setup does the equivalent manually. This should be in one place and used everywhere where we installing cert-manager. Considering `config/cert-manager` is our kustomization of cert manager installs and we remove these in all installations I'd be in favour of just adding this to `config/cert-manager/kustomization.yaml` 

@roivaz Do we use this elsewhere for real deployments where we might want webhooks enabled?

@eoinfennessy We can leave it like this for now, I'll look into changing it as part of a config directory cleanup.

config/mgc-install-guide/gatewayclass.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: GatewayClass
+metadata:
+  name: kuadrant-multi-cluster-gateway-instance-per-cluster
+spec:
+  controllerName: "kuadrant.io/mgc-gw-controller"
```
mikenairn, line 6: I think this should either just be added as part of the manager installation (It is expected that Gateway resources exist for it anyway) or just a separate step after mgc installation. Since it;s pretty much a requirement of MGC that this exists for anything to work currently I'd probably add it to the manager install.

@eoinfennessy It's fine for now

config/mgc-install-guide/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+resources:
+  - ../default
+  - namespace.yaml
+  - gatewayclass.yaml
```
mikenairn, line 4: We should able to remove this. Installation of mgc should be possible in all scenarios by applying the `config/default` kustomization, now that you have updated the namespace in that to a sensible default, it will always be correct and you don't need the namespace.yaml here.

config/mgc-install-guide/namespace.yaml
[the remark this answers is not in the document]
mikenairn: You should be bale to remove this namespace.yaml

config/quick-start/control-cluster/kustomization.yaml
```diff
@@ -1,8 +1,9 @@
 resources:
-  - controller
+  - ../../default
```
mikenairn, line 2: Should be fine now that you have updated the `namespace` in default

docs/how-to/mgc-setup.md
```diff
@@ -0,0 +1,163 @@
...
+```
+
+We will now create a `ClusterIssuer` to be used with `cert-manager`. For simplicity, we will create a self-signed cert issuer here, but [other issuers can also be configured](https://cert-manager.io/docs/configuration/).
+
+```bash
+cat <<EOF | kubectl apply -f -
+apiVersion: cert-manager.io/v1
+kind: ClusterIssuer
+metadata:
+  name: glbc-ca
```
mikenairn, line 143: Should probably really call this something different, `mgc-ca` maybe.

config/quick-start/control-cluster/kustomization.yaml
```diff
@@ -1,8 +1,9 @@
...
   - ../../istio
   - ../../metallb
   - ../../cert-manager
+  - ../../gateway-api
```
mikenairn, line 6: I think this is fine, it does seem to make more sense here over where it was before

config/mgc-install-guide/kustomization.yaml
eoinfennessy: This creates the `multi-cluster-gateways` namespace, which is used later in the guide as the namespace for the `ManagedZone`. Do you think I should delete this directory and create this namespace separately or use a different namespace for the `ManagedZone`?

mikenairn: Yeah, so IMO i don't think we should create that ns as part of any kustomization. This is a namespace that the user is going to create stuff in and should really be whatever they want to call it. In fact, i don't think anything that the user is expected to create should be created via kustomize and should be called out in the docs explicitly  as resources they should apply (You are already doing this in your doc for MZs and Issuers).

docs/how-to/mgc-setup.md
```diff
@@ -0,0 +1,163 @@
...
+
+This guide will show you how to install and configure the Multi-Cluster Gateway Controller in preexisting [Open Cluster Management](https://open-cluster-management.io/)-configured clusters.
+
+## Prerequisites
+
+- A **hub cluster** running the OCM control plane (v0.11.0 or greater)
+- Any number of additional **spoke clusters** that have been configured as OCM [ManagedClusters](https://open-cluster-management.io/concepts/managedcluster/)
+- [Kubectl](https://kubernetes.io/docs/tasks/tools/#kubectl) (>= v1.14.0)
+- [Kustomize](https://kubectl.docs.kubernetes.io/installation/kustomize/)
+- [Helm](https://helm.sh/docs/intro/quickstart/#install-helm)
```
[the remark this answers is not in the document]
jsmolar: I would consider swapping the Helm with a cert-manager as part of the prerequisites.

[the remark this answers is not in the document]
mikenairn: It will be if you are installing cert-manager. If you are just installing the controller, probably not.

jsmolar:
We managed to successfully install MGC into existing OCM clusters (1 hub 2 spoke) according to this guide.

docs/how-to/mgc-setup.md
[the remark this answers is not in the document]
eoinfennessy: Helm is required if installing `cert-manager`.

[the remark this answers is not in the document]
eoinfennessy: Apologies @jsmolar - I only saw your first message when I replied earlier. I'll add something in the prerequisites to the effect of `either a preexisting cert-manager installation or the kustomize and helm CLIs so that it can be installed`.

mikenairn:
@eoinfennessy I think this looks fine :+1:  

We will want to get rid of the `config/mgc-install-guide` eventually, but i think this should probably be done as part of a bigger task to tidy up this whole config directory to remove all the duplication that has crept in.

/lgtm
/approve

docs/how-to/mgc-setup.md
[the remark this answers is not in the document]
jsmolar: @eoinfennessy no worries. Thank you! My suggestion was based on a discussion we had earlier, that we should avoid using more tools than necessary in one example/quick start (Kustomize and Helm)

Issue #423, opened 2023-08-11
Demo DNSPolicy Geo and Weighted Loadbalancing


maleck13:

maleck13:
**What**

Put together a demo showing how to use DNSPolicy to setup to balance traffic based on geolocation and also with custom weights

PR #462, opened 2023-08-30
Label selectors for dnspolicy custom weights

### LINKED ISSUE #376: Kuberentes label selectors for the DNSPolicy custom settings
**Proposal:** use proper [Kubernetes label selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) to indicate the managed cluster resources that relate to a particular weighted load balancing custom config in a `DNSPolicy`, instead of the current [`value`](https://github.com/Kuadrant/multicluster-gateway-controller/blob/7027be4cdb264ebc6a185263754d99b821d61b32/pkg/apis/v1alpha1/dnspolicy_types.go#L55) field.

Other kinds of Kuadrant policies rely on label selectors for similar references between resources that are not direct references by name, e.g. https://github.com/Kuadrant/authorino/blob/579aff40bddd5863d997f77d6ce08b4af5e6b22c/api/v1beta1/auth_config_types.go#L282.

A couple advantages of using label selectors:
- Kubernetes users are usually familiar with those (or can refer to the Kubernetes docs);
- Implementation of the internal API call that effectively selects those related objects can be simplified;
- Support for exact label matchers and expressions (e.g. `kuadrant.io/az in (us, uk)`).

mikenairn:
Updates the DNSPolicy spec to use a kubernetes label selector instead of the `value` field when setting a custom weight. This brings it more inline with how other policies work, is a more familiar pattern in kubernetes generally and expands the uses cases for the custom weight feature.

The label used is no longer dictated by us, and any label a user wanted to use can now be specified in the DNSPolicy as part of the selector. Multiple labels can be specified if the user wanted more than one attribute to match before a weight is applied.

```
- weight: 255
  selector:
     matchLabels:
        my-label-1: val1
        my-label-2: val2
```

/closes #376


mikenairn:
@jasonmadigan fyi, when this is merged it will change the [demo](https://github.com/Kuadrant/multicluster-gateway-controller/issues/423l) you are doing slightly.

david-martin:
👀

jasonmadigan:
> @jasonmadigan fyi, when this is merged it will change the [demo](https://github.com/Kuadrant/multicluster-gateway-controller/issues/423) you are doing slightly.

No probs, thanks for the heads-up. I've scripted most of the terminal stuff in the DNSPolicy using [vhs](https://github.com/charmbracelet/vhs), so changing this when this lands or is usable won't be too big a deal. Is it usable now? I've not tried it.

pkg/dns/target.go
```diff
@@ -149,34 +127,38 @@ func (t *ClusterGatewayTarget) GetWeight() int {
...
 func (t *ClusterGatewayTarget) setWeight(defaultWeight int, customWeights []*v1alpha1.CustomWeight) {
-	weight := &defaultWeight
-	if t.ClusterAttributes.CustomWeight != nil && customWeights != nil {
-		for k := range customWeights {
-			cw := customWeights[k]
-			if *t.ClusterAttributes.CustomWeight == cw.Value {
+	weight := defaultWeight
+	for k := range customWeights {
+		cw := customWeights[k]
+		if selector, err := metav1.LabelSelectorAsSelector(cw.Selector); err == nil {
```
maleck13, line 153: what does it mean if there is an error here? does it mean it is an invalid selector?

maleck13:
are there any sample DNSPolicy resouces that need to be updated?

pkg/dns/target.go
mikenairn: I did dig into the code a bit to see if i could add a test that would actually cause this to throw an error, but it didn't really seem possible. It calls `labels.NewRequirement` internally and that will throw an error if the values are an incorrect length for the type of match you want to do, Exact in the case of Match Labels. These values will alwys be what is expected when they are passed though `LabelSelectorAsSelector` afaik.

Even if there is some way to cause this to throw an error it would be based on a misconfiguration of the selector and I'm not sure what else we can really do other than continue and use the default in that scenario.

mikenairn:
> are there any sample DNSPolicy resouces that need to be updated?

Everything that i can find, in this repo at least, has been updated.

david-martin:
Reviewed & verified locally.
I had an issue locally where the custom weight wasn't being applied to the dnsrecord, which went away after rebuilding and deploying a new image with extra logs.
However I wasn't able to reproduce it after. I suspect it may have been an old image deployed at first?
Either way, I'm happy with the combinations I've tried locally to say this is behaving as expected.

/approve

Will hold off on lgtm until any other reviewers are done.

pkg/dns/target.go
mikenairn: Found a way to make it fail. It checks the format of the labels and expects any with a / to have 2 parts, so this is an example of something that wouldn't work and would case an error here:

```
Selector: &metav1.LabelSelector{
   MatchLabels: map[string]string{
      "/tstlabel1": "TSTATTR",
   },
},
```

Added a test to with the above, but not really sure we can do anything other than ignore it. A log might be useful for debug purposes.

maleck13: hmm why would we not add something to the status to report that the label selector was wrong?

maleck13: Right. I guess we would need to bubble the error back up? The issue that I see is this is a public API and if I use it incorrectly, I would want to know what I have done wrong. With how it is now, it seems I would be left guessing? 
If we agree that the error should make its way to the status then this is something we could follow up with.

mikenairn: I added the error https://github.com/Kuadrant/multicluster-gateway-controller/pull/462/commits/6e09f71650c14f872080a6898dfb8e43f4abc60d

Makes sense i guess since it's something they could mess up.

guicassolato (approved):
Awesome!

pkg/dns/target.go
mikenairn:
``` status:
  conditions:
  - lastTransitionTime: "2023-08-31T09:34:39Z"
    message: 'reconcile DNSRecords error failed to create multi cluster gateway target
      for listener api : key: Invalid value: "/lb-attribute-custom-weight": prefix
      part must be non-empty '
    reason: ReconciliationError
    status: "False"
    type: Ready ```

Comment (mikenairn) [reply pkg/dns/target.go]:
```bash
kubectl get gateway prod-web -n multi-cluster-gateways -o json | jq '.status.conditions[] | select(.type=="kuadrant.io/DNSPolicyAffected")'
{
  "lastTransitionTime": "2023-08-31T09:37:21Z",
  "message": "policy failed. Object unaffected by policy kuadrant.io/v1alpha1, Kind=DNSPolicy in namespace multi-cluster-gateways with name prod-web with error failed to create multi cluster gateway target for listener api : key: Invalid value: \"/lb-attribute-custom-weight\": prefix part must be non-empty ",
  "observedGeneration": 1,
  "reason": "Invalid",
  "status": "False",
  "type": "kuadrant.io/DNSPolicyAffected"
}

```

PR #465, opened 2023-08-31
GH-311 gateway status instead of ocm api calls

### LINKED ISSUE #311: Update dnspolicy reconciler to use gateway status for placed cluster details
**What**

Update the dnspolicy reconciler to use the gateway status to build up the list of `dns.ClusterGateway` instead of having to query ocm APIs to retrieve the placed clusters.

With the introduction of the status aggregation of the hub Gateways as proposed [here](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/docs/proposals/status-aggregation.md), we will now have enough information on the Gateway resource itself to build up the `dns.MultiClusterGatewayTarget` that we need to pass to [SetEndpoints](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go#L114)

This should disconnect the dnspolicy entirely from ocm (apart from the cluster event handler) and fix a current issue where we are only dealing with IPAddress type addresses.

Requires:

* https://github.com/Kuadrant/multicluster-gateway-controller/issues/113

makslion:
## What 
Change in how we get information about clusters onto which we placed gateway. Instead of poking the OCM we now extract those from the GW CR itself. 
## Verification 
1. Follow one of the submariner PoCs up to the point where you have gateway placed on the multiple clusters 
2. Ensure that DNSRecords have endpoints set 

closes #311

### DIFF
--- a/pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
+++ b/pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
@@ -143,21 +143,26 @@ func (r *DNSPolicyReconciler) deleteGatewayDNSRecords(ctx context.Context, gatew
 
 func (r *DNSPolicyReconciler) buildClusterGateway(ctx context.Context, downstreamClusterName string, clusterAddress []gatewayv1beta1.GatewayAddress) (dns.ClusterGateway, error) {
 	var target dns.ClusterGateway
+	singleClusterAddresses := make([]gatewayv1beta1.GatewayAddress, len(clusterAddress))
 
 	mc := &clusterv1.ManagedCluster{}
 	if err := r.Client().Get(ctx, client.ObjectKey{Name: downstreamClusterName}, mc, &client.GetOptions{}); err != nil {
 		return target, err
 	}
 
-	for _, addr := range clusterAddress {
+	for i, addr := range clusterAddress {
 		addrType := gatewayv1beta1.IPAddressType
 		if *addr.Type == gateway.MultiClusterHostnameAddressType {
 			addrType = gatewayv1beta1.HostnameAddressType
 		}
-		addr.Type = &addrType
+
+		singleClusterAddresses[i] = gatewayv1beta1.GatewayAddress{
+			Type:  &addrType,
+			Value: addr.Value,
+		}
 	}
 
-	target = *dns.NewClusterGateway(mc, clusterAddress)
+	target = *dns.NewClusterGateway(mc, singleClusterAddresses)
 
 	return target, nil
 }
@@ -167,11 +172,6 @@ func getClusterGatewayAddresses(gw *gatewayv1beta1.Gateway) map[string][]gateway
 
 	for _, address := range gw.Status.Addresses {
 		var gatewayAddresses []gatewayv1beta1.GatewayAddress
-
-		//addressType := gatewayv1beta1.IPAddressType
-		//if *address.Type == gateway.MultiClusterHostnameAddressType {
-		//	addressType = gatewayv1beta1.HostnameAddressType
-		//}
 
 		//Default to Single Cluster (Normal Gateway Status)
 		cluster := "none"
--- a/pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
+++ b/pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
@@ -3,6 +3,9 @@ package dnspolicy
 import (
 	"context"
 	"fmt"
+	"strings"
+
+	clusterv1 "open-cluster-management.io/api/cluster/v1"
 
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
@@ -13,6 +16,7 @@ import (
 	"github.com/kuadrant/kuadrant-operator/pkg/reconcilers"
 
 	"github.com/Kuadrant/multicluster-gateway-controller/pkg/apis/v1alpha1"
+	"github.com/Kuadrant/multicluster-gateway-controller/pkg/controllers/gateway"
 	"github.com/Kuadrant/multicluster-gateway-controller/pkg/dns"
 )
 
@@ -47,14 +51,9 @@ func (r *DNSPolicyReconciler) reconcileGatewayDNSRecords(ctx context.Context, ga
 		return err
 	}
 
-	placed, err := r.Placer.GetPlacedClusters(ctx, gateway)
-	if err != nil {
-		log.V(3).Info("error getting placed clusters")
-		return err
-	}
-	clusters := placed.UnsortedList()
+	clusters := getGatewayAddresses(gateway)
 
-	log.V(3).Info("checking gateway for attached routes ", "gateway", gateway.Name, "clusters", placed)
+	log.V(3).Info("checking gateway for attached routes ", "gateway", gateway.Name, "clusters", clusters)
 
 	for _, listener := range gateway.Spec.Listeners {
 		var clusterGateways []dns.ClusterGateway
@@ -70,21 +69,20 @@ func (r *DNSPolicyReconciler) reconcileGatewayDNSRecords(ctx context.Context, ga
 		for _, downstreamCluster := range clusters {
 			// Only consider host for dns if there's at least 1 attached route to the listener for this host in *any* gateway
 
-			log.V(1).Info("checking downstream", "listener ", listener.Name)
-			attached, err := r.Placer.ListenerTotalAttachedRoutes(ctx, gateway, string(listener.Name), downstreamCluster)
-			if err != nil {
-				log.Error(err, "failed to get total attached routes for listener ", "listener", listener.Name)
-				continue
-			}
+			log.V(3).Info("checking downstream", "listener ", listener.Name)
+			attached := listenerTotalAttachedRoutes(gateway, downstreamCluster)
+
 			if attached == 0 {
 				log.V(1).Info("no attached routes for ", "listener", listener.Name, "cluster ", downstreamCluster)
 				continue
 			}
-			log.V(1).Info("hostHasAttachedRoutes", "host", listener.Name, "hostHasAttachedRoutes", attached)
-			cg, err := r.Placer.GetClusterGateway(ctx, gateway, downstreamCluster)
+			log.V(3).Info("hostHasAttachedRoutes", "host", listener.Name, "hostHasAttachedRoutes", attached)
+
+			cg, err := r.buildClusterGateway(ctx, gateway, downstreamCluster)
 			if err != nil {
 				return fmt.Errorf("get cluster gateway failed: %s", err)
 			}
+
 			clusterGateways = append(clusterGateways, cg)
 		}
 
@@ -141,4 +139,62 @@ func (r *DNSPolicyReconciler) deleteGatewayDNSRecords(ctx context.Context, gatew
 		}
 	}
 	return nil
+}
+
+func (r *DNSPolicyReconciler) buildClusterGateway(ctx context.Context, upstreamGateway *gatewayv1beta1.Gateway, downstreamCluster string) (dns.ClusterGateway, error) {
+	var target dns.ClusterGateway
+
+	mc := &clusterv1.ManagedCluster{}
+	if err := r.Client().Get(ctx, client.ObjectKey{Name: downstreamCluster}, mc, &client.GetOptions{}); err != nil {
+		return target, err
+	}
+
+	for _, address := range upstreamGateway.Status.Addresses {
+		if strings.Contains(address.Value, downstreamCluster) {
+			var gatewayAddresses []gatewayv1beta1.GatewayAddress
+
+			addressType := gatewayv1beta1.IPAddressType
+			if strings.Contains(string(*address.Type), string(gateway.MultiClusterHostnameAddressType)) {
+				addressType = gatewayv1beta1.HostnameAddressType
+			}
+
+			tmp := strings.Split(address.Value, "/")
+			addressValue := tmp[0]
+
+			if len(tmp) > 0 {
+				addressValue = tmp[len(tmp)-1]
+			}
+			gatewayAddresses = append(gatewayAddresses, gatewayv1beta1.GatewayAddress{
+				Type:  &addressType,
+				Value: addressValue,
+			})
+			target = *dns.NewClusterGateway(mc, gatewayAddresses)
+		}
+	}
+	return target, nil
+}
+
+func getGatewayAddresses(upstreamGateway *gatewayv1beta1.Gateway) []string {
+	var clusters []string
+
+	for _, address := range upstreamGateway.Status.Addresses {
+		value := strings.Split(address.Value, "/")
+		if strings.Contains(string(*address.Type), string(gateway.MultiClusterHostnameAddressType)) || strings.Contains(string(*address.Type), string(gateway.MultiClusterIPAddressType)) {
+			clusters = append(clusters, value[0])
+		}
+	}
+
+	return clusters
+}
+
+func listenerTotalAttachedRoutes(upstreamGateway *gatewayv1beta1.Gateway, downstreamCluster string) int {
+	listeners := 0
+
+	for _, listener := range upstreamGateway.Status.Listeners {
+		if strings.Contains(string(listener.Name), downstreamCluster) {
+			listeners = int(listener.AttachedRoutes)
+		}
+	}
+
+	return listeners
 }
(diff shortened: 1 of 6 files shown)


mikenairn (changes requested):
The main goal with this task was to remove the dependency on the OCM placement API in the dnspolicy controller and reconcilers. As this is, it's removed a small bit of it but not all. I'd start by removing this https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/controllers/dnspolicy/dnspolicy_controller.go#L66 and see what breaks and make it work without it.

You probably need to process the status of a gateway and build up a map of clusters and it's addresses and iterate over that rather than calling the placer functions, also the attached routes info should be available in both cases directly from the gateway.

Listeners also need to be processed differently to take into account the two types of Gateway (Multi and single):

```
kubectl get gateway prod-web -n multi-cluster-gateways -o json | jq .status.listeners
[
  {
    "attachedRoutes": 1,
    "conditions": [],
    "name": "kind-mgc-control-plane.api",
    "supportedKinds": []
  },
  {
    "attachedRoutes": 1,
    "conditions": [],
    "name": "kind-mgc-workload-1.api",
    "supportedKinds": []
  },
  {
    "attachedRoutes": 1,
    "conditions": [],
    "name": "kind-mgc-workload-2.api",
    "supportedKinds": []
  }
]
```

And the object that represents a "Cluster"  should be any GVK, rather than hard coded to an OCM  "ManagedCluster". This is just a resource that has labels so can be anything, but will always be a "ManagedCluster" in MGC.

test/integration/dnspolicy_controller_test.go
```diff
@@ -198,7 +208,7 @@ var _ = Describe("DNSPolicy", Ordered, func() {
 		var lbHash, dnsRecordName, wildcardDNSRecordName string
 
 		BeforeEach(func() {
-			gateway = testBuildGateway(TestPlacedGatewayName, gatewayClass.Name, TestAttachedRouteName, testNamespace)
+			gateway = testBuildGateway(TestPlacedGatewayName, testutil.DummyCRName, TestAttachedRouteName, testNamespace)
```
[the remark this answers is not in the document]
makslion: yes, and I mention that this should be removed [here](https://github.com/Kuadrant/multicluster-gateway-controller/pull/465/files#diff-2adf2dbe2f36d347bda43e252bea13a9d720c5e9548c14ef693d41433d9da4c6R213)

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -135,3 +132,13 @@ func (r *DNSPolicyReconciler) deleteGatewayDNSRecords(ctx context.Context, gatew
 	}
 	return nil
 }
+
+func getPlacedClusters(gateway *gatewayv1beta1.Gateway) []string {
```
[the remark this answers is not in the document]
makslion: renamed

test/integration/dnspolicy_controller_test.go
```diff
@@ -207,6 +217,14 @@ var _ = Describe("DNSPolicy", Ordered, func() {
 				if err := k8sClient.Get(ctx, client.ObjectKey{Name: gateway.Name, Namespace: gateway.Namespace}, gateway); err != nil {
 					return false
 				}
+
+				// TODO remove the workaround during https://github.com/Kuadrant/multicluster-gateway-controller/issues/330
+				// also use a proper gateway class
+				gateway.Status.Addresses = testBuildGatewayAddresses()
+				if err := k8sClient.Status().Update(ctx, gateway); err != nil {
+					return false
+				}
```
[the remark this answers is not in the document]
makslion: It is not dependent on the OCM, but the code still needs more persuasion to work with other Gateway implementations. Created #496 to implement.

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -135,3 +132,13 @@ func (r *DNSPolicyReconciler) deleteGatewayDNSRecords(ctx context.Context, gatew
...
+	var clusters []string
+
+	for _, address := range gateway.Status.Addresses {
+		clusters = append(clusters, strings.Split(address.Value, "/")[0])
+	}
+
+	return clusters
```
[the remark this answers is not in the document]
makslion: Right now it will ignore ISTO gateways (`clusters` will be empty) and then it will see that we have no `clusterGateways` for this gateway and will opt out of the reconciler cleaning up remaining DNSRecords for this gateway.

test/integration/dnspolicy_controller_test.go
```diff
@@ -882,343 +1199,4 @@ var _ = Describe("DNSPolicy", Ordered, func() {
...
-				patch := client.MergeFrom(gateway.DeepCopy())
-				addressType := mgcgateway.MultiClusterIPAddressType
-				gateway.Status.Addresses = []gatewayv1beta1.GatewayAddress{
-					{
-						Type:  &addressType,
-						Value: fmt.Sprintf("%s/%s", "kind-mgc-control-plane", TestAttachedRouteAddressOne),
-					},
-					{
-						Type:  &addressType,
-						Value: fmt.Sprintf("%s/%s", "kind-mgc-control-plane", TestAttachedRouteAddressTwo),
```
laurafitzgerald, line 1054: I notice this is removed from the test where it's copied to. Is the expectation that the status.address update is covered now by the reconciler or by some change to the test configuration?

laurafitzgerald:
I guess this https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/test/integration/helper_test.go#L63 and it's related functions and structs can be removed now.

test/integration/dnspolicy_controller_test.go
makslion: Right now it is moved [here](https://github.com/Kuadrant/multicluster-gateway-controller/pull/465/files#diff-2adf2dbe2f36d347bda43e252bea13a9d720c5e9548c14ef693d41433d9da4c6R96-R107), but we  do hope to have it [done](https://github.com/Kuadrant/multicluster-gateway-controller/issues/330) by the controller at some point

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -137,3 +135,61 @@ func (r *DNSPolicyReconciler) deleteGatewayDNSRecords(ctx context.Context, gatew
...
+
+	for _, address := range upstreamGateway.Status.Addresses {
+		value := strings.Split(address.Value, "/")
+		if strings.Contains(string(*address.Type), string(gateway.MultiClusterHostnameAddressType)) || strings.Contains(string(*address.Type), string(gateway.MultiClusterIPAddressType)) {
+			clusters = append(clusters, value[0])
+		}
+	}
+
+	return clusters
+}
```
mikenairn, line 183: I think this would probably be better changed to return a map of clusters and addresses by moving your address processing logic into here:

```
func getClusterGatewayAddresses(gw *gatewayv1beta1.Gateway) map[string][]gatewayv1beta1.GatewayAddress {
	var clusterAddrs map[string][]gatewayv1beta1.GatewayAddress

	for _, address := range gw.Status.Addresses {
		var gatewayAddresses []gatewayv1beta1.GatewayAddress

		addressType := gatewayv1beta1.IPAddressType
		if *address.Type == gateway.MultiClusterHostnameAddressType {
			addressType = gatewayv1beta1.HostnameAddressType
		}

		//Default to Single Cluster (Normal Gateway Status)
		cluster := "none"
		addressValue := address.Value

		//Check for Multi Cluster (MGC Gateway Status)
		if *address.Type == gateway.MultiClusterIPAddressType || *address.Type == gateway.MultiClusterHostnameAddressType {
			tmp := strings.Split(address.Value, "/")
			//If this fails something is wrong and the value hasn't been set correctly
			if len(tmp) == 2 {
				cluster = tmp[0]
				addressValue = tmp[1]
			}
		}

		gatewayAddresses = append(gatewayAddresses, gatewayv1beta1.GatewayAddress{
			Type:  &addressType,
			Value: addressValue,
		})
		clusterAddrs[cluster] = gatewayAddresses
	}

	return clusterAddrs
}
```

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -137,3 +135,61 @@ func (r *DNSPolicyReconciler) deleteGatewayDNSRecords(ctx context.Context, gatew
...
+	listeners := 0
+
+	for _, listener := range upstreamGateway.Status.Listeners {
+		if strings.Contains(string(listener.Name), downstreamCluster) {
+			listeners = int(listener.AttachedRoutes)
+		}
+	}
+
+	return listeners
+}
```
mikenairn, line 195: This is always going to return the number of attached routes for that last listener. You need to pass in the listener you want the attached routes for and find that one in the list.

I also wouldn't use `strings.Contains`, if someone creates a listener with the same name as a cluster it will give a false positive. You also can't do this for normal gateway listeners, it will only work for MGC ones. I'd split the listener name by '.' instead and assume if you have a length greater than 1 that the cluster name is the first element see https://github.com/Kuadrant/multicluster-gateway-controller/pull/315#discussion_r1259370454.

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -137,3 +135,61 @@ func (r *DNSPolicyReconciler) deleteGatewayDNSRecords(ctx context.Context, gatew
...
 	return nil
 }
+
+func (r *DNSPolicyReconciler) buildClusterGateway(ctx context.Context, upstreamGateway *gatewayv1beta1.Gateway, downstreamCluster string) (dns.ClusterGateway, error) {
+	var target dns.ClusterGateway
+
+	mc := &clusterv1.ManagedCluster{}
+	if err := r.Client().Get(ctx, client.ObjectKey{Name: downstreamCluster}, mc, &client.GetOptions{}); err != nil {
+		return target, err
+	}
```
mikenairn, line 145: Leaving this here still ties us to MGC. I think the logic around this should be that if we have a multi cluster gateway (You will know this after processing the addresses), we don't check for the existence of a `clusterv1.ManagedCluster` but instead use the gateway itself as the cluster i.e. `dns.NewClusterGateway(upstreamGateway, gatewayAddresses)`.

Admittedly the naming is confused, but `cluster` can be any `metav1.Object` that can have labels attached to it. In the case of a single cluster Gateway, using the gateway itself for this makes the most sense.

mikenairn (changes requested):

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
makslion: Would it be a good idea to address it during the #496 ?

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -137,3 +135,61 @@ func (r *DNSPolicyReconciler) deleteGatewayDNSRecords(ctx context.Context, gatew
...
+	var clusters []string
+
+	for _, address := range upstreamGateway.Status.Addresses {
+		value := strings.Split(address.Value, "/")
+		if strings.Contains(string(*address.Type), string(gateway.MultiClusterHostnameAddressType)) || strings.Contains(string(*address.Type), string(gateway.MultiClusterIPAddressType)) {
+			clusters = append(clusters, value[0])
+		}
+	}
+
+	return clusters
```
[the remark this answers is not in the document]
makslion: renamed

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
makslion: refactored the function

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
makslion: refactored

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -137,3 +135,81 @@ func (r *DNSPolicyReconciler) deleteGatewayDNSRecords(ctx context.Context, gatew
...
+func listenerTotalAttachedRoutes(upstreamGateway *gatewayv1beta1.Gateway, downstreamCluster, specListenerName string, addresses []gatewayv1beta1.GatewayAddress) int {
+	for _, listener := range upstreamGateway.Status.Listeners {
+		// assuming all adresses of the same type on the gateway
+		// for Multi Cluster (MGC Gateway Status)
+		if *addresses[0].Type == gateway.MultiClusterIPAddressType || *addresses[0].Type == gateway.MultiClusterHostnameAddressType {
+			clusterName, listenerName, found := strings.Cut(string(listener.Name), ".")
+			if !found {
+				return 0
+			}
+			if clusterName == downstreamCluster && (listenerName == specListenerName || specListenerName == "wildcard") {
```
mikenairn, line 204: What is the check for "wildcard" for?

makslion: To allow for the wildcard listeners. I could also check for `*` in the hostname :thinking:  And I think I'll do that. And could finally justify the use of the `contains` for checking names and `*` symbols in the listener

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -137,3 +135,81 @@ func (r *DNSPolicyReconciler) deleteGatewayDNSRecords(ctx context.Context, gatew
...
+func listenerTotalAttachedRoutes(upstreamGateway *gatewayv1beta1.Gateway, downstreamCluster string, specListener gatewayv1beta1.Listener, addresses []gatewayv1beta1.GatewayAddress) int {
+	for _, statusListener := range upstreamGateway.Status.Listeners {
+		// assuming all adresses of the same type on the gateway
+		// for Multi Cluster (MGC Gateway Status)
+		if *addresses[0].Type == gateway.MultiClusterIPAddressType || *addresses[0].Type == gateway.MultiClusterHostnameAddressType {
+			clusterName, listenerName, found := strings.Cut(string(statusListener.Name), ".")
+			if !found {
+				return 0
+			}
+			if clusterName == downstreamCluster && (listenerName == string(specListener.Name) || strings.Contains(string(*specListener.Hostname), "*")) {
```
mikenairn, line 204: I'm not seeing why we need to check the host for it being a wildcard, you just need to know if the listener has attached routes or not. Are you trying to get round an issue you are seeing, or was this based on some other bit of code that was being used previously?

makslion: in here I'm matching listeners from spec to the status block to see if I have routes attached. The match is based on the listener's name and the cluster - this is how I find the correct one. In the case of the wildcard, I want to match all of the listeners on the cluster. It is based on the assumption that the wildcard is of a "top level" that could match any listener, I'm returning routes of the first listener.

makslion: or you are saying that in case of the wildcard we will have the spec name of the `wildcard` that would correspond with the same name in the status of the gateway? i.e. `workload-cluster-1.wildcard`?

mikenairn: Why do you think you need to do anything special if it has a wildcard hostname? You just need to find the listener by name and return the attached routes value for it. The gateway controller updating the status will deal with setting the values, we don't need to put our own logic around it.

makslion: changed. And updated tests to explicitly reflect desired behaviour

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -137,3 +135,81 @@ func (r *DNSPolicyReconciler) deleteGatewayDNSRecords(ctx context.Context, gatew
...
+	}
+
+	return clusterAddrs
+}
+
+func listenerTotalAttachedRoutes(upstreamGateway *gatewayv1beta1.Gateway, downstreamCluster string, specListener gatewayv1beta1.Listener, addresses []gatewayv1beta1.GatewayAddress) int {
+	for _, statusListener := range upstreamGateway.Status.Listeners {
+		// assuming all adresses of the same type on the gateway
+		// for Multi Cluster (MGC Gateway Status)
+		if *addresses[0].Type == gateway.MultiClusterIPAddressType || *addresses[0].Type == gateway.MultiClusterHostnameAddressType {
```
mikenairn, line 199: Not ideal having to pass in all the addresses here just to work out if it's an MGC updated status we are dealing with or not. I can see why you have done it this way though, so maybe a better way of determining the type of Gateway we are dealing with, and the expected status, can be looked into later.

makslion: adding this to the [stockpile](https://github.com/Kuadrant/multicluster-gateway-controller/issues/496#issuecomment-1736930282) of "fix later"

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
mikenairn: Created specific follow on for this https://github.com/Kuadrant/multicluster-gateway-controller/issues/622

PR #470, opened 2023-09-01
Scripts and resources for the DNSPolicy demo

jasonmadigan:
Demo: https://youtu.be/TnUE3G6bsoQ

Scripts and resources checked in.

Closes: https://github.com/Kuadrant/multicluster-gateway-controller/issues/423


docs/demos/dns-policy/dnspolicy.tape
sergioifg94, line 1: I'm guessing this .tape file is used in some software to automate the recording of the terminal. Maybe we could mention describe that in the document?

docs/demos/dns-policy/dnspolicy-demo.md
sergioifg94, line 1: Not important at all but I see a few commands that depend on `yq` and `jq`, I was wondering if it might be better to use `-o=jsonpath='...'` instead, in case someone wants to try it and doesn't have those tools

sergioifg94 (commented):
It looks great! Just had a couple of comments on very minor things

docs/demos/dns-policy/dnspolicy.tape
jasonmadigan: Yeah, this is used by https://github.com/charmbracelet/vhs to script the demo run. You're right, that's worth mentioning somewhere (I'm not sure if I can stick comments in the tapes - I think I can)

docs/demos/dns-policy/dnspolicy-demo.md
jasonmadigan: agree - less deps would be preferable

docs/demos/dns-policy/dnspolicy.tape
sergioifg94: Great tool btw, really makes the demo more reproduceable, and better looking!

jasonmadigan:
Some changes based on your suggestions pushed now @sergioifg94

Issue #400, opened 2023-08-03
refactor our integration tests to report errors


maleck13:
**What**

currently we use a lot of the below pattern for our tests 

```
Eventually(func() bool {
  existingDNSPolicy := &v1alpha1.DNSPolicy{}
  err := k8sClient.Get(ctx, client.ObjectKey{Name: dnsPolicy.Name, Namespace: testNamespace}, existingDNSPolicy)
  // must exist
  Expect(err).ToNot(HaveOccurred())
  return metadata.HasFinalizer(existingDNSPolicy, DNSPolicyFinalizer)
}, time.Second*5, time.Second).Should(BeTrue())
```

In many cases it may be  a better to use the following to ensure we don't lose the error

```
Eventually(func() bool {
  existingDNSPolicy := &v1alpha1.DNSPolicy{}
  err := k8sClient.Get(ctx, client.ObjectKey{Name: dnsPolicy.Name, Namespace: testNamespace}, existingDNSPolicy)
  if err != nil{
      return err
  }
 if !metadata.HasFinalizer(existingDNSPolicy, DNSPolicyFinalizer){
   return fmt.Errorf("expected metadata to have finalizer %s",DNSPolicyFinalizer )
  }
  return  nil 
}, time.Second*5, time.Second).Should(BeNil())
```

PR #471, opened 2023-09-04
Refactor tests to report errors clearly

Ygnas:
Closes: #400

### DIFF
--- a/test/integration/dnsheathcheckprobe_controller_test.go
+++ b/test/integration/dnsheathcheckprobe_controller_test.go
@@ -4,6 +4,7 @@ package integration
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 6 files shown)


maleck13:
This looks a lot more informative now on failure. @mikenairn WDYT? I think this is a good approach for these types of errors

mikenairn:
> This looks a lot more informative now on failure. @mikenairn WDYT? I think this is a good approach for these types of errors

Looks good to me. Returning errors makes sense. :+1:

maleck13:
@Ygnas I think we can mark this as ready then I will merge

Ygnas:
> @Ygnas I think we can mark this as ready then I will merge

Well there is more tests to be done.

maleck13:
@Ygnas ah ok cool. Then I think the approach is good and lets covert the rest :) If you wanted to merge in several PRs that would also be fine to avoid potential conflicts and rebases?

maleck13 (approved):
/lgtm

Issue #454, opened 2023-08-24
Don't rely on files in the repo for ocm-walkthrough


laurafitzgerald:
**Why**
We don't want the ocm walkthrough to rely on any local git repo config.

**What**
Remove the use of files that exist in the ocm-walkthrough for kubeconfig context.
Potentially we could just have them contained in the commands themselves as with the yaml files for Gateway etc.

laurafitzgerald:
cc @maleck13

jasonmadigan:
Would this just be the kubeconfig's? IIRC most of the other commands in the how-to are heredocs.

maleck13:
Is there an example of where we do this?

laurafitzgerald:
There's two locations in the guide where we do `kind export kubeconfig` refering to `pwd/local`

PR #482, opened 2023-09-05
use kubectl config instead of kind context and remove need to clone repo locally 

laurafitzgerald:
Addresses https://github.com/Kuadrant/multicluster-gateway-controller/issues/454
And also removes the addition of the step to clone the repo locally from the ocm-walkthrough due to the fact that it's used as the quickstart where we didn't want to have the need to clone the repo locally.

### DIFF
--- a/docs/how-to/ocm-control-plane-walkthrough.md
+++ b/docs/how-to/ocm-control-plane-walkthrough.md
@@ -14,33 +14,23 @@ We will start with a single cluster and move to multiple clusters to illustrate 
 - AWS account with Route 53 enabled
 - [Docker Mac Net Connect](https://github.com/chipmk/docker-mac-net-connect) (macOS users only)
 
->**Note:** :exclamation: this walkthrough will setup a zone in your AWS account and make changes to it for DNS purposes
+>**Note:** :exclamation: this walkthrough will set up a zone in your AWS account and make changes to it for DNS purposes
 
 ## Installation and Setup
-* Clone this repo locally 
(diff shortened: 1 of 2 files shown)


docs/how-to/ocm-control-plane-walkthrough.md
```diff
@@ -61,13 +51,24 @@ The script will
...
+For this walkthrough, we're going to use multiple terminal sessions/windows.
+
+Open two windows, which we'll refer to throughout this walkthrough as:
+
+* `T1` (Hub Cluster)
+* `T2` (Workloads cluster)
+
+* NOTE: MCG_SUB_DOMAIN env var is required in both terminals
+
+In `T1` run `kubectl config use-context kind-mgc-control-plane`
```
laurafitzgerald, line 66: This wont work, it's setting the context for all terminals.

docs/how-to/ocm-control-plane-walkthrough.md
```diff
@@ -61,13 +51,24 @@ The script will
...
+## Open terminal sessions and set cluster context
+
+For this walkthrough, we're going to use multiple terminal sessions/windows.
+
+Open two windows, which we'll refer to throughout this walkthrough as:
+
+* `T1` (Hub Cluster)
+* `T2` (Workloads cluster)
+
+Set the kubecontext for each terminal, refer back to these commands if re-config is needed, this will store a
```
jasonmadigan, line 64: `this will store a` - missing a bit at end? or maybe can be removed

jasonmadigan:
one comment, rest looks good

PR #494, opened 2023-09-07
Allow istio operator location to be set

maleck13:
When installing in an existing spoke cluster, the istioperator resource can be named differently. We currently only expect it to be called `istiocontrolplane`. This can be overridden in the operator but to do this via OCM we need to set params and inject them into the subscription we create.

**Verification**

From a local setup

- Delete any existing managedclusteraddon resources
- Delete the add-on manager if it is running in the cluster
- Targeting the hub run `run-addon-manager`
- create a new managedclusteraddon  ``` kubectl apply -f - <<EOF
apiVersion: addon.open-cluster-management.io/v1alpha1
kind: ManagedClusterAddOn
metadata:
 name: kuadrant-addon
 namespace: kind-mgc-workload-1
spec:
 installNamespace: open-cluster-management-agent-addon
EOF```
- verify that Kudarant is installed and that the subscription env is present  (in the spoke)
- `k get subscription kuadrant-operator -n kuadrant-system -o=yaml`

```
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  creationTimestamp: "2023-09-07T11:48:59Z"
  generation: 2
  labels:
    operators.coreos.com/kuadrant-operator.operators: ""
  name: kuadrant-operator
  namespace: kuadrant-system
  ownerReferences:
  - apiVersion: work.open-cluster-management.io/v1
    kind: AppliedManifestWork
    name: 43967cf4fa7b6c9c1f4014eb104077ac73d86aabc3d2c4ae6c51babdf5898540-addon-kuadrant-addon-deploy-0
    uid: 466f2fd2-6a81-4b8c-b192-dd379cfd25ad
  resourceVersion: "47998"
  uid: b9fc56ca-d03d-477c-99e3-3d5d3fb313ae
spec:
  channel: alpha
  config:
    env:
    - name: ISTIOOPERATOR_NAME
      value: istiocontrolplane
    - name: ISTIOCONFIGMAP_NAME
      value: istio
    - name: ISTIOOPERATOR_NAMESPACE
      value: istio-system
  installPlanApproval: Automatic
  name: kuadrant-operator
  source: operatorhubio-catalog
  sourceNamespace: olm

```

add the following annotation in hub to the mangedclusteraddon 

```
k annotate managedclusteraddon kuadrant-addon "addon.open-cluster-management.io/values"='{"IstioOperator":"installed-state"}' -n cluster1

```
check in the spoke that the env value for `ISTIOOPERATOR_NAME` has been updated


R-Lawton:
👀

PR #495, opened 2023-09-07
#472: Add Interval field to DNS Health checks

### LINKED ISSUE #472: allow default polling time to be overridden via the health check spec
related to https://github.com/Kuadrant/multicluster-gateway-controller/issues/336 

Currently we only have a default polling interval for the health checks. We should allow this to be set by the admin so that they can control the load hitting their services and also allow faster discovery of issues.

In addition we should ensure we validate that the polling interval is not set below a minimum reasonable value

sergioifg94:
> Closes #472

## Description

Add spec.healthCheck.interval field that allows the customisation of the interval the health checks are performed at

## Verification steps

1. Run the walkthrough until the step of the creation of the DNSPolicy
2. Create the DNSPolicy with the following field in the step:
    ```yaml
    healthCheck:
      interval: 90s
    ```
3. Verify that the resulting DNSHealthCheckProbes have the interval set to the interval from the DNSPolicy


laurafitzgerald:
Verified code works as described in the steps

pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -154,6 +154,8 @@ func (r *DNSPolicyReconciler) reconcileResources(ctx context.Context, dnsPolicy
 		return err
 	}
 
+	dnsPolicy.Default()
```
mikenairn, line 157: Is this a pattern used elsewhere? seems odd to me to change the values of something that is already created. Is there some reason we would do this over setting a default value in the CRD? see `defaultWeight` [here](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/apis/v1alpha1/dnspolicy_types.go#L66) and [here](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/config/crd/bases/kuadrant.io_dnspolicies.yaml#L148)

sergioifg94: It is a pattern used by the webhooks. Basically by implementing the `Default` and `Validate` method, it's very straightforward to setup webhooks for the CR. I thought I'd implement them now so we can reuse them if we want to create webhooks for the DNSPolicy, but I agree we should use the default value in the CRD as well

mikenairn: Presumably that would normally be called in a mutating webhook before the object is ever stored in etcd. Just seems odd to call it as part of the reconciliation loop.

sergioifg94: Yes, I'd remove that call from the reconciliation loop when the webhook is added, so it's not duplicated, but as we don't have the webhook yet I put it in the reconcile, so that any following logic can assume that all values have been set

PR #499, opened 2023-09-08
account for probes relating to multiple listeners

laurafitzgerald:
**What**
Account for multiple listeners resulting in more than one probe associate with an ip address/hostname which was the check used to get probes for setting endpoints based on healthy check probes.

The change itself is small.
Adding some additional unit and e2e tests to cover this case as well. 

Relates to https://github.com/Kuadrant/multicluster-gateway-controller/issues/272

### DIFF
--- a/pkg/controllers/dnspolicy/dns_helper.go
+++ b/pkg/controllers/dnspolicy/dns_helper.go
@@ -268,8 +268,9 @@ func (dh *dnsHelper) setEndpoints(ctx context.Context, mcgTarget *dns.MultiClust
 	// count will track whether a new endpoint has been removed.
 	// first newEndpoints are checked based on probe status and removed if unhealthy true and the consecutive failures are greater than the threshold.
 	removedEndpoints := 0
+
 	for i := 0; i < len(newEndpoints); i++ {
-		checkProbes := getProbesForEndpoint(newEndpoints[i], probes)
+		checkProbes := getProbesForEndpoint(newEndpoints[i], probes, mcgTarget.Gateway.Name, string(listener.Name))
 		if len(checkProbes) == 0 {
 			continue
 		}
@@ -453,11 +454,11 @@ func (dh *dnsHelper) getDNSHealthCheckProbes(ctx context.Context, gateway *gatew
 	})
 }
 
-func getProbesForEndpoint(endpoint *v1alpha1.Endpoint, probes []*v1alpha1.DNSHealthCheckProbe) []*v1alpha1.DNSHealthCheckProbe {
+func getProbesForEndpoint(endpoint *v1alpha1.Endpoint, probes []*v1alpha1.DNSHealthCheckProbe, gatewayName, listenerName string) []*v1alpha1.DNSHealthCheckProbe {
 	retProbes := []*v1alpha1.DNSHealthCheckProbe{}
 	for _, probe := range probes {
 		for _, target := range endpoint.Targets {
-			if strings.Contains(probe.Name, target) {
+			if dnsHealthCheckProbeName(target, gatewayName, listenerName) == probe.Name {
 				retProbes = append(retProbes, probe)
 			}
 		}
--- a/test/integration/dnspolicy_controller_test.go
+++ b/test/integration/dnspolicy_controller_test.go
@@ -81,7 +81,7 @@ func testBuildGateway(gwName, gwClassName, hostname, ns, dnspolicy string) *gate
 					Protocol: gatewayv1beta1.HTTPProtocolType,
 				},
 				{
-					Name:     gatewayv1beta1.SectionName(TestWildCardListenerName),
+					Name:     TestWildCardListenerName,
 					Hostname: &wildcardHost,
 					Port:     gatewayv1beta1.PortNumber(80),
 					Protocol: gatewayv1beta1.HTTPProtocolType,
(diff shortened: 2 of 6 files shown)


test/integration/dnspolicy_controller_test.go
```diff
@@ -1220,5 +1220,162 @@ var _ = Describe("DNSPolicy", Ordered, func() {
...
+					Protocol: gatewayv1beta1.HTTPProtocolType,
+				}
+
+				gateway.Spec.Listeners = append(gateway.Spec.Listeners, otherListener)
+				Expect(k8sClient.Update(ctx, gateway)).To(BeNil())
+
+				probeList := &v1alpha1.DNSHealthCheckProbeList{}
+				Eventually(func() error {
+					Expect(k8sClient.List(ctx, probeList, &client.ListOptions{Namespace: testNamespace})).To(BeNil())
+					if len(probeList.Items) != 4 {
```
laurafitzgerald, line 1337: Noting that there are 4 probes here for this test. two addresses, two listeners. previous tests had 2 probes.

test/integration/dnspolicy_controller_test.go
```diff
@@ -1220,5 +1220,162 @@ var _ = Describe("DNSPolicy", Ordered, func() {
...
+						return fmt.Errorf("expected %v probes, got %v", 4, len(probeList.Items))
+					}
+					return nil
+				}, TestTimeoutLong, TestRetryIntervalMedium).Should(BeNil())
+				Expect(len(probeList.Items)).To(Equal(4))
+
+				//
+				Eventually(func() error {
+					getProbe := &v1alpha1.DNSHealthCheckProbe{}
+					if err = k8sClient.Get(ctx, client.ObjectKey{Name: fmt.Sprintf("%s-test-dns-policy-%s", TestAttachedRouteAddressOne, OtherAttachedRouteName), Namespace: testNamespace}, getProbe); err != nil {
```
laurafitzgerald, line 1347: Noting that the probe being set to unhealthy here relates to the other listener/dnsrecord such that the expected endpoints are a full set as this probe doesn't belong to the listener/dnsrecord that's tested at the end of this test.

pkg/controllers/dnspolicy/dns_helper.go
```diff
@@ -453,11 +454,11 @@ func (dh *dnsHelper) getDNSHealthCheckProbes(ctx context.Context, gateway *gatew
 	})
 }
 
-func getProbesForEndpoint(endpoint *v1alpha1.Endpoint, probes []*v1alpha1.DNSHealthCheckProbe) []*v1alpha1.DNSHealthCheckProbe {
+func getProbesForEndpoint(endpoint *v1alpha1.Endpoint, probes []*v1alpha1.DNSHealthCheckProbe, listenerName string) []*v1alpha1.DNSHealthCheckProbe {
 	retProbes := []*v1alpha1.DNSHealthCheckProbe{}
 	for _, probe := range probes {
 		for _, target := range endpoint.Targets {
-			if strings.Contains(probe.Name, target) {
+			if strings.Contains(probe.Name, target) && strings.Contains(probe.Name, listenerName) {
```
sergioifg94, line 461: This looks good, but I wonder if it'd be a good idea to externalise this logic to a separate function, same with the logic that creates the name of the probe when the probes are created. That way if it changes it would be easier to remember to make sure that they match. Wdyt?

laurafitzgerald: Yeah. For sure I can do that. One other thing I considered was adding the listener name to the labels. We have them in annotations but these don't offer the same functionality as labels when it comes to LabelSelector for ListOptions.

sergioifg94: Yeah I think it was me who added these as annotations instead of labels and I kinda regret it

laurafitzgerald: I've kept using the name rather than a rewrite on the annotations/labels for now. Ready for review again if you have some time. 🙏

pkg/controllers/dnspolicy/dns_helper.go
```diff
@@ -268,8 +268,9 @@ func (dh *dnsHelper) setEndpoints(ctx context.Context, mcgTarget *dns.MultiClust
 	// count will track whether a new endpoint has been removed.
 	// first newEndpoints are checked based on probe status and removed if unhealthy true and the consecutive failures are greater than the threshold.
 	removedEndpoints := 0
+
 	for i := 0; i < len(newEndpoints); i++ {
-		checkProbes := getProbesForEndpoint(newEndpoints[i], probes)
+		checkProbes := getProbesForEndpoint(newEndpoints[i], probes, dnsPolicy.Name, string(listener.Name))
```
mikenairn, line 273: Why use the `dnsPoilicy.Name` here? Would gateway  name not make more sense? Would also be more consistent with the DNSRecord naming see https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/controllers/dnspolicy/dns_helper.go#L402

pkg/controllers/dnspolicy/dns_helper.go
```diff
@@ -453,11 +454,11 @@ func (dh *dnsHelper) getDNSHealthCheckProbes(ctx context.Context, gateway *gatew
 	})
 }
 
-func getProbesForEndpoint(endpoint *v1alpha1.Endpoint, probes []*v1alpha1.DNSHealthCheckProbe) []*v1alpha1.DNSHealthCheckProbe {
+func getProbesForEndpoint(endpoint *v1alpha1.Endpoint, probes []*v1alpha1.DNSHealthCheckProbe, policyName, listenerName string) []*v1alpha1.DNSHealthCheckProbe {
 	retProbes := []*v1alpha1.DNSHealthCheckProbe{}
 	for _, probe := range probes {
 		for _, target := range endpoint.Targets {
-			if strings.Contains(probe.Name, target) {
+			if endpointHealthCheckProbe(target, policyName, listenerName, probe.Name) {
```
mikenairn, line 461: `endpointHealthCheckProbe` seems oddly named, would probably be clearer what is happening if you just did the check for equality here:
```
dnsHealthCheckProbeName(target, policyName, listenerName) == probe.Name
```

pkg/controllers/dnspolicy/dns_helper.go
laurafitzgerald: Yeah nice. Didn't know about that one. I've used the dnsnaming function here now. And gateway name rather than policy name. Changes are pushed.

pkg/controllers/dnspolicy/dns_helper.go
laurafitzgerald: yep. makes sense. addressed in the latest push. 👍

Issue #410, opened 2023-08-08
refactor the gateway controller


maleck13:
**What**

the gateway controller is overly complex

- break out status handling
- break out deletion handling
- break out gateway params handling

philbrookes:
This issue is stale because it has been open for 60 days with no activity.

philbrookes:
This issue was closed because it has been inactive for 30 days since being marked as stale.

PR #510, opened 2023-09-12
Ensure gateway not marked programmed if error with tls secrets

### LINKED ISSUE #509: bug: hub gateway marked as programmed when TLS secret missing
related to #507 
**What**

If a TLS secret is specified in the gateway definition but is not present, the controller ignores this and marks the gateway as programmed. 
The Gateway in the hub should not be marked as programmed until all TLS secrets are present.

**Reproduce**

Create a Gateway with HTTPS listener
```
kind: Gateway
metadata:
  name: prod-web
  labels:
    cluster.open-cluster-management.io/placement: http-gateway
spec:
  gatewayClassName: kuadrant-multi-cluster-gateway-instance-per-cluster
  listeners:
  - allowedRoutes:
      namespaces:
        from: All
    name: specific
    hostname: 'specific.cb.hcpapps.net'
    port: 443
    protocol: HTTPS
    tls:
      mode: Terminate
      certificateRefs:
      - kind: Secret
        name: specific.cb.hcpapps.net

```

See that the gateway is placed. The downstream gateway provider does not mark the gateway programmed as the secret is missing. The MGC marks the gateway as Programmed 

```
cbrookes-mac:multicluster-gateway-controller maleck13$ k get gateways -A
NAMESPACE                         NAME       CLASS                                                 ADDRESS   PROGRAMMED   AGE
kuadrant-multi-cluster-gateways   prod-web   istio                                                                        98s
multi-cluster-gateways            prod-web   kuadrant-multi-cluster-gateway-instance-per-cluster             True         99s
```

maleck13:
fixes #509

### DIFF
--- a/pkg/controllers/gateway/gateway_controller.go
+++ b/pkg/controllers/gateway/gateway_controller.go
@@ -171,8 +171,14 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	// gateway now in expected state, place gateway and its associated objects in correct places. Update gateway spec/metadata
 	log.V(3).Info("reconcileDownstreamFromUpstreamGateway result ", "requeue", requeue, "status", programmedStatus, "clusters", clusters, "Err", reconcileErr)
 	if reconcileErr != nil {
+		//TODO (cbrookes) refactor how status is handled in this controller
 		if errors.Is(reconcileErr, gracePeriod.ErrGracePeriodNotExpired) || requeue {
 			log.V(3).Info("requeueing gateway ", "error", reconcileErr, "requeue", requeue)
+			programmedCondition := buildProgrammedCondition(upstreamGateway.Generation, clusters, metav1.ConditionUnknown, reconcileErr)
+			meta.SetStatusCondition(&upstreamGateway.Status.Conditions, programmedCondition)
+			if !isDeleting(upstreamGateway) && !reflect.DeepEqual(upstreamGateway.Status, previous.Status) {
+				return reconcile.Result{}, r.Status().Update(ctx, upstreamGateway)
+			}
 			return reconcile.Result{
 				Requeue:      true,
 				RequeueAfter: 30 * time.Second,
@@ -244,6 +250,7 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	upstreamGateway.Status.Listeners = allListenerStatuses
 
 	acceptedCondition := buildAcceptedCondition(upstreamGateway.Generation, metav1.ConditionTrue)
+	fmt.Println("PROGRAMMED CONDITION ** ", err, programmedStatus)
 	programmedCondition := buildProgrammedCondition(upstreamGateway.Generation, clusters, programmedStatus, err)
 
 	meta.SetStatusCondition(&upstreamGateway.Status.Conditions, acceptedCondition)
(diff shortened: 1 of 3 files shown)


maleck13:
related to https://github.com/Kuadrant/multicluster-gateway-controller/issues/525

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -244,6 +250,7 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	upstreamGateway.Status.Listeners = allListenerStatuses
 
 	acceptedCondition := buildAcceptedCondition(upstreamGateway.Generation, metav1.ConditionTrue)
+	fmt.Println("PROGRAMMED CONDITION ** ", err, programmedStatus)
```
jasonmadigan, line 253: can remove

jasonmadigan:
Changes look good. Important enough to land during freeze?

maleck13:
@jasonmadigan yes I thnk we can merge. Can approve

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -171,8 +171,14 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	// gateway now in expected state, place gateway and its associated objects in correct places. Update gateway spec/metadata
 	log.V(3).Info("reconcileDownstreamFromUpstreamGateway result ", "requeue", requeue, "status", programmedStatus, "clusters", clusters, "Err", reconcileErr)
 	if reconcileErr != nil {
+		//TODO (cbrookes) refactor how status is handled in this controller
```
mikenairn, line 174: Do we have an issue for this?

jasonmadigan (approved):

pkg/controllers/gateway/gateway_controller.go
maleck13: yes although it could prob do with expanding on the details https://github.com/Kuadrant/multicluster-gateway-controller/issues/410

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -356,7 +364,7 @@ func (r *GatewayReconciler) getTLSSecrets(ctx context.Context, upstreamGateway *
 			}
 		}
 	}
-	return tlsSecrets, nil
+	return tlsSecrets, listenerTLSErr
 }
```
mikenairn, line 368: This is probably fine, but it is worth pointing out that this changes the behaviour a bit since nothing is going to get placed on the downstream clusters until all tls secrets that are expected exist in the control plane.

I suppose since we are syncing secrets for gateways it's up to us to decide what the behaviour of our controller should be when secrets don't exist, but would it ever be possible that users just don't want to add their TLS secrets to the control plane cluster like this? Maybe they already have a mechanism to create/sync secrets to their clusters.

maleck13: That is possible I guess. But at this point not something we could say we support. I am trying to think what would trigger a user to want to do it this way? 
If we feel the behaviour here is reasonable, I think it is fine to say we don't support that. If we get enough asks we can change our decision on this ofc

mikenairn: Someone wanting to use MGC, but not have it manage TLS. I already have clusters, i already use certmanager in some way to generate TLS secrets for gateways but i want to use MGC/OCM to manage gateways across these clusters.

If/when we do the status aggregation from the downstream clusters you could still achieve this if it is something we wanted to allow.

PR #523, opened 2023-09-12
Refactor endpoint probe check

laurafitzgerald:
**What**
A refactor on how we are evaluating endpoints based on health checks
Doing this first and updating the MultiClusterGatewayTarget before attempting to setEndpoints.
Unit tests have been adjusted
No changes required to the integration tests as the functionality here isn't changing, just the approach.

**Verification**
Follow steps described in https://github.com/Kuadrant/multicluster-gateway-controller/pull/412 as this is a refactor, functionality shouldn't be changed.
Also integration tests covered all cases and those have passed without any change.


pkg/dns/target.go
```diff
@@ -153,6 +153,70 @@ func (t *ClusterGatewayTarget) setGeo(defaultGeo GeoCode) {
...
+	return retProbes
+}
+
+func dnsHealthCheckProbeName(address, gatewayName, listenerName string) string {
+	return fmt.Sprintf("%s-%s", address, dnsRecordName(gatewayName, listenerName))
+}
+
+func dnsRecordName(gatewayName, listenerName string) string {
+	return fmt.Sprintf("%s-%s", gatewayName, listenerName)
+}
```
laurafitzgerald, line 218: Need to find a better way to manage these as they are currently duplicated as cannot import the package.

pkg/dns/target.go
```diff
@@ -153,6 +153,70 @@ func (t *ClusterGatewayTarget) setGeo(defaultGeo GeoCode) {
...
+	//Build a map of gateway addresses and their health status
+	gwAddressHealth := map[string]bool{}
+	allunhealthy := true
+	for _, cgt := range t.ClusterGatewayTargets {
+		for _, gwa := range cgt.GatewayAddresses {
+			gwProbes := getProbeForGatewayAddress(probes, gwa, t.Gateway.Name, string(listener.Name))
+			//if all probes are unhealthy return as we want to include all endpoints
+			if len(gwProbes) > 1 {
+				fmt.Println("more than 1 probe found")
+			}
```
mikenairn, line 172: Can remove this i think?

pkg/dns/target.go
```diff
@@ -153,6 +153,70 @@ func (t *ClusterGatewayTarget) setGeo(defaultGeo GeoCode) {
...
+	if len(probes) == 0 {
+		return
+	}
+
+	//Build a map of gateway addresses and their health status
+	gwAddressHealth := map[string]bool{}
+	allunhealthy := true
+	for _, cgt := range t.ClusterGatewayTargets {
+		for _, gwa := range cgt.GatewayAddresses {
+			gwProbes := getProbeForGatewayAddress(probes, gwa, t.Gateway.Name, string(listener.Name))
```
mikenairn, line 168: Why would we ever have more than one probe for a gateway/listeenr/address?

pkg/dns/target.go
mikenairn: Hmm, not ideal. We really shouldn't be using the name as the unique id anyway, labels and selectors would probably be a better way of selecting matching resources, not that that helps here.

You could at least get rid of the duplication of `dnsRecordName` by changing `dnsHealthCheckProbeName` to `fmt.Sprintf("%s-%s", address, gatewayName, listenerName)`, your not gaining much by calling it.

pkg/controllers/dnspolicy/dns_helper_test.go
```diff
@@ -1023,805 +966,16 @@ func Test_dnsHelper_setEndpoints(t *testing.T) {
...
-					},
-					{
-						DNSName:    "test.example.com",
-						Targets:    []string{"lb-0ecjaw.test.example.com"},
-						RecordType: "CNAME",
-						RecordTTL:  dns.DefaultCnameTTL,
-					},
-				},
-			},
-		},
```
mikenairn, line 1818: Lot of test cases being removed here. Are the not still valid? Do these need to be moved/rewritten for `RemoveUnhealthyGatewayAddresses`?

laurafitzgerald: yes, they're not valid for setEndpoints anymore but could/should be rewritten for the new function which is doing the logic now. 👍

pkg/dns/target.go
laurafitzgerald: we shouldn't, probably being too cautious here. i'll make an update. 👍

pkg/dns/target.go
laurafitzgerald: the current setup of labels and selectors doesn't provide what we need to pick up the correct probename. We could refactor that. Would you preference be to do that here or given we are using this method currently on main should we create another JIRA for it? i don't mind either way.

pkg/controllers/dnspolicy/dns_helper_test.go
laurafitzgerald: I should say the integration tests required no change and they cover a lot of cases here including multiple listeners. 😄

pkg/dns/target.go
mikenairn: Yeah i'd say just leave it using the name and we look at it later. DNSRecord has the same issue, we should really be doing the same thing for all created resources.

pkg/controllers/dnspolicy/dns_helper_test.go
mikenairn: If you think they are covered by integration tests we might not need them, or at least all of them. Some tests for `RemoveUnhealthyGatewayAddresses` might be useful though, and also might be a bit easier to test the results rather than checking all endpoints in a DNSRecord.

pkg/dns/target.go
mikenairn: Since we are using the name as the id, i doubt you can really. That said, maybe you could have multiple probes eventually. 

If you decide to leave the check for more than one probe, the logic around deciding if any are healthy or not is probably wrong. It;s currently iterating all probes, setting `allunhealthy` to true if any are healthy, but only using the health of the last probe in the list when setting the value in `gwAddressHealth`

pkg/dns/target.go
```diff
@@ -153,6 +153,70 @@ func (t *ClusterGatewayTarget) setGeo(defaultGeo GeoCode) {
...
+				fmt.Println("more than 1 probe found")
+			}
+			for _, probe := range gwProbes {
+				probeHealthy := true
+				if probe.Status.Healthy != nil {
+					probeHealthy = *probe.Status.Healthy
+				}
+				if probeHealthy && probe.Spec.FailureThreshold != nil && probe.Status.ConsecutiveFailures < *probe.Spec.FailureThreshold {
+					allunhealthy = false
+				}
```
mikenairn, line 180: This is going to set `allunhealthy` to false when any are unhealthy is it not?

pkg/controllers/dnspolicy/dns_helper_test.go
laurafitzgerald: i pushed a set of tests for that function. Because there's no return it's really just a coverage test and would only really identify if the function changes that there were some issues with nils for example. It's making me think that the call for this could be done as part of the Construction of the GatewayTarget. wdyt?

pkg/dns/target.go
laurafitzgerald: Given how we are currently constructing the name if there were more than one I would expect that to be a bug https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/controllers/dnspolicy/dnspolicy_healthchecks.go#L132

pkg/dns/target.go
mikenairn: I see, `allunhealthy` is set to true initially

pkg/controllers/dnspolicy/dns_helper_test.go
mikenairn: The test can still test the  `MultiClusterGatewayTarget` is in the expected state after calling the function even if it returns nothing. This would be the same if you did it via a constructor or not.

TestCase:
1. Create the MultiClusterGatewayTarget 
 ```
mgt := &MultiClusterGatewayTarget{
				Gateway:               tt.fields.Gateway,
				ClusterGatewayTargets: tt.fields.ClusterGatewayTargets,
				LoadBalancing:         tt.fields.LoadBalancing,
			}
```
3. Call RemoveUnhealthyGatewayAddresses
```
mgt.RemoveUnhealthyGatewayAddresses(tt.args.probes, tt.args.listener)
```
5. Check  MultiClusterGatewayTarget.ClusterGatewayTargets are what you expect them to be
```
if mgt.ClusterGatewayTargets != testCase.want { <- This will need to actually check the equality of the arrays
   t.Errorf("got = %v, want %v", got, testCase.want)
}
```

pkg/dns/target.go
laurafitzgerald: Captured this here https://github.com/Kuadrant/multicluster-gateway-controller/issues/534

pkg/controllers/dnspolicy/dns_helper_test.go
laurafitzgerald: facepalm. thanks. 👍

mikenairn:
Changes look good to me, can merge after the release is done.  :+1:

mikenairn (approved):

Issue #540, opened 2023-09-13
[BUG]Quick start script failing periodically waiting for istio system


R-Lawton:
## WHAT: 
The quick start script is falling periodically while waiting for the following line to finish https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/hack/.deployUtils#L423

Error:
```
deployment.apps/mgc-cert-manager-webhook condition met
Waiting for istio deployments to be ready
deployment.apps/istio-operator condition met
error: no matching resources found
```

maleck13:
hmm looking at the script I see

```
kubectl -n istio-operator wait --timeout=300s --for=condition=Available deployments --all
kubectl -n istio-system wait --timeout=300s --for=condition=Available deployments --all
```
So perhaps the namespace `istio-operator` is not there?

PR #542, opened 2023-09-13
Gh 540

R-Lawton:
`export MGC_BRANCH=gh-540 export  MGC_ACCOUNT=R-Lawton`

`curl https://raw.githubusercontent.com/R-Lawton/multicluster-gateway-controller/gh-540/hack/quickstart-setup.sh | bash`

Screenshot 2023-09-14 at 12 45 33

### DIFF
--- a/hack/.deployUtils
+++ b/hack/.deployUtils
@@ -418,9 +418,9 @@ deployQuickStartControl() {
     kubectl -n metallb-system wait --for=condition=ready pod --selector=app=metallb --timeout=300s
     echo "Waiting for cert-manager deployments to be ready"
     kubectl -n cert-manager wait --timeout=300s --for=condition=Available deployments --all
-    echo "Waiting for istio deployments to be ready"
-    kubectl -n istio-operator wait --timeout=350s --for=condition=Available deployments --all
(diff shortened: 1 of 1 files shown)


maleck13:
So I was thinking about this, I am not sure increasing the timeout is actually the solution. 
If we look at the error
```
deployment.apps/mgc-cert-manager-webhook condition met
Waiting for istio deployments to be ready
deployment.apps/istio-operator condition met
error: no matching resources found
```
it is that there are no matching resources found. So I think it is more likely that we are doing the check before the deployment is in place. So adding more time wont stop the error IMO

R-Lawton:
> So I was thinking about this, I am not sure increasing the timeout is actually the solution. If we look at the error
> 
> ```
> deployment.apps/mgc-cert-manager-webhook condition met
> Waiting for istio deployments to be ready
> deployment.apps/istio-operator condition met
> error: no matching resources found
> ```
> 
> it is that there are no matching resources found. So I think it is more likely that we are doing the check before the deployment is in place. So adding more time wont stop the error IMO

Ok ill try a few other things, it is running now the last few times I've run in with the change but might still be flaky

maleck13 (approved):

PR #557, opened 2023-09-14
Fixes #553

### LINKED ISSUE #553: bug: if gcp is selected in the Quickstart Script, and a user has an AWS access key env-var, the script will setup an AWS ManagedZone + secret, rather than a GCP one
Noticed while testing https://github.com/Kuadrant/multicluster-gateway-controller/issues/480

If gcp is selected in the Quickstart Script, and a user has an AWS access key env-var, the script will setup an AWS ManagedZone + secret, rather than a GCP one. Chatted with @R-Lawton about it - reckon its because the initial selection context gets lost between scripts,. The script that sets up the zones + secrets defaults to AWS if a `MGC_AWS_ACCESS_KEY_ID` env var is detected.

jasonmadigan:
Attempting to fix https://github.com/Kuadrant/multicluster-gateway-controller/issues/553

### DIFF
--- a/hack/.deployUtils
+++ b/hack/.deployUtils
@@ -322,13 +322,13 @@ deployPrometheusForFederation() {
 
 # quickstart-setup specific functions
 
-configureController() {
-    clusterName=${1}
-    kubectl config use-context kind-${clusterName}
-    echo "Initialize local dev setup for the controller on ${clusterName}"
+setupAWSProvider() {
+  if [ "$MGC_AWS_ACCESS_KEY_ID" == "" ]; then
+    echo "MGC_AWS_ACCESS_KEY_ID is not set"
+    exit 1
+  fi
 
-  if [ -n "$MGC_AWS_ACCESS_KEY_ID" ]; then
-    cat <<EOF | kubectl apply -f -
+  kubectl apply -f - <<EOF
 apiVersion: v1
 kind: Secret
 metadata:
@@ -339,9 +339,7 @@ stringData:
   AWS_ACCESS_KEY_ID: ${MGC_AWS_ACCESS_KEY_ID}
   AWS_SECRET_ACCESS_KEY: ${MGC_AWS_SECRET_ACCESS_KEY}
   AWS_REGION: ${MGC_AWS_REGION}
-EOF
-
-    cat <<EOF | kubectl apply -f -
+---
 apiVersion: v1
 kind: ConfigMap
 metadata:
@@ -351,9 +349,7 @@ data:
   AWS_DNS_PUBLIC_ZONE_ID: ${MGC_AWS_DNS_PUBLIC_ZONE_ID}
   ZONE_ROOT_DOMAIN: ${MGC_ZONE_ROOT_DOMAIN}
   LOG_LEVEL: "${LOG_LEVEL}"
-EOF
-
-    cat <<EOF | kubectl apply -f -
+---
 apiVersion: kuadrant.io/v1alpha1
 kind: ManagedZone
 metadata:
@@ -367,8 +363,10 @@ spec:
     name: ${KIND_CLUSTER_PREFIX}aws-credentials
     namespace: multicluster-gateway-controller-system
 EOF
-else
-    cat <<EOF | kubectl apply -f -
+}
+
+setupGCPProvider() {
+  kubectl apply -f - <<EOF
 apiVersion: v1
 kind: Secret
 metadata:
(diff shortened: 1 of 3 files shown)


hack/quickstart-setup.sh
```diff
@@ -23,12 +23,25 @@ if [ -z $MGC_ACCOUNT ]; then
 MGC_ACCOUNT=${MGC_ACCOUNT:="kuadrant"}
 fi
 
-source /dev/stdin <<< "$(curl -s https://raw.githubusercontent.com/${MGC_ACCOUNT}/multicluster-gateway-controller/${MGC_BRANCH}/hack/.quickstartEnv)"
-source /dev/stdin <<< "$(curl -s https://raw.githubusercontent.com/${MGC_ACCOUNT}/multicluster-gateway-controller/${MGC_BRANCH}/hack/.kindUtils)"
-source /dev/stdin <<< "$(curl -s https://raw.githubusercontent.com/${MGC_ACCOUNT}/multicluster-gateway-controller/${MGC_BRANCH}/hack/.cleanupUtils)"
-source /dev/stdin <<< "$(curl -s https://raw.githubusercontent.com/${MGC_ACCOUNT}/multicluster-gateway-controller/${MGC_BRANCH}/hack/.deployUtils)"
-source /dev/stdin <<< "$(curl -s https://raw.githubusercontent.com/${MGC_ACCOUNT}/multicluster-gateway-controller/${MGC_BRANCH}/hack/.startUtils)"
-source /dev/stdin <<< "$(curl -s https://raw.githubusercontent.com/${MGC_ACCOUNT}/multicluster-gateway-controller/${MGC_BRANCH}/hack/.setupEnv)"
+if [ -n "$MGC_LOCAL_QUICKSTART_SCRIPTS_MODE" ]; then
```
jasonmadigan, line 26: Added this for us devs - makes testing local changes on disk a bit easier.

jasonmadigan:
@R-Lawton WDYT?

hack/.deployUtils
```diff
@@ -327,8 +327,10 @@ configureController() {
     kubectl config use-context kind-${clusterName}
     echo "Initialize local dev setup for the controller on ${clusterName}"
 
-  if [ -n "$MGC_AWS_ACCESS_KEY_ID" ]; then
-    cat <<EOF | kubectl apply -f -
+    if [ "$DNS_PROVIDER" == "aws" ]; then
+        if [ -n "$MGC_AWS_ACCESS_KEY_ID" ]; then
```
maleck13, line 331: adding validation here might not be bad thing

```
if [ "$MGC_AWS_ACCESS_KEY_ID" == "" ]; then
     echo "MGC_AWS_ACCESS_KEY_ID is not set"
```
etc

maleck13: Actually, perhaps a small refactor here overall would be better?

like a switch and some functions?
```
case $DNS_PROVIDER in
   aws)
      echo "setting up aws dns provider"
      awsProvider()
      ;;
   gcp)
      echo "setting up gcp dns provider"
      gcpProvider()
      ;;
   *)
     echo "unknown provider"
     exit
     ;;
esac

```

mikenairn: My 2cents, we should not be relying on environment variables anywhere. Its just a bad idea and something i mentioned we should try and avoid previously.

mikenairn: > My 2cents, we should not be relying on environment variables anywhere. Its just a bad idea and something i mentioned we should try and avoid previously.

That said, we are unlikely to change how the quick start works at this stage in the game, so what is being proposed makes sense.

R-Lawton:
current changes work locally for me but i agree with the suggested changes for Craig

hack/.deployUtils
jasonmadigan: Yeah, I'd agree with that principle too - but, like you allude to, I think we'd have to unwind most of our quickstart to do so. bash isn't great for passing args/params around generally given its nature

Have refactored configureController a bit to use a switch instead

hack/.deployUtils
```diff
@@ -406,9 +395,35 @@ spec:
...
 
+configureController() {
+    clusterName=${1}
+    kubectl config use-context kind-${clusterName}
+    echo "Initialize local dev setup for the controller on ${clusterName}"
+
+    case $DNS_PROVIDER in
+      aws)
+          echo "setting up aws dns provider"
+          if [ "$MGC_AWS_ACCESS_KEY_ID" == "" ]; then
```
maleck13, line 408: I would put this in the function tbh

maleck13 (approved):
minor change suggested.  I haven't tried this locally. Don't want to tear down env currently as fixing differnet issue

jasonmadigan:
@maleck13 fixed (agreed). Will do one more quick spin for both providers locally, then merge if all OK?

Issue #568, opened 2023-09-18
Let's Encrypt setup failing


laurafitzgerald:
### **Steps to reproduce**
Follow https://docs.kuadrant.io/getting-started/ using aws as the provider
export MGC_SUB_DOMAIN=myapp.$MGC_ZONE_ROOT_DOMAIN

Follow https://github.com/Kuadrant/multicluster-gateway-controller/blob/a4a74009517a92049e24ff738cc0f2246fc6c587/docs/tlspolicy/tls-policy.md#lets-encrypt-issuer-for-route53-hosted-domain

Create gateway
```
kubectl apply -f - <<EOF
apiVersion: gateway.networking.k8s.io/v1beta1
kind: Gateway
metadata:
  name: prod-web
  namespace: multi-cluster-gateways
spec:
  gatewayClassName: kuadrant-multi-cluster-gateway-instance-per-cluster
  listeners:
  - allowedRoutes:
      namespaces:
        from: All
    name: api
    hostname: $MGC_SUB_DOMAIN
    port: 443
    protocol: HTTPS
    tls:
      mode: Terminate
      certificateRefs:
        - name: apps-hcpapps-tls
          kind: Secret
EOF
```
Label Gateway `kubectl --context kind-mgc-control-plane label gateway prod-web "cluster.open-cluster-management.io/placement"="http-gateway" -n multi-cluster-gateways`

Create tlspolicy using Let's Encryt Issuer
```
kubectl apply -f - <<EOF
apiVersion: kuadrant.io/v1alpha1
kind: TLSPolicy
metadata:
  name: prod-web
  namespace: multi-cluster-gateways
spec:
  targetRef:
    name: prod-web
    group: gateway.networking.k8s.io
    kind: Gateway
  issuerRef:
    group: cert-manager.io
    kind: Issuer
    name: le-production
EOF
```
Create a DNSPolicy
```
kubectl apply -f - <<EOF
apiVersion: kuadrant.io/v1alpha1             
kind: DNSPolicy
metadata:
  name: prod-web
  namespace: multi-cluster-gateways
spec:
  targetRef:
    name: prod-web
    group: gateway.networking.k8s.io
    kind: Gateway     
EOF  
```

### **Results**
See that the secret generated has the name specified in the gateway and a suffix

```
kc get secret -n multi-cluster-gateways
NAME                     TYPE     DATA   AGE
apps-hcpapps-tls-lqk9w   Opaque   1      2m55s
```

[...]

Issue #507, opened 2023-09-11
TLSPolicy testing


maleck13:
**What**

Pre-req : 
- Existing environment, this can be from the quick start setup but ensure you are using the release branch

`export MGC_BRANCH=<release-branch under test>`

With an environment setup you should also have some form of placement configured with at least 1 ManagedCluster selected

Gist https://gist.github.com/maleck13/11ad810698a78c7e2012e824671d6211 

This will be there by default with the quick start.

You should also have an cert issuer present (with quick start this will be self signed issuer)

**N.B. Useful sample resources in [this guide](https://docs.kuadrant.io/multicluster-gateway-controller/docs/how-to/multicluster-gateways-walkthrough/)**
- [ ] Setup a gateway definition and label it with the placement label `cluster.open-cluster-management.io/placement: [placement name] `
- [ ] Add 3 HTTPS listeners  one wildcard and two non-wildcard and observe the gateway does not get into a programmed state
- [ ] provide a manual secret for one of the listeners (IE create a secret with the defined name)
- [ ] Define a TLSPolicy (see example below) 
- [ ] Validate it is clear from the TLSPolicy status that the policy has succeeded
- [ ] Validate that the gateway is updated with a policyAffected condition
- [ ] Observe the HTTPS secrets are created for any listener without an existing backing secret and are synced to the cluster
- [ ] Observe the HTTPS secrets are not replaced for listeners with existing manual secrets
- [ ] Validate that a wildcard cert was created and a non wildcard (should be able to do this in the browser)
- [ ] With TLSPolicy in place define some additional HTTPS listeners with named secrets, observe these are also created and synced
- [ ] With TLSPolicy in place define a HTTP listener and check no secrets created
- [ ] With TLSPolicy in place delete a HTTPS listener and observe the secrets are cleaned up and the gateway remains available
- [ ] Delete the TLSPolicy and observe only the TLS secrets created by TLSPolicy are removed (the manual one should stay)
- [ ] Set up a LetsEncrypt ClusterIssuer and re create the TLSPolicy using this issuer
- [ ] Observe that we have LE certs secrets and they synced
- [ ] Validate the HTTPs via the browser or curl
- [ ] Define an invalid TLSPolicy IE that references a Issuer that doesn't exists, validate that it is clear from the status what the issue is
- [ ] Define an invalid TLSPolicy IE that references a none existent gateway validate that it is clear from the status what the issue is

**Example TLSPolicy**

```
[...]

Issue #382, opened 2023-07-27
Investigate removing the ManagedZone API


maleck13:

maleck13:
https://github.com/orgs/Kuadrant/discussions/7

mikenairn:
POC removing ManagedZones can be found here https://github.com/mikenairn/multi-cluster-traffic-controller/commit/3aa68f8ebef54fcb48c955e3cb8487dbe716d733

The ManagedZones API can be removed fairly easily in favour of listing zones from the current provider. In the POC i completely removed the ManagedZone API but we might want to consider leaving it as it may be useful for testing or more generally to have the ability to easily create new zones as a completely separate piece of functionality.

Main things that needed changed:

DNSPolicy:

```.yaml
apiVersion: kuadrant.io/v1alpha1
kind: DNSPolicy
metadata:
  name: prod-web
  namespace: multi-cluster-gateways
spec:
  targetRef:
    name: prod-web
    group: gateway.networking.k8s.io
    kind: Gateway
  provider:
    secretRef:
      name: mgc-aws-credentials
      namespace: multi-cluster-gateways
    config:
      zoneIDFilter:
        - Z04114632NOABXYWH93QUl
```

* Add provider.secrefRef in order to attach provider credentials directly
* Added provider.config to add provider configuration such as `zoneIDilter` to limit what zones we want to work with from the provider. (More could be added here)

It might be worth considering coming back round to the idea of a separate DNSProvider API if adding more config to a provider is something we want to do. It would also tie in with what cert manager does re. Issuers.

DNSRecord:

```.yaml
apiVersion: kuadrant.io/v1alpha1
kind: DNSRecord
metadata:
  name: myapp.mn.hcpapps.net
  namespace: multi-cluster-gateways
spec:
  provider:
    secretRef:
      name: mgc-aws-credentials
      namespace: multi-cluster-gateways
    config:
      zoneIDFilter:
        - Z04114632NOABXYWH93QU
  zoneId: Z04114632NOABXYWH93QU
  endpoints:
  - dnsName: test1.myapp.mn.hcpapps.net
    recordTTL: 60
    recordType: A
    targets:
    - 172.32.200.0
```

* Add provider field. Copied from the dnspolicy when it creates records.
* Add zoneId field, the zone id selected to publish into. (This shouldn't change during dnspolicy reconcile after initial selection)

Note: Moving everything to a DNSProver resource might look better here rather than having to copy the entire thing to the record.

[...]

PR #572, opened 2023-09-19
docs updates

laurafitzgerald:
Small docs updates around the creation of the issuer for the Let'sEncrypt setup.

Update relates to https://github.com/Kuadrant/multicluster-gateway-controller/issues/507 and https://github.com/Kuadrant/multicluster-gateway-controller/issues/568

### DIFF
--- a/docs/tlspolicy/tls-policy.md
+++ b/docs/tlspolicy/tls-policy.md
@@ -130,7 +130,7 @@ Any type of Issuer that is supported by CertManager can be referenced in the TLS
 
 Create a secret containing AWS access key and secret:
 ```bash
-kubectl create secret generic mgc-aws-credentials --from-literal=AWS_ACCESS_KEY_ID=<AWS_ACCESS_KEY_ID> --from-literal=AWS_SECRET_ACCESS_KEY=<AWS_SECRET_ACCESS_KEY> -n multi-cluster-gateways
+kubectl create secret generic le-aws-credentials --from-literal=AWS_SECRET_ACCESS_KEY=<AWS_SECRET_ACCESS_KEY> -n multi-cluster-gateways
 ```
 
 Create a new Issuer:
@@ -139,6 +139,7 @@ apiVersion: cert-manager.io/v1
 kind: Issuer
 metadata:
   name: le-production
+  namespace: multi-cluster-gateways
 spec:
   acme:
     email: <YOUR EMAIL>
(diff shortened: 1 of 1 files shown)


david-martin:
/lgtm
This makes sense

docs/tlspolicy/tls-policy.md
```diff
@@ -151,10 +152,10 @@ spec:
           route53:
             hostedZoneID: <YOUR HOSTED ZONE ID>
             region: us-east-1
-            accessKeyID: <AWS_SECRET_ACCESS_KEY>
+            accessKeyID: <AWS_SECRET_ACCESS_KEY_ID>
```
jasonmadigan, line 155: This looked a little strange to me (i.e. one part of AWS creds inline, one a secret)

Not for now, but I saw the `Issuer` docs for cert-manager mention you can also create an `accessKeyIDSecretRef`:

```
accessKeyIDSecretRef:
  name: prod-route53-credentials-secret
  key: access-key-id
```

We could probably swap to use this later on (or maybe support both)

cc: @mikenairn

jasonmadigan: I am fine with this change landing as-is btw

mikenairn: Yes, you should be able to do it that way, but for some reason it didn't work with the version of the CRDs we were installing. Meant to look into why as it would be better that way, and would allow us to use the same secret as we use for dns [see](https://github.com/Kuadrant/multicluster-gateway-controller/issues/382#issuecomment-1661872837.)

mikenairn: I think the changes here look fine though, can look into why that didn't work separately.

mikenairn: Example of the error i was getting ```Error from server (BadRequest): error when creating "scratch/kuadrant_tlspolicy/issuer-le-staging.yaml": Issuer in version "v1" cannot be handled as a Issuer: strict decoding error: unknown field "spec.acme.solvers[0].dns01.route53.accessKeyIDSecretRef" ```

laurafitzgerald: So we're good to keep this as is for now?

EDIT: although i should note that this change doesn't resolve the full issue of getting Let'sEncrypt setup and working correctly with MGC. Still working on that.

mikenairn: Opened a PR to update this https://github.com/Kuadrant/multicluster-gateway-controller/pull/584

PR #582, opened 2023-09-20
Update cert manager crds

mikenairn:
**Remove cached helm charts during cleanup**
Removes all cached helm chart directories from config during cleanup (make clean or make local-cleanup).
There is a bug in kustomize where the helm chart chached locally is never updated when the version of the chart in the kustomization is changed. This could potentially lead to different versions being used locally in development and running the quick start.

**Update cached crd versions (v1.7.1 -> 1.12.2)**
Update the cached cert-manager crds to the correct version (1.12.2) and updated integration tests to use this version of the crds.
Adds a basic script that can be used to download the desired crds (config/cert-manager/update-crds.sh).

### DIFF
--- a/test/integration/suite_test.go
+++ b/test/integration/suite_test.go
@@ -85,7 +85,7 @@ var _ = BeforeSuite(func() {
 		CRDDirectoryPaths: []string{
 			filepath.Join("../../", "config", "crd", "bases"),
 			filepath.Join("../../", "config", "gateway-api", "crd", "standard"),
-			filepath.Join("../../", "config", "cert-manager", "crd", "v1.7.1"),
+			filepath.Join("../../", "config", "cert-manager", "crd", "v1.12.2"),
 			filepath.Join("../../", "config", "ocm", "crd"),
 		},
 		ErrorIfCRDPathMissing: true,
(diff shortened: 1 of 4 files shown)


test/integration/suite_test.go
```diff
@@ -85,7 +85,7 @@ var _ = BeforeSuite(func() {
 		CRDDirectoryPaths: []string{
 			filepath.Join("../../", "config", "crd", "bases"),
 			filepath.Join("../../", "config", "gateway-api", "crd", "standard"),
-			filepath.Join("../../", "config", "cert-manager", "crd", "v1.7.1"),
+			filepath.Join("../../", "config", "cert-manager", "crd", "v1.12.2"),
```
mikenairn, line 88: @david-martin Do you think there is any advantage to keeping version of these crds rather than just creating a "latest" directory and keeping that up to date?

david-martin: I can't think of a reason to keep older versions.

mikenairn: Decided to rename the directory so we should never have to update these tests again

maleck13 (approved):

PR #584, opened 2023-09-20
Update tls docs

mikenairn:
Update the LE Issuer example to use reference all AWS credentials from the secret. The version of cert-manager that is installed by default in local development(v1.12.2) should have these options.


mikenairn:
The issue i was seeing with this was caused by the wrong version of cert manager being installed locally see https://github.com/Kuadrant/multicluster-gateway-controller/pull/582

jasonmadigan:
Nice one

/approve
/lgtm

jasonmadigan (approved):

jasonmadigan:
Manually merging. Will trigger a docs redeploy.

jasonmadigan:
@laurafitzgerald ^ fyi

Issue #574, opened 2023-09-19
Kuadrant addon does not work on OpenShift by default


pehala:
Kuadrant addon relies on the existence of `olm/operatorhubio-catalog` CatalogSource, which is installed by default if you follow https://olm.operatorframework.io/docs/getting-started/. On OpenShift the OLM is by default and so there is no need to follow this guide, however, it does not have the aforementioned CatalogSource, for community operators it uses
 `openshift-marketplace/community-operators`.

Workaround for this is manually creating the CatalogSource that the addon expects by copying the one OpenShift uses.

maleck13:
@david-martin @jasonmadigan I think the solution to this one for now, is to call out in the install doc that you need to do the above in an OpenShift cluster

PR #593, opened 2023-09-21
#574 - addon catalogsource for OpenShift installs

jasonmadigan:
Updated docs with a note around CatalogSource for OpenShift installs of the Kuadrant addon

Workaround for #574


maleck13 (commented):
small change otherwise looks good

PR #594, opened 2023-09-21
validate the presence of specified issuer or clusterisssuer in tlspolicy

### LINKED ISSUE #581: Incorrect status reported in tlspolicy when issuerRef points to an issuer that doesn't exist
**Steps to reproduce non issuer**
Follow https://docs.kuadrant.io/getting-started/
Follow https://docs.kuadrant.io/multicluster-gateway-controller/docs/how-to/multicluster-gateways-walkthrough/ until the creation of the TLSPolicy
Apply a tlspolicy that points to an issuer that doesn't exist
```
spec:
  issuerRef:
    group: cert-manager.io
    kind: Issuer
    name: something-else
```

Observe status
```
 status:
    conditions:
    - lastTransitionTime: "2023-09-20T09:04:40Z"
      message: Gateway is TLS Enabled
      reason: GatewayTLSEnabled
      status: "True"
      type: Ready
    observedGeneration: 2
```

laurafitzgerald:
# What

Closes #581 

Verification 
Follow https://docs.kuadrant.io/getting-started/
Follow https://docs.kuadrant.io/multicluster-gateway-controller/docs/how-to/multicluster-gateways-walkthrough/ until the creation of the TLSPolicy

### Error should be reported in status for ClusterIssuer that doesn't exist in cluster
Create TLSPolicy pointing to ClusterIssuer with a name that doesn't exist in the cluster
Validate that error is present in the status of the TLSPolicy
```
 - lastTransitionTime: "2023-09-21T10:13:13Z"
      message: ClusterIssuer.cert-manager.io "blah" not found
      reason: ReconciliationError
      status: "False"
      type: Ready
```
### Error should be reported in status for Issuer that doesn't exist in the namespace of the tlspolicy

Create an issuer in a namespace that is different to the TLSPolicy - steps here to do that https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/docs/tlspolicy/tls-policy.md#lets-encrypt-issuer-for-route53-hosted-domain - update the namespace in the issuer in the steps to another namespace in the cluster

Validate that error is present in the status of the TLSPolicy
```
- lastTransitionTime: "2023-09-21T10:13:13Z"
      message: Issuer.cert-manager.io "le-staging" not found
      reason: ReconciliationError
      status: "False"
      type: Ready
```

### Error should NOT be reported in status for Issuer that does exist in the same namespace as the tlspolicy

Create the same issuer but in the same namespace as the tlspolicy e.g. multi-cluster-gateways
Validate the status is good
```
    - lastTransitionTime: "2023-09-21T10:21:11Z"
      message: Gateway is TLS Enabled
      reason: GatewayTLSEnabled
      status: "True"
      type: Ready
```

### Error should NOT be reported in status for ClusterIssuer that does exist in the same namespace as the tlspolicy

Update the policy to have a valid ClusterIssuer glbc
e.g. 
```
issuerRef:
      group: cert-manager.io
      kind: ClusterIssuer
      name: glbc-ca
```
Validate the status is good
```
    - lastTransitionTime: "2023-09-21T10:21:11Z"
      message: Gateway is TLS Enabled
      reason: GatewayTLSEnabled
      status: "True"
      type: Ready
```

### Error should NOT be reported in status for Issuer that does not specify kind
Update the policy to have the following issuerRef - create the issuer in the same namespace if needed
e.g. 
```
issuerRef:
      group: cert-manager.io
      name: le-staging
```
Validate the status is good
```
    - lastTransitionTime: "2023-09-21T10:21:11Z"
      message: Gateway is TLS Enabled
      reason: GatewayTLSEnabled
      status: "True"
      type: Ready
```

### DIFF
--- a/pkg/controllers/tlspolicy/tlspolicy_controller.go
+++ b/pkg/controllers/tlspolicy/tlspolicy_controller.go
@@ -65,6 +65,8 @@ type TLSPolicyReconciler struct {
 //+kubebuilder:rbac:groups=kuadrant.io,resources=tlspolicies,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=kuadrant.io,resources=tlspolicies/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=kuadrant.io,resources=tlspolicies/finalizers,verbs=update
+//+kubebuilder:rbac:groups="cert-manager.io",resources=issuers,verbs=get;list;
+//+kubebuilder:rbac:groups="cert-manager.io",resources=clusterissuers,verbs=get;list;
 
 func (r *TLSPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	log := r.Logger().WithValues("TLSPolicy", req.NamespacedName)
@@ -147,6 +149,11 @@ func (r *TLSPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 func (r *TLSPolicyReconciler) reconcileResources(ctx context.Context, tlsPolicy *v1alpha1.TLSPolicy, targetNetworkObject client.Object) error {
 	// validate
 	err := tlsPolicy.Validate()
+	if err != nil {
+		return err
+	}
+
+	err = validateIssuer(ctx, r.Client(), tlsPolicy)
 	if err != nil {
 		return err
 	}
--- a/test/e2e/gateway_single_spoke_test.go
+++ b/test/e2e/gateway_single_spoke_test.go
@@ -87,7 +87,7 @@ var _ = Describe("Gateway single target cluster", func() {
 		err = tconfig.HubClient().Create(ctx, gw)
 		Expect(err).ToNot(HaveOccurred())
 
-		By("setting up  TLSPolicy in the hub")
+		By("setting up TLSPolicy in the hub")
 		tlsPolicy = &mgcv1alpha1.TLSPolicy{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      testID,
--- a/test/integration/tlspolicy_controller_test.go
+++ b/test/integration/tlspolicy_controller_test.go
@@ -27,6 +27,7 @@ var _ = Describe("TLSPolicy", Ordered, func() {
 
 	var testNamespace string
 	var gatewayClass *gatewayv1beta1.GatewayClass
+	var issuer *certmanv1.Issuer
 
 	BeforeAll(func() {
 		logger = zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true))
@@ -42,6 +43,11 @@ var _ = Describe("TLSPolicy", Ordered, func() {
 
 	BeforeEach(func() {
 		CreateNamespace(&testNamespace)
+		issuer = NewTestIssuer("testissuer", testNamespace)
+		Expect(k8sClient.Create(ctx, issuer)).To(BeNil())
+		Eventually(func() error { //issuer exists
+			return k8sClient.Get(ctx, client.ObjectKey{Name: issuer.Name, Namespace: issuer.Namespace}, issuer)
+		}, TestTimeoutMedium, TestRetryIntervalMedium).ShouldNot(HaveOccurred())
 	})
 
 	AfterEach(func() {
@@ -54,6 +60,11 @@ var _ = Describe("TLSPolicy", Ordered, func() {
 		Expect(k8sClient.List(ctx, &policyList)).To(BeNil())
 		for _, policy := range policyList.Items {
 			k8sClient.Delete(ctx, &policy)
+		}
+		issuerList := certmanv1.IssuerList{}
+		Expect(k8sClient.List(ctx, &issuerList)).To(BeNil())
+		for _, issuer := range issuerList.Items {
+			k8sClient.Delete(ctx, &issuer)
 		}
 	})
 
@@ -74,7 +85,7 @@ var _ = Describe("TLSPolicy", Ordered, func() {
 			Expect(err).ToNot(HaveOccurred())
 		})
 
-		Context("valid target and policy", func() {
+		Context("valid target, issuer and policy", func() {
 
 			BeforeEach(func() {
 				gateway = NewTestGateway("test-gateway", gwClassName, testNamespace).
@@ -84,7 +95,8 @@ var _ = Describe("TLSPolicy", Ordered, func() {
 					return k8sClient.Get(ctx, client.ObjectKey{Name: gateway.Name, Namespace: gateway.Namespace}, gateway)
 				}, TestTimeoutMedium, TestRetryIntervalMedium).ShouldNot(HaveOccurred())
 				tlsPolicy = NewTestTLSPolicy("test-tls-policy", testNamespace).
-					WithTargetGateway(gateway.Name).TLSPolicy
+					WithTargetGateway(gateway.Name).
+					WithIssuer("testissuer", certmanv1.IssuerKind, "cert-manager.io").TLSPolicy
 				Expect(k8sClient.Create(ctx, tlsPolicy)).To(BeNil())
 				Eventually(func() error { //tls policy exists
 					return k8sClient.Get(ctx, client.ObjectKey{Name: tlsPolicy.Name, Namespace: tlsPolicy.Namespace}, tlsPolicy)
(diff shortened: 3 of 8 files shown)


pkg/controllers/tlspolicy/tlspolicy_controller.go
```diff
@@ -151,6 +152,24 @@ func (r *TLSPolicyReconciler) reconcileResources(ctx context.Context, tlsPolicy
...
+		}
+	}
+	if issuerKind == v1.IssuerKind {
+		issuer := &v1.Issuer{}
+		err = r.Client().Get(ctx, client.ObjectKey{Name: issuerName, Namespace: tlsPolicy.Namespace}, issuer)
+		if err != nil {
+			return err
+		}
+	}
+
```
sergioifg94, line 172:
```suggestion
	// validate that the issuer specified exists
	issuerName := tlsPolicy.Spec.IssuerRef.Name
	issuerKind := tlsPolicy.Spec.IssuerRef.Kind
	var issuer client.Object
	issuerNamespace := ""
	if issuerKind == v1.ClusterIssuerKind {
		issuer = &v1.ClusterIssuer{}
	} else if issuerKind == v1.IssuerKind || issuerKind == "" {
		issuer = &v1.Issuer{}
		issuerNamespace = tlsPolicy.Namespace
	} else {
		return fmt.Errorf("issuer kind not supported: %s. Must be either ClusterIssuer or Issuer", issuerKind)
	}

	err = r.Client().Get(ctx, client.ObjectKey{Name: issuerName, Namespace: issuerNamespace}, issuer)
	if err != nil {
		return err
	}

```

I saw in the documentation of `IssuerRef.Kind` that if the value is left empty it's treated as `Issuer`. Also added an extra `else` for the case of any other invalid values, and a tiny bit of refactoring

laurafitzgerald: Nice. Pushed. 👍

pkg/controllers/tlspolicy/tlspolicy_controller.go
```diff
@@ -21,6 +21,7 @@ import (
 	"fmt"
 
 	"github.com/go-logr/logr"
+	v1 "github.com/jetstack/cert-manager/pkg/apis/certmanager/v1"
```
mikenairn, line 24:
```suggestion
	certmanv1 "github.com/jetstack/cert-manager/pkg/apis/certmanager/v1"
```

Avoids confusion with our own v1 api, and consistent with other places we import it https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/controllers/tlspolicy/tlspolicy_certmanager_certificates.go#L8

pkg/controllers/tlspolicy/tlspolicy_controller.go
```diff
@@ -151,6 +152,25 @@ func (r *TLSPolicyReconciler) reconcileResources(ctx context.Context, tlsPolicy
...
+		issuer = &v1.Issuer{}
+		issuerNamespace = tlsPolicy.Namespace
+	} else {
+		return fmt.Errorf("issuer kind not supported: %s. Must be either ClusterIssuer or Issuer", issuerKind)
+	}
+
+	err = r.Client().Get(ctx, client.ObjectKey{Name: issuerName, Namespace: issuerNamespace}, issuer)
+	if err != nil {
+		return err
+	}
```
mikenairn, line 172: Looks fine, you might want consider moving into it's own method, presumably we might eventually want to call the same thing from a webhook re. https://github.com/Kuadrant/multicluster-gateway-controller/issues/595

You also might want to double check what the cert-manager code looks like where it loads the Issuer,  it must have similar logic around the Kind etc..

test/e2e/gateway_single_spoke_test.go
```diff
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	v1 "github.com/jetstack/cert-manager/pkg/apis/certmanager/v1"
```
mikenairn, line 14: import name here also v1 -> certmanv1

test/integration/tlspolicy_controller_test.go
```diff
@@ -84,7 +95,7 @@ var _ = Describe("TLSPolicy", Ordered, func() {
 					return k8sClient.Get(ctx, client.ObjectKey{Name: gateway.Name, Namespace: gateway.Namespace}, gateway)
 				}, TestTimeoutMedium, TestRetryIntervalMedium).ShouldNot(HaveOccurred())
 				tlsPolicy = NewTestTLSPolicy("test-tls-policy", testNamespace).
-					WithTargetGateway(gateway.Name).TLSPolicy
+					WithTargetGateway(gateway.Name).WithIssuer("testissuer", certmanv1.IssuerKind, "cert-manager.io").TLSPolicy
```
mikenairn, line 98: nit: Subjective, but when using a builder pattern like this i think it looks a bit neater, and easier to understand, when each piece is on a new line:

```
tlsPolicy = NewTestTLSPolicy("test-tls-policy", testNamespace).
					WithTargetGateway(gateway.Name).
					WithIssuer("testissuer", certmanv1.IssuerKind, "cert-manager.io").TLSPolicy
```

test/util/suite_config.go
```diff
@@ -10,6 +10,7 @@ import (
 	"strconv"
 
 	"github.com/goombaio/namegenerator"
+	v1 "github.com/jetstack/cert-manager/pkg/apis/certmanager/v1"
```
mikenairn, line 13: import name here also v1 -> certmanv1

test/util/test_types.go
```diff
@@ -3,6 +3,7 @@
 package testutil
 
 import (
+	v1 "github.com/jetstack/cert-manager/pkg/apis/certmanager/v1"
```
mikenairn, line 6: import name here also v1 -> certmanv1

pkg/controllers/tlspolicy/tlspolicy_controller.go
mikenairn: Does make me wonder if this is the right place for this validation :thinking:

laurafitzgerald: I moved it to a certmanager_helper file.. lmk if you think that's suitable.

laurafitzgerald:
the e2e tests are passing locally for me. so i'm looking into that.

pkg/controllers/tlspolicy/tlspolicy_controller.go
laurafitzgerald: The logic is here for cert-manager https://github.com/cert-manager/cert-manager/blob/master/pkg/issuer/helper.go#L57-L74 
I can refactor what we have to be closer. There's a bit in setting up the helper in order to use that function itself but can take a look if you think it's of value?

laurafitzgerald: Pushed up a refactor for now.

mikenairn: Hmm, looking at that has made me realise that this check isn't going to work unless we have permission to get Issuers and ClusterIssuers. We can obviously add that [see](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/controllers/gateway/gateway_controller.go#L81-L87), but all we are really testing is that the mgc controller can get them, not that cert manager can. Presumably we end up in the same situation if cert manager was configured incorrectly for whatever reason.

In an ideal world, we wouldn't need to do any of this and CertManager itself would add a status to the Certificates when the issuer isn't found/accessible and we aggregate the status back from all the Certificates into the policy status :thinking:

pkg/controllers/tlspolicy/tlspolicy_controller.go
```diff
@@ -65,6 +65,8 @@ type TLSPolicyReconciler struct {
 //+kubebuilder:rbac:groups=kuadrant.io,resources=tlspolicies,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=kuadrant.io,resources=tlspolicies/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=kuadrant.io,resources=tlspolicies/finalizers,verbs=update
+// +kubebuilder:rbac:groups="cert-manager.io",resources=issuers,verbs=get;list;
+// +kubebuilder:rbac:groups="cert-manager.io",resources=clusterissuers,verbs=get;list;
```
mikenairn, line 69:
```suggestion
//+kubebuilder:rbac:groups="cert-manager.io",resources=issuers,verbs=get;list;
//+kubebuilder:rbac:groups="cert-manager.io",resources=clusterissuers,verbs=get;list;
```

test/integration/tlspolicy_controller_test.go
```diff
@@ -42,6 +43,11 @@ var _ = Describe("TLSPolicy", Ordered, func() {
 
 	BeforeEach(func() {
 		CreateNamespace(&testNamespace)
+		issuer = NewTestIssuer("testissuer", testNamespace)
+		Expect(k8sClient.Create(ctx, issuer)).To(BeNil())
+		Eventually(func() error { //issuer exists
+			return k8sClient.Get(ctx, client.ObjectKey{Name: issuer.Name, Namespace: issuer.Namespace}, issuer)
+		}, TestTimeoutMedium, TestRetryIntervalMedium).ShouldNot(HaveOccurred())
```
mikenairn, line 50: Since we now have logic around the types Issuers in our controller, would it be worth creating a ClusterIssuer also and adding a test that uses that?

mikenairn:
Couple of comments, but looks fine i think.

It did make me wonder if this is the best way to do this kind of thing, ideally we wouldn't be doing validation that is expected to be done by cert manager (the thing this policy is wrapping), but rather rely on aggregating the status of created cert manager resources (Certificates), or expected generated resources (CertificateRequests), since this would be a truer reflection of the state of the policy.

Example CertificateRequests status(cert manager 1.13.0), that we could get this info from:

```
kubectl get certificaterequests mn-hcpapps-net-tls-1 -n single-cluster-gateways -o json | jq '.status.conditions[] | select(.type == "Ready")'
{
  "lastTransitionTime": "2023-09-22T07:42:03Z",
  "message": "Referenced \"ClusterIssuer\" not found: clusterissuer.cert-manager.io \"glbc-caaaaa\" not found",
  "reason": "Pending",
  "status": "False",
  "type": "Ready"
}
```

Since we aren't going to do that aggregation now, and will likely be done in line with the proposed [kuadrant policy status/conditions](https://github.com/Kuadrant/architecture/pull/9), a basic check like this is probably fine to avoid unnecessary confusion in the meantime.

PR #596, opened 2023-09-21
update tests for HTTP expectations with TLS Policy

### LINKED ISSUE #578: Fix tests for TLSPolicy related to HTTP endpoints
### **Steps to Reproduce**

Follow https://docs.kuadrant.io/getting-started/ with the provider of your choice

Create a gateway with one listener that is HTTP protocol
```
kubectl apply -f - <<EOF                                                                                          
apiVersion: gateway.networking.k8s.io/v1beta1
kind: Gateway
metadata:
  name: prod-web
spec:
  gatewayClassName: kuadrant-multi-cluster-gateway-instance-per-cluster
  listeners:
  - allowedRoutes:
      namespaces:
        from: All
    name: api
    hostname: $MGC_SUB_DOMAIN
    port: 443
    protocol: HTTP
    tls:
      mode: Terminate
      certificateRefs:
        - name: apps-hcpapps-tls
          kind: Secret
EOF
```

Label gateway with `kubectl label gateway prod-web -n multi-cluster-gateways cluster.open-cluster-management.io/placement=http-gateway`

Create a TLSPolicy 
glbc is fine for this. 
```
kubectl create -f - << EOF
kind: TLSPolicy
metadata:
  name: prod-web
spec:
  targetRef:
    name: prod-web
    kind: Gateway
  issuerRef:
    group: cert-manager.io
    kind: ClusterIssuer
    name: glbc-ca
EOF
```

Secret is created when it shouldn't
```
kc get secret -n multi-cluster-gateways 
NAME                 TYPE                DATA   AGE
apps-hcpapps-tls     kubernetes.io/tls   3      6s
```

philbrookes:
closes #578 

Update TLS Policy HTTP tests, to expect certificate if TLS section is present.

### DIFF
--- a/test/integration/tlspolicy_controller_test.go
+++ b/test/integration/tlspolicy_controller_test.go
@@ -161,13 +161,32 @@ var _ = Describe("TLSPolicy", Ordered, func() {
 				}, TestTimeoutMedium, TestRetryIntervalMedium).ShouldNot(HaveOccurred())
 			})
 
-			It("should not create any certificates", func() {
+			It("should not create any certificates when TLS is not present", func() {
 				Consistently(func() []certmanv1.Certificate {
 					certList := &certmanv1.CertificateList{}
 					err := k8sClient.List(ctx, certList, &client.ListOptions{Namespace: testNamespace})
 					Expect(err).ToNot(HaveOccurred())
 					return certList.Items
 				}, time.Second*10, time.Second).Should(BeEmpty())
+			})
+
+			It("should create certificate when TLS is present", func() {
+				certNS := gatewayv1beta1.Namespace(testNamespace)
+				patch := client.MergeFrom(gateway.DeepCopy())
+				gateway.Spec.Listeners[0].TLS = &gatewayv1beta1.GatewayTLSConfig{
+					Mode: Pointer(gatewayv1beta1.TLSModeTerminate),
+					CertificateRefs: []gatewayv1beta1.SecretObjectReference{
+						{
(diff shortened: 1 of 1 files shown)


test/integration/tlspolicy_controller_test.go
```diff
@@ -170,6 +170,32 @@ var _ = Describe("TLSPolicy", Ordered, func() {
...
+					certList := &certmanv1.CertificateList{}
+					err := k8sClient.List(ctx, certList, &client.ListOptions{Namespace: testNamespace})
+					Expect(err).ToNot(HaveOccurred())
+					for _, cert := range certList.Items {
+						if cert.Name == "test-tls-secret" {
+							return true
+						}
+					}
+					return false
+				}, time.Second*10, time.Second).Should(BeTrue())
```
mikenairn, line 196: Most of these Eventually/Consistently steps now return an `error` instead of a `bool`. This is to make it easier to see what the error is when/if it fails.

Also, since this is just expecting a cert to exist with a particular name, could it not be re-written as:

```
Eventually(func() error {
	cert := &certmanv1.Certificate{}
	return k8sClient.Get(ctx, client.ObjectKey{Name: "test-tls-secret", Namespace: testNamespace}, cert)
}, time.Second*10, time.Second).Should(BeNil())
```

philbrookes: yeah, good point, I did it the other way as I wasn't certain what the name was going to be, but now it is known your suggestion is much tidier. I'll update it

PR #600, opened 2023-09-25
allow source ns to be overridden

maleck13:
kubectl annotate managedclusteraddon kuadrant-addon "addon.open-cluster-management.io/values"='{"CatalogSource":"openshift-marketplace/community-operators", "CatalogSourceNS":"somens"}' -n local-cluster


jasonmadigan:
We probably want to mention this in https://docs.kuadrant.io/multicluster-gateway-controller/docs/installation/service-protection-installation/, but that shouldn't hold this up.

PR #604, opened 2023-09-26
gateway.networking.k8s.io/policy should be a label

jasonmadigan:
`gateway.networking.k8s.io/policy` should be a label, rather than an annotation:

See: https://gateway-api.sigs.k8s.io/geps/gep-713/#kubectl-plugin

Noticed while checking out `gwctl` (should be able to use it after this). 

I think RLP and AuthPolicy may be missing this label.


jasonmadigan:
I guess we should apply to the probes + zones

maleck13 (approved):

PR #607, opened 2023-09-26
Update policy status on TargetNotFound

### LINKED ISSUE #361: DNSPolicy and TLSPolicy should reflect missing target in ready condition
**What**

Update policy reconcilers (DNSPolicy and TLSPolicy) to better handle the case when a target gateway has been deleted or never existed. Currently the cleanp expected in the policy is executed, but we also return the [error](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/controllers/dnspolicy/dnspolicy_controller.go#L98) causing errors in the logs:

```
2023-07-21T11:15:50+01:00       ERROR   Reconciler error        {"controller": "dnspolicy", "controllerGroup": "kuadrant.io", "controllerKind": "DNSPolicy", "DNSPolicy": {"name":"prod-web","namespace":"multi-cluster-gateways"}, "namespace": "multi-cluster-gateways", "name": "prod-web", "reconcileID": "d1b39683-8e62-4731-961c-11e6939595f0", "error": "Gateway.gateway.networking.k8s.io \"prod-web\" not found"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
        /home/mnairn/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.4/pkg/internal/controller/controller.go:329
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
        /home/mnairn/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.4/pkg/internal/controller/controller.go:274
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
        /home/mnairn/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.4/pkg/internal/controller/controller.go:235
``` 
This should be updated to instead reconcile the status of the policy and update a condition with some information about the missing target, and not return an error.

**Steps to reproduce non existent gateway for TLSPolicy**
Follow https://docs.kuadrant.io/getting-started/
Follow https://docs.kuadrant.io/multicluster-gateway-controller/docs/how-to/multicluster-gateways-walkthrough/ until the creation of the TLSPolicy
Apply a tlspolicy that points to a gateway that doesn't exist
```
spec:
    targetRef:
      kind: Gateway
      name: prod-web-non-existent
```

Note that no status is updated
`kc get tlspolicy -A -o yaml`
Note the following log in the controller
```
2023-09-20T08:40:23Z	ERROR	Reconciler error	{"controller": "tlspolicy", "controllerGroup": "kuadrant.io", "controllerKind": "TLSPolicy", "TLSPolicy": {"name":"prod-web","namespace":"multi-cluster-gateways"}, "namespace": "multi-cluster-gateways", "name": "prod-web", "reconcileID": "6d26a714-8dd4-4191-8882-76825fdb1b45", "error": "Gateway.gateway.networking.k8s.io \"prod-web-non-existent\" not found"}

```

mikenairn:
Updates the policy status (DNSPolicy and TLSPolicy) when the target resource is not found. The Ready condition is set to false with the reason "TargetNotFound" and the message is set to the appropriate error detailing the missing object.

```
{
  "lastTransitionTime": "2023-09-26T17:20:17Z",
  "message": "target not found : Gateway.gateway.networking.k8s.io \"prod-web-istio\" not found",
  "reason": "TargetNotFound",
  "status": "False",
  "type": "Ready"
}
```

closes #361

### DIFF
--- a/pkg/controllers/dnspolicy/dnspolicy_controller.go
+++ b/pkg/controllers/dnspolicy/dnspolicy_controller.go
@@ -93,8 +93,11 @@ func (r *DNSPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		if !markedForDeletion {
 			if apierrors.IsNotFound(err) {
 				log.V(3).Info("Network object not found. Cleaning up")
-				err := r.deleteResources(ctx, dnsPolicy, nil)
-				return ctrl.Result{}, err
+				delResErr := r.deleteResources(ctx, dnsPolicy, nil)
+				if delResErr == nil {
+					delResErr = err
+				}
+				return r.reconcileStatus(ctx, dnsPolicy, fmt.Errorf("%w : %w", conditions.ErrTargetNotFound, delResErr))
 			}
 			return ctrl.Result{}, err
 		}
@@ -126,25 +129,13 @@ func (r *DNSPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 
 	specErr := r.reconcileResources(ctx, dnsPolicy, targetNetworkObject)
 
-	newStatus := r.calculateStatus(dnsPolicy, specErr)
-	dnsPolicy.Status = *newStatus
-
-	if !equality.Semantic.DeepEqual(previous.Status, dnsPolicy.Status) {
-		updateErr := r.Client().Status().Update(ctx, dnsPolicy)
-		if updateErr != nil {
-			// Ignore conflicts, resource might just be outdated.
-			if apierrors.IsConflict(updateErr) {
-				return ctrl.Result{Requeue: true}, nil
-			}
-			return ctrl.Result{}, updateErr
-		}
-	}
+	statusResult, statusErr := r.reconcileStatus(ctx, dnsPolicy, specErr)
 
 	if specErr != nil {
 		return ctrl.Result{}, specErr
 	}
 
-	return ctrl.Result{}, nil
+	return statusResult, statusErr
 }
 
 func (r *DNSPolicyReconciler) reconcileResources(ctx context.Context, dnsPolicy *v1alpha1.DNSPolicy, targetNetworkObject client.Object) error {
@@ -237,6 +228,28 @@ func (r *DNSPolicyReconciler) deleteResources(ctx context.Context, dnsPolicy *v1
 	return r.updateGatewayCondition(ctx, metav1.Condition{Type: string(DNSPolicyAffected)}, gatewayDiffObj)
 }
 
+func (r *DNSPolicyReconciler) reconcileStatus(ctx context.Context, dnsPolicy *v1alpha1.DNSPolicy, specErr error) (ctrl.Result, error) {
+	newStatus := r.calculateStatus(dnsPolicy, specErr)
+
+	if !equality.Semantic.DeepEqual(newStatus, dnsPolicy.Status) {
+		dnsPolicy.Status = *newStatus
+		updateErr := r.Client().Status().Update(ctx, dnsPolicy)
+		if updateErr != nil {
+			// Ignore conflicts, resource might just be outdated.
+			if apierrors.IsConflict(updateErr) {
+				return ctrl.Result{Requeue: true}, nil
+			}
+			return ctrl.Result{}, updateErr
+		}
+	}
+
+	if errors.Is(specErr, conditions.ErrTargetNotFound) {
+		return ctrl.Result{Requeue: true}, nil
+	}
+
+	return ctrl.Result{}, nil
+}
+
 func (r *DNSPolicyReconciler) calculateStatus(dnsPolicy *v1alpha1.DNSPolicy, specErr error) *v1alpha1.DNSPolicyStatus {
 	newStatus := dnsPolicy.Status.DeepCopy()
 	if specErr != nil {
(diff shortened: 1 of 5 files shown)


pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -93,8 +93,11 @@ func (r *DNSPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		if !markedForDeletion {
 			if apierrors.IsNotFound(err) {
 				log.V(3).Info("Network object not found. Cleaning up")
-				err := r.deleteResources(ctx, dnsPolicy, nil)
-				return ctrl.Result{}, err
+				delResErr := r.deleteResources(ctx, dnsPolicy, nil)
+				if delResErr == nil {
+					delResErr = err
+				}
+				return r.reconcileStatus(ctx, dnsPolicy, fmt.Errorf("%w : %w", conditions.ErrTargetNotFound, delResErr))
```
mikenairn, line 100: Similar to how kuadrant operator currently deals with reconciling the status https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/authpolicy_controller.go#L66-L70

pkg/controllers/tlspolicy/tlspolicy_controller.go
```diff
@@ -53,6 +53,8 @@ const (
 	TLSPolicyAffected            conditions.ConditionType = "kuadrant.io/TLSPolicyAffected"
 )
 
+var ErrTargetNotFound = errors.New("target not found")
```
mikenairn, line 56: remove

pkg/controllers/dnspolicy/dnspolicy_controller.go
maleck13: would it work to return the specErr here at the end? I think we want it reported but also want it requeued? Other option would be to wrap the notfound err in a targetNotFoundErr and then have the reconcile status return a requeue on specific error types?

pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -237,6 +228,24 @@ func (r *DNSPolicyReconciler) deleteResources(ctx context.Context, dnsPolicy *v1
...
+		updateErr := r.Client().Status().Update(ctx, dnsPolicy)
+		if updateErr != nil {
+			// Ignore conflicts, resource might just be outdated.
+			if apierrors.IsConflict(updateErr) {
+				return ctrl.Result{Requeue: true}, nil
+			}
+			return ctrl.Result{}, updateErr
+		}
+	}
+
```
maleck13, line 245: so here we could have a switch perhaps? that return a Result with a requeue true?

mikenairn: Presumably that would just keep trying even if a target is never created. Is that the behaviour we would want?

test/integration/dnspolicy_controller_test.go
```diff
@@ -1330,9 +1370,11 @@ var _ = Describe("DNSPolicy", Ordered, func() {
 				}
 
 				Eventually(func() error {
+					err = k8sClient.Get(ctx, client.ObjectKey{Name: gateway.Name, Namespace: gateway.Namespace}, gateway)
+					Expect(err).NotTo(HaveOccurred())
 					gateway.Spec.Listeners = append(gateway.Spec.Listeners, otherListener)
 					return k8sClient.Update(ctx, gateway)
-				}).Should(BeNil())
+				}, TestTimeoutMedium, TestRetryIntervalMedium).ShouldNot(HaveOccurred())
```
mikenairn, line 1377: This was failing ~60% of the time for me locally. I think the main issue is that the testEnv client is a fake client that updates the whole resource, including the status, when you call update. Needed to get the resource first and update straight away to avoid a conflict.

pkg/controllers/dnspolicy/dnspolicy_controller.go
maleck13: Yeah I think that is the best option for now. 
1) return an error or requeue
2) follow on to requeue policies (perhaps based on some label constructed based on the target ref) if an addEvent happens

pkg/controllers/dnspolicy/dnspolicy_controller.go
maleck13: more detailed conversation above. Will close this one

pkg/controllers/dnspolicy/dnspolicy_controller.go
philbrookes: https://github.com/Kuadrant/multicluster-gateway-controller/issues/610

pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -237,6 +228,28 @@ func (r *DNSPolicyReconciler) deleteResources(ctx context.Context, dnsPolicy *v1
...
+	if errors.Is(specErr, conditions.ErrTargetNotFound) {
+		return ctrl.Result{Requeue: true}, nil
```
maleck13, line 247: just curious does this also result in a backoff time wise or does it just put it at the end of the queue again and again? I am guessing the latter

mikenairn: I couldn't find any docs to confirm this, but from the testing i did it appear to backoff the same as if it was an error

mikenairn: Yeah, looks like it does https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/internal/controller/controller.go#L341

maleck13 (approved):

test/integration/dnspolicy_controller_test.go
philbrookes: My PR applying these changes as a patch has merged now. If you rebase this it should resolve this problem.

Issue #587, opened 2023-09-20
automate tlspolicy manual testing 


laurafitzgerald:
All of these steps should be automated - review the existing tests https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/test/integration/tlspolicy_controller_test.go

The steps below are a copy from https://github.com/Kuadrant/multicluster-gateway-controller/issues/507 with any that failed which are covered by bugs listed here marked -> https://github.com/Kuadrant/multicluster-gateway-controller/issues/507#issuecomment-1725504441

- [x] e2e - Setup a gateway definition and label it with the placement label `cluster.open-cluster-management.io/placement: [placement name] `
- [x] integration - Add 3 HTTPS listeners  one wildcard and two non-wildcard and observe the gateway does not get into a programmed state
- [x] integration - Define a TLSPolicy -> Validate it is clear from the TLSPolicy status that the policy has succeeded
- [x] integration - Validate that the gateway is updated with a policyAffected condition - #586
- [x] e2e - Observe the HTTPS secrets are created for any listener and are synced to the cluster
- [ ] e2e - Validate that a wildcard cert was created and a non wildcard (should be able to do this in the browser)
- [ ] e2e - With TLSPolicy in place define some additional HTTPS listeners with named secrets, observe these are also created and synced
- [x] integration - With TLSPolicy in place define a HTTP listener with no TLS and check no secrets created #578 
- [ ] e2e - With TLSPolicy in place delete a HTTPS listener and observe the secrets are cleaned up and the gateway remains available
- [x] integration - With TLSPolicy in place delete a HTTPS Listener -> see relevant certificates re cleaned up
- [ ] e2e - Delete the TLSPolicy and observe only the TLS secrets created by TLSPolicy are removed 
- [x] integration - Delete the TLSPolicy -> see certificate CRs removed
- [x] (the manual one should stay) #585 
- [x] Define an invalid TLSPolicy IE that references a Issuer that doesn't exists, validate that it is clear from the status what the issue is #581
- [x] Define an invalid TLSPolicy IE that references a none existent gateway validate that it is clear from the status what the issue is #361

philbrookes:
>  Setup a gateway definition and label it with the placement label cluster.open-cluster-management.io/placement: [placement name] 
[...]

PR #609, opened 2023-09-27
verify wildcard cert on http get request

laurafitzgerald:
Addresses parts of https://github.com/Kuadrant/multicluster-gateway-controller/issues/587

Adding e2e tests to cover the e2e cases listed in the issue.


laurafitzgerald:
I've re-run the tests a number of times since adding a couple of debug logs and they've passed each time. I'd like to understand the flake but can't seem to get it to fail again. I can leave the additional debug logs in for now in case it crops up again but if you are okay with that I think it's food for review now @philbrookes @mikenairn if you have some time.

philbrookes (approved):

PR #618, opened 2023-10-02
make service protection an option

maleck13:
For testing purposes it can be useful to trigger the install of kuadrant service protection manually. To support this I have added a var `MGC_DEPLOY_SERVICE_PROTECTION` it default to true and will install service protection by default. If false it will still install OLM but wont create a managedaddon resource to install kuadrant in the spoke clusters

**Verfication**
set the following envars on top of the normal quickstart

export MGC_BRANCH=optional-service-protection
export MGC_DEPLOY_SERVICE_PROTECTION=false

run the quick start.

verify the kuadrant-system namespace is not created and no `managedclusteraddon` resources are created. 

install kuadrant service protection manually

create a catalogsource in the hub cluster

```
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: kuadrant-operator
  namespace: olm
spec:
  displayName: Kuadrant
  grpcPodConfig:
    securityContextConfig: restricted
  image: quay.io/kuadrant/kuadrant-operator-catalog:0.4.0-rc1
  sourceType: grpc
  updateStrategy:
    registryPoll:
      interval: 60m
```

create a managedclusteraddon resource in the managed cluster namespace

```
apiVersion: addon.open-cluster-management.io/v1alpha1
kind: ManagedClusterAddOn
metadata:
  annotations:
    addon.open-cluster-management.io/values: '{"CatalogSource":"kuadrant-operator", "Channel":"preview"}'
  name: kuadrant-addon
  namespace: kind-mgc-control-plane
spec:
  installNamespace: open-cluster-management-agent-addon
```

you should see kuadrant is now installed using the preview channel and new catalog source


jasonmadigan:
👀

jasonmadigan:
one suggestion to fix an error

hack/quickstart-setup.sh
```diff
@@ -76,6 +76,10 @@ fi
 if [[ -z "${MGC_WORKLOAD_CLUSTERS_COUNT}" ]]; then
   MGC_WORKLOAD_CLUSTERS_COUNT=1
 fi
+if [[ -z "${MGC_DEPLOY_SERVICE_PROTECTION}"]]; then
```
jasonmadigan, line 79:
```suggestion
if [[ -z "${MGC_DEPLOY_SERVICE_PROTECTION}" ]]; then
```

jasonmadigan:
```bash
export MGC_BRANCH=optional-service-protection
export MGC_DEPLOY_SERVICE_PROTECTION=false
```


```
kubectl --context kind-mgc-control-plane get managedclusteraddon -A
No resources found
```

```
kubectl --context kind-mgc-control-plane get operators kuadrant-operator -n kuadrant-system -o yaml
Error from server (NotFound): operators.operators.coreos.com "kuadrant-operator" not found
```

```
k  get ns kuadrant-system
Error from server (NotFound): namespaces "kuadrant-system" not found
```

mikenairn:
Looks fine.

Not sure i totally understand the point of creating a separate quick start script now though if it has so many optional paths. I would have thought the point of it was it was simple and consistent in what it was doing.

maleck13:
yeah it is a fair point. Perhaps this is not actually needed. Instead to not use service protection you could follow the "existing cluster install"

maleck13:
closed as we don't actually need this after all and it increases the complexity of the quick start and testing it

PR #629, opened 2023-10-18
Add DNSPolicy Routing Strategy

### LINKED ISSUE #655: Add simple routing strategy to DNSPolicy
Add simple routing strategy to DNSPolicy https://github.com/Kuadrant/architecture/pull/30

mikenairn:
Adds a routing strategy field to the DNSPolicy spec that determines how the policy with generate endpoints for any created DNSRecords.

Two routing strategies are allowed, `simple` and `loadbalanced`. Simple will creates a single DNS record (A or CNAME) for each listener/hostname with all ip/hostnames as targets. LoadBalanced works as before by creating a more complex record structure with CNAMES and A records using Geo and Weighted routing strategies to achieve loadbalancing functionality.

The strategy field is currently marked as immutable and it should not be changed after initial DNSPolicy creation.

closes #655

### DIFF
--- a/bundle/manifests/kuadrant.io_dnspolicies.yaml
+++ b/bundle/manifests/kuadrant.io_dnspolicies.yaml
@@ -156,6 +156,12 @@ spec:
                         type: integer
                     type: object
                 type: object
+              routingStrategy:
+                default: loadbalanced
+                enum:
+                - simple
+                - loadbalanced
+                type: string
               targetRef:
                 description: PolicyTargetReference identifies an API object to apply
                   policy to. This should be used as part of Policy resources that
@@ -194,6 +200,7 @@ spec:
                 - name
                 type: object
             required:
+            - routingStrategy
             - targetRef
             type: object
           status:
--- a/pkg/apis/v1alpha1/dnspolicy_types.go
+++ b/pkg/apis/v1alpha1/dnspolicy_types.go
@@ -25,6 +25,13 @@ import (
 	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 )
 
+type RoutingStrategy string
+
+const (
+	SimpleRoutingStrategy       RoutingStrategy = "simple"
+	LoadBalancedRoutingStrategy RoutingStrategy = "loadbalanced"
+)
+
 // DNSPolicySpec defines the desired state of DNSPolicy
 type DNSPolicySpec struct {
 
@@ -37,6 +44,11 @@ type DNSPolicySpec struct {
 
 	// +optional
 	LoadBalancing *LoadBalancingSpec `json:"loadBalancing"`
+
+	// +required
+	// +kubebuilder:validation:Enum=simple;loadbalanced
+	// +kubebuilder:default=loadbalanced
+	RoutingStrategy RoutingStrategy `json:"routingStrategy"`
 }
 
 type LoadBalancingSpec struct {
--- a/pkg/controllers/dnspolicy/dns_helper.go
+++ b/pkg/controllers/dnspolicy/dns_helper.go
@@ -33,8 +33,9 @@ const (
 )
 
 var (
-	ErrNoManagedZoneForHost = fmt.Errorf("no managed zone for host")
-	ErrAlreadyAssigned      = fmt.Errorf("managed host already assigned")
+	ErrUnknownRoutingStrategy = fmt.Errorf("unknown routing strategy")
+	ErrNoManagedZoneForHost   = fmt.Errorf("no managed zone for host")
+	ErrAlreadyAssigned        = fmt.Errorf("managed host already assigned")
 )
 
 type dnsHelper struct {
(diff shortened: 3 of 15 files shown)


maleck13:
want to merge this to single-cluster branch?
I almost have a pr now for a flag to turn off OCM integration

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -205,11 +203,14 @@ func getClusterGatewayAddresses(gw *gatewayv1beta1.Gateway) map[string][]gateway
 			}
 		}
 
-		gatewayAddresses = append(gatewayAddresses, gatewayv1beta1.GatewayAddress{
+		if _, ok := clusterAddrs[cluster]; !ok {
```
maleck13, line 206: yeah this is what I expected nice!

maleck13:
This is looking good to me, we need to wait on @makslion PR to allow setting things like geo and weight in a single cluster (IE to something custom like what we did with ManagedCLuster)?

mikenairn:
> This is looking good to me, we need to wait on @makslion PR to allow setting things like geo and weight in a single cluster (IE to something custom like what we did with ManagedCLuster)?

It should already be possible to do this by adding the same labels to the gateway itself in a single cluster context . Admittedly there is no test for it, but would be fairly easy to add now.

pkg/apis/v1alpha1/dnspolicy_types.go
```diff
@@ -37,6 +44,12 @@ type DNSPolicySpec struct {
 
 	// +optional
 	LoadBalancing *LoadBalancingSpec `json:"loadBalancing"`
+
+	// +required
+	// +kubebuilder:validation:Enum=simple;loadbalanced
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="RoutingStrategy is immutable"
+	// +kubebuilder:default=loadbalanced
+	RoutingStrategy RoutingStrategy `json:"routingStrategy"`
```
mikenairn, line 52: routingStrategy is required, but having the default means you don't have to specify it and will default to "loadbalanced" (The current behaviour)

pkg/controllers/dnspolicy/dns_helper.go
```diff
@@ -147,7 +148,74 @@ func withGatewayListener[T metav1.Object](gateway common.GatewayWrapper, listene
...
+	}
+
+	//ToDO This could possibly result in an invalid record since you can't have multiple CNAME target values https://github.com/Kuadrant/multicluster-gateway-controller/issues/663
+	if len(hostValues) > 0 {
+		endpoint := createOrUpdateEndpoint(hostname, hostValues, v1alpha1.CNAMERecordType, "", dns.DefaultTTL, currentEndpoints)
+		endpoints = append(endpoints, endpoint)
+	}
+
+	return endpoints
+}
```
mikenairn, line 216: Created a follow onto add validation around the Gateway/Routing Strategy https://github.com/Kuadrant/multicluster-gateway-controller/issues/663

I'm not 100% sure how to do it at the moment as i would prefer to leave this method just returning endpoints. validation should probably happen earlier on as part of processing the addresses which Maks is changing currently.

bundle/manifests/kuadrant.io_dnspolicies.yaml
```diff
@@ -156,6 +156,15 @@ spec:
...
                     type: object
                 type: object
+              routingStrategy:
+                default: loadbalanced
+                enum:
+                - simple
+                - loadbalanced
+                type: string
+                x-kubernetes-validations:
+                - message: RoutingStrategy is immutable
```
maleck13, line 166: I assume we are making this immutable because it is currently not possible to move between the two strategies? Is that something we could support at somepoint? I worry that it means deleting and removing your DNS to move between the two options

pkg/controllers/dnspolicy/dns_helper.go
```diff
@@ -147,7 +148,74 @@ func withGatewayListener[T metav1.Object](gateway common.GatewayWrapper, listene
...
+	if !equality.Semantic.DeepEqual(old, dnsRecord) {
+		return dh.Update(ctx, dnsRecord)
+	}
+
+	return nil
+}
+
+// getSimpleEndpoints returns the endpoints for the given MultiClusterGatewayTarget using the simple routing strategy
+
+func (dh *dnsHelper) getSimpleEndpoints(mcgTarget *dns.MultiClusterGatewayTarget, hostname string, currentEndpoints map[string]*v1alpha1.Endpoint) []*v1alpha1.Endpoint {
```
maleck13, line 186: might want to change the name of this type? Bit confusing when we are also working in single cluster

mikenairn: I assume you are talking about `MultiClusterGatewayTarget` and yes it needs renamed. Won't be in this PR though.

bundle/manifests/kuadrant.io_dnspolicies.yaml
mikenairn: This could be me being overly cautious. It might not actually be an issue going between record types (A vs CNAME) for the target host. Let me double check that.

test/policy_integration/dnspolicy_controller_test.go
```diff
@@ -381,1115 +355,1162 @@ var _ = Describe("DNSPolicy", Ordered, func() {
...
+											"SetIdentifier": Equal(""),
+											"RecordTTL":     Equal(v1alpha1.TTL(300)),
+										})),
+									),
+								}),
+							}),
+						))
+				}, TestTimeoutMedium, TestRetryIntervalMedium, ctx).Should(Succeed())
+			})
+
```
maleck13, line 1511: This is a pretty great test, but there is a lot going on in this file now. Perhaps as a follow up we could think about whether we should split this test in someway to make it easier to read / mantain? IE single vs multi or something

maleck13 (approved):
Happy to merge this, some minor comments.

test/policy_integration/dnspolicy_controller_test.go
mikenairn: yes, its far too long now, and a lot more could be added so we need to look at how we want these to be separated out. Probably at the very least a separate test file for multi/single and simple/loadbalanced contexts.

maleck13 (approved):

bundle/manifests/kuadrant.io_dnspolicies.yaml
mikenairn: I removed the validation as it should actually be possible to do this. I have created a follow on issue to look at updating the AWS provider which will currently error out when trying to update an existing record with a change of type (CNAME vs A).

https://github.com/Kuadrant/multicluster-gateway-controller/issues/664

All provider errors are already put into the DNSPOlicy ready condition so it will visible there.

Google _should_ work, but i didn't try it.

Issue #628, opened 2023-10-17
DNSRecord is not deleted when DNSPolicy is deleted


ficap:
When Gateway with bound DNSPolicy and HTTPRoute is deleted and then HTTPRoute and DNSPolicy are deleted associated DNSRecord object stays present and unchanged still referencing deleted gateway.
Steps to reproduce:
1. create Gateway
2. create DNSPolicy and bind it to the Gateway
3. create HTTPRoute and bind it to the Gateway
4. delete Gateway and wait for DNSPolicie's ready status to change to False
5. delete HTTPRoute and DNSPolicy
6. DNSRecord stays present and keeps pointing to IP of already deleted Gateway

PR #635, opened 2023-10-20
gh-628 fix deletion of dnsrecord and certificates on deletion of gateway target for dnspolicy and tlspolicy

laurafitzgerald:
Addressed #628  

# Verification
## DNSRecord
Follow the steps described in the issue, the dnsrecord should be deleted.

## Health Check Probes
kubectl create secret generic probe-headers --from-literal=test=dGVzdAo= -n multi-cluster-gateways

Follow the steps describe in the issue with the following adjustments. 
Add the following block to your dnspolicy
```
healthCheck:
      additionalHeadersRef:
        name: probe-headers
      endpoint: /
      expectedResponses:
      - 200
      - 201
      - 301
      failureThreshold: 1
      port: 443
      protocol: https
```
there should be one dnshealthcheck probe present in the cluster now
e.g.
```
kc get dnshealthcheckprobes -A        

NAMESPACE                NAME                        HEALTHY   LAST CHECKED
multi-cluster-gateways   172.31.200.0-prod-web-api   true      3s
```

Delete the dnspolicy, the healthcheck probe should be deleted.

## Certificates
- create gateway
- label gateway
- create tlspolicy
- verify certificates are created
- delete gateway
- verify certificates are deleted.

### DIFF
--- a/pkg/controllers/dnspolicy/dnspolicy_controller.go
+++ b/pkg/controllers/dnspolicy/dnspolicy_controller.go
@@ -195,13 +195,12 @@ func (r *DNSPolicyReconciler) deleteResources(ctx context.Context, dnsPolicy *v1
 	if err != nil {
 		return err
 	}
-
-	if err := r.reconcileDNSRecords(ctx, dnsPolicy, gatewayDiffObj); err != nil {
+	if err = r.deleteDNSRecords(ctx, dnsPolicy); err != nil {
 		log.V(3).Info("error reconciling DNS records from delete, returning", "error", err)
 		return err
 	}
 
-	if err := r.reconcileHealthChecks(ctx, dnsPolicy, gatewayDiffObj); err != nil {
+	if err := r.deleteProbes(ctx, dnsPolicy); err != nil {
 		return err
 	}
 
--- a/pkg/controllers/dnspolicy/dnspolicy_controller.go
+++ b/pkg/controllers/dnspolicy/dnspolicy_controller.go
@@ -159,7 +159,7 @@ func (r *DNSPolicyReconciler) reconcileResources(ctx context.Context, dnsPolicy 
 		return errors.Join(fmt.Errorf("reconcile DNSRecords error %w", err), updateErr)
 	}
 
-	if err = r.reconcileHealthChecks(ctx, dnsPolicy, gatewayDiffObj); err != nil {
+	if err = r.reconcileHealthCheckProbes(ctx, dnsPolicy, gatewayDiffObj); err != nil {
 		gatewayCondition = conditions.BuildPolicyAffectedCondition(DNSPolicyAffected, dnsPolicy, targetNetworkObject, conditions.PolicyReasonInvalid, err)
 		updateErr := r.updateGatewayCondition(ctx, gatewayCondition, gatewayDiffObj)
 		return errors.Join(fmt.Errorf("reconcile HealthChecks error %w", err), updateErr)
@@ -200,7 +200,7 @@ func (r *DNSPolicyReconciler) deleteResources(ctx context.Context, dnsPolicy *v1
 		return err
 	}
 
-	if err := r.deleteProbes(ctx, dnsPolicy); err != nil {
+	if err := r.deleteHealthCheckProbes(ctx, dnsPolicy); err != nil {
 		return err
 	}
 
--- a/pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
+++ b/pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
@@ -33,7 +33,7 @@ func (r *DNSPolicyReconciler) reconcileDNSRecords(ctx context.Context, dnsPolicy
 
 	// Reconcile DNSRecords for each gateway directly referred by the policy (existing and new)
 	for _, gw := range append(gwDiffObj.GatewaysWithValidPolicyRef, gwDiffObj.GatewaysMissingPolicyRef...) {
-		log.V(1).Info("reconcileDNSRecords: gateway with valid and missing policy ref", "key", gw.Key())
+		log.V(1).Info("reconcileDNSRecords: gateway with valid or missing policy ref", "key", gw.Key())
 		err := r.reconcileGatewayDNSRecords(ctx, gw.Gateway, dnsPolicy)
 		if err != nil {
 			return err
@@ -124,9 +124,17 @@ func (r *DNSPolicyReconciler) reconcileGatewayDNSRecords(ctx context.Context, ga
 }
 
 func (r *DNSPolicyReconciler) deleteGatewayDNSRecords(ctx context.Context, gateway *gatewayv1beta1.Gateway, dnsPolicy *v1alpha1.DNSPolicy) error {
+	return r.deleteDNSRecordsWithLabels(ctx, commonDNSRecordLabels(client.ObjectKeyFromObject(gateway), client.ObjectKeyFromObject(dnsPolicy)), dnsPolicy.Namespace)
+}
+
+func (r *DNSPolicyReconciler) deleteDNSRecords(ctx context.Context, dnsPolicy *v1alpha1.DNSPolicy) error {
+	return r.deleteDNSRecordsWithLabels(ctx, policyDNSRecordLabels(client.ObjectKeyFromObject(dnsPolicy)), dnsPolicy.Namespace)
+}
+
+func (r *DNSPolicyReconciler) deleteDNSRecordsWithLabels(ctx context.Context, lbls map[string]string, namespace string) error {
 	log := crlog.FromContext(ctx)
 
-	listOptions := &client.ListOptions{LabelSelector: labels.SelectorFromSet(commonDNSRecordLabels(client.ObjectKeyFromObject(gateway), client.ObjectKeyFromObject(dnsPolicy)))}
+	listOptions := &client.ListOptions{LabelSelector: labels.SelectorFromSet(lbls), Namespace: namespace}
 	recordsList := &v1alpha1.DNSRecordList{}
 	if err := r.Client().List(ctx, recordsList, listOptions); err != nil {
 		return err
--- a/pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
+++ b/pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
@@ -23,23 +23,21 @@ import (
 func (r *DNSPolicyReconciler) reconcileDNSRecords(ctx context.Context, dnsPolicy *v1alpha1.DNSPolicy, gwDiffObj *reconcilers.GatewayDiff) error {
 	log := crlog.FromContext(ctx)
 
+	log.V(3).Info("reconciling dns records")
 	for _, gw := range gwDiffObj.GatewaysWithInvalidPolicyRef {
 		log.V(1).Info("reconcileDNSRecords: gateway with invalid policy ref", "key", gw.Key())
-		err := r.deleteGatewayDNSRecords(ctx, gw.Gateway, dnsPolicy)
-		if err != nil {
-			return err
+		if err := r.deleteGatewayDNSRecords(ctx, gw.Gateway, dnsPolicy); err != nil {
+			return fmt.Errorf("error deleting dns records for gw %v: %w", gw.Gateway.Name, err)
 		}
 	}
 
 	// Reconcile DNSRecords for each gateway directly referred by the policy (existing and new)
 	for _, gw := range append(gwDiffObj.GatewaysWithValidPolicyRef, gwDiffObj.GatewaysMissingPolicyRef...) {
 		log.V(1).Info("reconcileDNSRecords: gateway with valid or missing policy ref", "key", gw.Key())
-		err := r.reconcileGatewayDNSRecords(ctx, gw.Gateway, dnsPolicy)
-		if err != nil {
-			return err
+		if err := r.reconcileGatewayDNSRecords(ctx, gw.Gateway, dnsPolicy); err != nil {
+			return fmt.Errorf("error reconciling dns records for gateway %v: %w", gw.Gateway.Name, err)
 		}
 	}
-
 	return nil
 }
 
--- a/pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
+++ b/pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
@@ -20,33 +20,33 @@ import (
 	"github.com/Kuadrant/multicluster-gateway-controller/pkg/apis/v1alpha1"
 )
 
-func (r *DNSPolicyReconciler) reconcileHealthChecks(ctx context.Context, dnsPolicy *v1alpha1.DNSPolicy, gwDiffObj *reconcilers.GatewayDiff) error {
+func (r *DNSPolicyReconciler) reconcileHealthCheckProbes(ctx context.Context, dnsPolicy *v1alpha1.DNSPolicy, gwDiffObj *reconcilers.GatewayDiff) error {
 	log := crlog.FromContext(ctx)
 
 	log.V(3).Info("reconciling health checks")
-	for _, gw := range append(gwDiffObj.GatewaysWithValidPolicyRef, gwDiffObj.GatewaysMissingPolicyRef...) {
-		log.V(3).Info("reconciling probes", "gateway", gw.Name)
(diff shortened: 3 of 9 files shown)


pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -25,15 +25,15 @@ func (r *DNSPolicyReconciler) reconcileDNSRecords(ctx context.Context, dnsPolicy
 
 	for _, gw := range gwDiffObj.GatewaysWithInvalidPolicyRef {
 		log.V(1).Info("reconcileDNSRecords: gateway with invalid policy ref", "key", gw.Key())
-		err := r.deleteGatewayDNSRecords(ctx, gw.Gateway, dnsPolicy)
+		err := r.deleteDNSRecords(ctx, dnsPolicy)
```
mikenairn, line 28: This will need to use both the gateway and policy labels since it should only be deleting resources created for the gateway being processed (gw.Gateway). What `deleteGatewayDNSRecords` was doing originally was correct in this scenario.

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -123,10 +123,10 @@ func (r *DNSPolicyReconciler) reconcileGatewayDNSRecords(ctx context.Context, ga
 	return nil
 }
 
-func (r *DNSPolicyReconciler) deleteGatewayDNSRecords(ctx context.Context, gateway *gatewayv1beta1.Gateway, dnsPolicy *v1alpha1.DNSPolicy) error {
+func (r *DNSPolicyReconciler) deleteDNSRecords(ctx context.Context, dnsPolicy *v1alpha1.DNSPolicy) error {
 	log := crlog.FromContext(ctx)
 
-	listOptions := &client.ListOptions{LabelSelector: labels.SelectorFromSet(commonDNSRecordLabels(client.ObjectKeyFromObject(gateway), client.ObjectKeyFromObject(dnsPolicy)))}
+	listOptions := &client.ListOptions{LabelSelector: labels.SelectorFromSet(policyDNSRecordLabels(client.ObjectKeyFromObject(dnsPolicy)))}
```
mikenairn, line 129: You will need to be able to delete using both the combined (common) labels (when deleting only records for a single gateway) and just policy labels (When you are cleaning everything up on policy or gateway deletion). I'd suggest creating a `deleteDNSRecordsWithLabels` (or some similarly named method) and call that from other methods, such as : 

```golang
func (r *DNSPolicyReconciler) deleteGatewayDNSRecords(ctx context.Context, gateway *gatewayv1beta1.Gateway, dnsPolicy *v1alpha1.DNSPolicy) error {
	return r.deleteDNSRecordsWithLabels(ctx, commonDNSRecordLabels(client.ObjectKeyFromObject(gateway), client.ObjectKeyFromObject(dnsPolicy)))
}

func (r *DNSPolicyReconciler) deleteDNSRecords(ctx context.Context, dnsPolicy *v1alpha1.DNSPolicy) error {
	return r.deleteDNSRecordsWithLabels(ctx, (policyDNSRecordLabels(client.ObjectKeyFromObject(dnsPolicy))))
}

func (r *DNSPolicyReconciler) deleteDNSRecordsWithLabels(ctx context.Context, lbls map[string]string) error {
	log := crlog.FromContext(ctx)

	listOptions := &client.ListOptions{LabelSelector: labels.SelectorFromSet(lbls)}
	recordsList := &v1alpha1.DNSRecordList{}
	if err := r.Client().List(ctx, recordsList, listOptions); err != nil {
		return err
	}

	for _, record := range recordsList.Items {
		if err := r.DeleteResource(ctx, &record); client.IgnoreNotFound(err) != nil {
			log.Error(err, "failed to delete DNSRecord")
			return err
		}
	}
	return nil
}
```

pkg/controllers/tlspolicy/tlspolicy_certmanager_certificates.go
```diff
@@ -27,14 +27,14 @@ func (r *TLSPolicyReconciler) reconcileCertificates(ctx context.Context, tlsPoli
 
 	for _, gw := range gwDiffObj.GatewaysWithInvalidPolicyRef {
 		log.V(1).Info("reconcileCertificates: gateway with invalid policy ref", "key", gw.Key())
-		if err := r.deleteGatewayCertificates(ctx, gw.Gateway, tlsPolicy); err != nil {
+		if err := r.deleteCertificates(ctx, tlsPolicy); err != nil {
```
mikenairn, line 30: Same here, need to be bale to delete just the gateway certificates, not all.

pkg/controllers/tlspolicy/tlspolicy_controller.go
```diff
@@ -196,7 +196,7 @@ func (r *TLSPolicyReconciler) deleteResources(ctx context.Context, tlsPolicy *v1
 		return err
 	}
 
-	if err := r.reconcileCertificates(ctx, tlsPolicy, gatewayDiffObj); err != nil {
+	if err := r.deleteCertificates(ctx, tlsPolicy); err != nil {
```
mikenairn, line 199: :+1:

pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -195,8 +195,7 @@ func (r *DNSPolicyReconciler) deleteResources(ctx context.Context, dnsPolicy *v1
 	if err != nil {
 		return err
 	}
-
-	if err := r.reconcileDNSRecords(ctx, dnsPolicy, gatewayDiffObj); err != nil {
+	if err = r.deleteDNSRecords(ctx, dnsPolicy); err != nil {
```
mikenairn, line 198: :+1: 

I think you  will need to do the same for `reconcileHealthChecks` though.

pkg/controllers/tlspolicy/tlspolicy_certmanager_certificates.go
```diff
@@ -65,14 +65,14 @@ func (r *TLSPolicyReconciler) reconcileGatewayCertificates(ctx context.Context,
...
 
-func (r *TLSPolicyReconciler) deleteUnexpectedGatewayCertificates(ctx context.Context, expectedCerts []*certmanv1.Certificate, gateway *gatewayv1beta1.Gateway, tlsPolicy *v1alpha1.TLSPolicy) error {
+func (r *TLSPolicyReconciler) deleteUnexpectedCertificates(ctx context.Context, expectedCerts []*certmanv1.Certificate, tlsPolicy *v1alpha1.TLSPolicy) error {
 	log := crlog.FromContext(ctx)
 
-	listOptions := &client.ListOptions{LabelSelector: labels.SelectorFromSet(tlsCertificateLabels(client.ObjectKeyFromObject(gateway), client.ObjectKeyFromObject(tlsPolicy)))}
+	listOptions := &client.ListOptions{LabelSelector: labels.SelectorFromSet(policyTLSCertificateLabels(client.ObjectKeyFromObject(tlsPolicy)))}
 	certList := &certmanv1.CertificateList{}
 	if err := r.Client().List(ctx, certList, listOptions); err != nil {
 		return err
```
mikenairn, line 78: Same suggestion as DNS, add a `deleteXWithLabels` might be a better option.

mikenairn (changes requested):
Looks good, just a couple of small changes that i think we need to make sure we aren't deleting the wrong resources when processing the gateways that have an invalid policy ref.

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -124,9 +124,17 @@ func (r *DNSPolicyReconciler) reconcileGatewayDNSRecords(ctx context.Context, ga
...
+
+func (r *DNSPolicyReconciler) deleteDNSRecords(ctx context.Context, dnsPolicy *v1alpha1.DNSPolicy) error {
+	return r.deleteDNSRecordsWithLabels(ctx, (policyDNSRecordLabels(client.ObjectKeyFromObject(dnsPolicy))))
+}
+
+func (r *DNSPolicyReconciler) deleteDNSRecordsWithLabels(ctx context.Context, lbls map[string]string) error {
 	log := crlog.FromContext(ctx)
 
-	listOptions := &client.ListOptions{LabelSelector: labels.SelectorFromSet(commonDNSRecordLabels(client.ObjectKeyFromObject(gateway), client.ObjectKeyFromObject(dnsPolicy)))}
+	listOptions := &client.ListOptions{LabelSelector: labels.SelectorFromSet(lbls)}
```
mikenairn, line 137: Probably want the namespace in the listOptions  also `Namespace: dnsPolicy.Namespace`

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
```diff
@@ -66,25 +67,45 @@ func (r *DNSPolicyReconciler) createOrUpdateProbes(ctx context.Context, expected
...
+	return r.deleteProbesWithLabels(ctx, commonDNSRecordLabels(client.ObjectKeyFromObject(gateway), client.ObjectKeyFromObject(dnsPolicy)))
+}
+
+func (r *DNSPolicyReconciler) deleteProbes(ctx context.Context, dnsPolicy *v1alpha1.DNSPolicy) error {
+	return r.deleteProbesWithLabels(ctx, policyDNSRecordLabels(client.ObjectKeyFromObject(dnsPolicy)))
+}
+
+func (r *DNSPolicyReconciler) deleteProbesWithLabels(ctx context.Context, lbls map[string]string) error {
+	probes := &v1alpha1.DNSHealthCheckProbeList{}
+	listOptions := &client.ListOptions{LabelSelector: labels.SelectorFromSet(lbls)}
```
mikenairn, line 80: Probably want the namespace in the listOptions  also `Namespace: dnsPolicy.Namespace`

pkg/controllers/tlspolicy/tlspolicy_certmanager_certificates.go
```diff
@@ -65,14 +65,18 @@ func (r *TLSPolicyReconciler) reconcileGatewayCertificates(ctx context.Context,
...
-func (r *TLSPolicyReconciler) deleteUnexpectedGatewayCertificates(ctx context.Context, expectedCerts []*certmanv1.Certificate, gateway *gatewayv1beta1.Gateway, tlsPolicy *v1alpha1.TLSPolicy) error {
+func (r *TLSPolicyReconciler) deleteCertificates(ctx context.Context, tlsPolicy *v1alpha1.TLSPolicy) error {
+	return r.deleteCertificatesWithLabels(ctx, []*certmanv1.Certificate{}, policyTLSCertificateLabels(client.ObjectKeyFromObject(tlsPolicy)))
+}
+
+func (r *TLSPolicyReconciler) deleteCertificatesWithLabels(ctx context.Context, expectedCerts []*certmanv1.Certificate, lbls map[string]string) error {
 	log := crlog.FromContext(ctx)
 
-	listOptions := &client.ListOptions{LabelSelector: labels.SelectorFromSet(tlsCertificateLabels(client.ObjectKeyFromObject(gateway), client.ObjectKeyFromObject(tlsPolicy)))}
+	listOptions := &client.ListOptions{LabelSelector: labels.SelectorFromSet(lbls)}
```
mikenairn, line 79: Probably want the namespace in the listOptions also Namespace: dnsPolicy.Namespace

laurafitzgerald:
The e2e test failure here is intermittant and appears to be not releted to the change. 
The failure happens on error `[debug] Cert error: 'wildcard hostname not found in the certificate via get request'` which is in relation to tlspolicy and secret creation. The first two failures were on the AWS suite and the third is on GCP suite. Locally this test is passing for me on AWS backed cluster.

Aside from this feedback has been addressed and verification steps are updated. 

I will keep investigating the failure but it seems intermittent and unrelated.

cc @mikenairn

pkg/controllers/tlspolicy/tlspolicy_certmanager_certificates.go
```diff
@@ -65,14 +65,18 @@ func (r *TLSPolicyReconciler) reconcileGatewayCertificates(ctx context.Context,
 	return nil
 }
 
-func (r *TLSPolicyReconciler) deleteGatewayCertificates(ctx context.Context, gateway *gatewayv1beta1.Gateway, tlsPolicy *v1alpha1.TLSPolicy) error {
-	return r.deleteUnexpectedGatewayCertificates(ctx, []*certmanv1.Certificate{}, gateway, tlsPolicy)
+func (r *TLSPolicyReconciler) deleteGatewayCertificates(ctx context.Context, expectedCerts []*certmanv1.Certificate, gateway *gatewayv1beta1.Gateway, tlsPolicy *v1alpha1.TLSPolicy) error {
+	return r.deleteCertificatesWithLabels(ctx, expectedCerts, commonTLSCertificateLabels(client.ObjectKeyFromObject(gateway), client.ObjectKeyFromObject(tlsPolicy)), tlsPolicy.Namespace)
```
mikenairn, line 69: Any reason you have removed the `deleteUnexpectedGatewayCertificates` method here, but chose to keep `deleteUnexpectedGatewayProbes` in the dnspolicy_healthchecks reconcile?

pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -195,13 +195,12 @@ func (r *DNSPolicyReconciler) deleteResources(ctx context.Context, dnsPolicy *v1
...
 		log.V(3).Info("error reconciling DNS records from delete, returning", "error", err)
 		return err
 	}
 
-	if err := r.reconcileHealthChecks(ctx, dnsPolicy, gatewayDiffObj); err != nil {
+	if err := r.deleteProbes(ctx, dnsPolicy); err != nil {
 		return err
```
mikenairn, line 204: nit: might be better to name this to delete\<resource name\> for consistency i.e. `deleteHealthCheckProbes`

mikenairn (approved):

mikenairn:
@laurafitzgerald Looks good to me. Left a couple of comments, will let you reply to those before sticking the lgtm on it.

pkg/controllers/tlspolicy/tlspolicy_certmanager_certificates.go
laurafitzgerald: My thinking at the time was to bring it in line with the DNSRecords reconciation, but looking at it now it makes sense to keep it. I'm bringing the three more in line with each other so I'll return the unexpectedcertificates to matche the unexpectedhealthcheckprobes. Doesn't seem a need for it in the dnsrecords reconciliation though so I wont add it there. Commit coming soon.

pkg/controllers/tlspolicy/tlspolicy_certmanager_certificates.go
```diff
@@ -25,70 +25,88 @@ import (
...
-			return err
+		log.V(1).Info("reconcileCertificates: gateway with valid or missing policy ref", "key", gw.Key())
+		expectedCertificates := r.expectedCertificatesForGateway(ctx, gw.Gateway, tlsPolicy)
+		if err := r.createOrUpdateGatewayCertificates(ctx, expectedCertificates); err != nil {
+			return fmt.Errorf("error creating and updating expected certificates for gateway %v: %w", gw.Gateway.Name, err)
+		}
+		if err := r.deleteUnexpectedCertificates(ctx, expectedCertificates, gw.Gateway, tlsPolicy); err != nil {
+			return fmt.Errorf("error removing unexpected certificate for gateway %v: %w", gw.Gateway.Name, err)
 		}
 	}
```
mikenairn, line 46: Personally i kind of preferred the pattern this (and DNSRecord) was using of having a reconcileGateway\<Resource\> method that deals with reconciling resources for a given Gateway instead of having it broken up inside the main reconcile\<Resource\> method.

- reconcile\<Resource\> - Reconcile resources(create/update/delete) for the current policy
- reconcileGateway\<Resource\> - Reconcile resources(create/update/delete) for the current policy and gateway
- delete\<Resource\> - Delete resources for the current policy
- deleteGateway\<Resource\> - Delete resources for the current policy and gateway

I'd say if we are trying to make these controllers consistent (looks like you have copied the pattern from health checks) we should actually update the DNSRecord one to deal with unexpected gateway resources in the same way as the other two.

Since none of that is in the scope of this bug fix i'd say we can come back to this later.

mikenairn (approved):

PR #638, opened 2023-10-25
GH-496 GW target own cluster

### LINKED ISSUE #496: Update DNSPolicy reconciler to function with non mgc gateways (e.g. istio)
## What 
After the #311 the DNS Policy reconciler is no longer coupled with OCM. This enables us to process other Gateway Implementations. However, the logic, currently, implies that we are dealing with the multicluster gateways. 

https://github.com/Kuadrant/multicluster-gateway-controller/pull/465/files#r1324149471

## How 
- [x] Update reconciler logic to be suitable for both single- and multi-cluster implementations of the Gateway 
- [x] Update tests to ensure we have coverage for both types of gateway status, multi cluster (MGC) and singe (ISTIO etc ...)

makslion:
From the original only 1 task is left that are covered here: validate Gateway CR when it arrives for the reconciliation loop and add a test for an invalid Gateway. 

In this context invalid == has a mixture single- and multicluster addresses. We expect only one type of addresses to be present at the Gateway CR.

I'm also giving a second life to the `traffic.Gateway` transforming it into the `utils.GatewayWrapper` to preform the mentioned check. Keeping it separate from main packages for 

- Simplicity of policy controllers departure 
- Possibilities to use the wrapper in other packages without the fear of import cycles  

closes #496


pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -146,8 +146,10 @@ func (r *DNSPolicyReconciler) buildClusterGateway(ctx context.Context, downstrea
 	singleClusterAddresses := make([]gatewayv1beta1.GatewayAddress, len(clusterAddress))
 
 	mc := &clusterv1.ManagedCluster{}
-	if err := r.Client().Get(ctx, client.ObjectKey{Name: downstreamClusterName}, mc, &client.GetOptions{}); err != nil {
-		return target, err
+	if downstreamClusterName != dns.NoneCluster {
```
maleck13, line 149: nit: NoneCluster reads awkwardly, what are we trying to convey? Is it that there is no downstream cluster or that we are in a single cluster?

makslion: It will be replaced with the gateway instead

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -45,42 +46,47 @@ func (r *DNSPolicyReconciler) reconcileDNSRecords(ctx context.Context, dnsPolicy
 	return nil
 }
 
-func (r *DNSPolicyReconciler) reconcileGatewayDNSRecords(ctx context.Context, gateway *gatewayv1beta1.Gateway, dnsPolicy *v1alpha1.DNSPolicy) error {
+func (r *DNSPolicyReconciler) reconcileGatewayDNSRecords(ctx context.Context, gw *gatewayv1beta1.Gateway, dnsPolicy *v1alpha1.DNSPolicy) error {
 	log := crlog.FromContext(ctx)
 
-	if err := r.dnsHelper.removeDNSForDeletedListeners(ctx, gateway); err != nil {
+	trafficGateway, err := traffic.NewGateway(gw)
```
mikenairn, line 52: The traffic pkg is probably a bit of an old concept now since we don't work with anything other than Gateways. It's the right idea, use a wrapper like this, but I'm wondering we should maybe just delete the traffic package (Pretty sure it[s not sued anywhere) and put it elsewhere :thinking:

pkg/traffic/gateway.go
```diff
@@ -10,19 +10,38 @@ import (
...
+
+	}
+	return gw, nil
+}
+
+func isMultiClusterAddressType(addressType gatewayv1beta1.AddressType) bool {
+	return addressType == gateway.MultiClusterIPAddressType || addressType == gateway.MultiClusterHostnameAddressType
 }
 
 type Gateway struct {
```
mikenairn, line 42: This is already here, but we probably should avoid calling it "Gateway", GatewayWrapper would probably be better similar to what the kuadrant operator does see https://github.com/Kuadrant/kuadrant-operator/blob/main/pkg/common/gatewayapi_utils.go#L355

pkg/traffic/gateway.go
```diff
@@ -10,19 +10,38 @@ import (
...
+
+	for i, address := range gw.Status.Addresses {
+		if i == 0 {
+			gw.isMultiCluster = isMultiClusterAddressType(*address.Type)
+			continue
+		}
+		if gw.isMultiCluster == isMultiClusterAddressType(*address.Type) {
+			continue
+		}
+		return nil, fmt.Errorf("gateway is invalid: inconsistent status addresses")
```
mikenairn, line 32: Not really following this logic, can you explain what it's doing?

pkg/traffic/gateway.go
```diff
@@ -122,3 +141,7 @@ func (a *Gateway) GetListenerByHost(host string) *gatewayv1beta1.Listener {
 func (a *Gateway) ExposesOwnController() bool {
 	return false
 }
+
+func (a *Gateway) IsMultiCluster() bool {
+	return a.isMultiCluster
```
mikenairn, line 146: I'd probably avoid setting a boolean in the wrapper struct and only calculating it once in a constructor, but instead add the logic to  `IsMultiCluster` and do every time it's called.

mikenairn: Presumably a false value from here could also mean "Unknown" since the Gateway might have no status yet?

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
makslion: I'd still like to keep it away from any serious packages to not beg for import cycles. Refactored it a bit to get rid of old juju :thinking:

pkg/traffic/gateway.go
```diff
@@ -10,19 +10,38 @@ import (
 	gatewayv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 
 	"github.com/Kuadrant/multicluster-gateway-controller/pkg/_internal/slice"
+	"github.com/Kuadrant/multicluster-gateway-controller/pkg/controllers/gateway"
```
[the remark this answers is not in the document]
makslion: the import is purely for the `MultiClusterIPAddressType` and `MultiClusterHostnameAddressType` but you have a point. I'll keep moving those constants until they find a home :thinking: What if they reside in this package and all usages in normal packages will be replaced with functions? :thinking:

pkg/traffic/gateway.go
makslion: `GatewayWrapper` it is. Renamed.

pkg/traffic/gateway.go
makslion: So it goes through the list of all the addresses and ensures they are the same type. 
The first address sets the boolean flag on the wrapper; all addresses after this are compared against this flag. In case the address type and previously set flag don't agree we have an error

pkg/traffic/gateway.go
makslion: Why would you avoid it? Do we expect updated copies of Gateway CR mid-reconciliation, update `Status.Addresses[]`, or the boolean flag "manually" to invalidate the check? 
The idea I had: before doing any reconciliation we validate the Gateway we got and if it is fine we stop checking it singe Gateway should remain the same (any updates will be dealt with but a new loop with a new check in the beginning)

makslion: If the gateway has no status - we don't really care about it being a single or multicluster one :thinking:

pkg/controllers/dnspolicy/dns_helper.go
```diff
@@ -124,7 +124,7 @@ func (dh *dnsHelper) buildDNSRecordForListener(gateway *gatewayv1beta1.Gateway,
 }
 
 // getDNSRecordForListener returns a v1alpha1.DNSRecord, if one exists, for the given listener in the given v1alpha1.ManagedZone.
-// It needs a reference string to enforce DNS record serving a single traffic.Interface owner
+// It needs a reference string to enforce DNS record serving a single utils.Interface owner
```
mikenairn, line 127: You could probably just remove this line, I've read it 5 times and i don't know what it means, maybe meant something once :shrug: .

pkg/utils/gateway_wrapper.go
```diff
@@ -0,0 +1,80 @@
...
+
+	}
+	return gw, nil
+}
+
+func isMultiClusterAddressType(addressType gatewayv1beta1.AddressType) bool {
+	return addressType == MultiClusterIPAddressType || addressType == MultiClusterHostnameAddressType
+}
+
+func (a *GatewayWrapper) GetKind() string {
```
mikenairn, line 41: "g" would probably be a better receiver name

pkg/utils/gateway_wrapper.go
```diff
@@ -0,0 +1,80 @@
...
+	return "GatewayWrapper"
+}
+
+func (a *GatewayWrapper) IsMultiCluster() bool {
+	return a.isMultiCluster
+}
```
mikenairn, line 47: I think we would be better to split the validation part from the check for multi cluster. Reason being we need to do other validation of the gateway, and it would all be better done at the start as part of an explicit action. I'm thinking something like:

```
func (g *GatewayWrapper) Validate() error {
	for _, address := range g.Status.Addresses {
		if isMultiClusterAddress(address) != g.IsMultiCluster() {
			return fmt.Errorf("gateway is invalid: inconsistent status addresses")
		}
	}
	return nil
}

func (g *GatewayWrapper) IsMultiCluster() bool {
	if len(g.Status.Addresses) > 0 {
		return isMultiClusterAddress(g.Status.Addresses[0])
	}
	return false
}
```

pkg/utils/gateway_wrapper.go
```diff
@@ -0,0 +1,80 @@
...
+func (a *GatewayWrapper) GetNamespaceName() types.NamespacedName {
+	return types.NamespacedName{
+		Namespace: a.Namespace,
+		Name:      a.Name,
+	}
+}
+
+func (a *GatewayWrapper) String() string {
+	return fmt.Sprintf("kind: %v, namespace/name: %v", a.GetKind(), a.GetNamespaceName())
+}
```
mikenairn, line 58: I think we could remove `GetKind()`, `GetNamespaceName()`, and  `String()`, not sure we need any of that.

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -185,7 +188,7 @@ func (r *DNSPolicyReconciler) buildClusterGateway(ctx context.Context, clusterNa
 	return target, nil
 }
 
-func getClusterGatewayAddresses(gw *gatewayv1beta1.Gateway) map[string][]gatewayv1beta1.GatewayAddress {
+func getClusterGatewayAddresses(gw *utils.GatewayWrapper) map[string][]gatewayv1beta1.GatewayAddress {
```
mikenairn, line 191: Could `getClusterGatewayAddresses` be moved to the wrapper?

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -216,11 +219,10 @@ func getClusterGatewayAddresses(gw *gatewayv1beta1.Gateway) map[string][]gateway
 	return clusterAddrs
 }
 
-func listenerTotalAttachedRoutes(upstreamGateway *gatewayv1beta1.Gateway, downstreamCluster string, specListener gatewayv1beta1.Listener, addresses []gatewayv1beta1.GatewayAddress) int {
```
mikenairn, line 219: Could listenerTotalAttachedRoutes be moved to the wrapper?

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -167,12 +172,10 @@ func (r *DNSPolicyReconciler) buildClusterGateway(ctx context.Context, clusterNa
...
-		if addrType == gateway.MultiClusterHostnameAddressType {
-			addrType = gatewayv1beta1.HostnameAddressType
-		}
-		if addrType == gateway.MultiClusterIPAddressType {
-			addrType = gatewayv1beta1.IPAddressType
+		addrType, multicluster := utils.AddressTypeToSingleCluster(addr)
+
+		if !multicluster {
+			addrType = *addr.Type
 		}
```
mikenairn, line 179: We don't need this check here. This should work with any gateway and we don't need to check other types of addresses only our own and convert them to the appropriate single cluster type.

pkg/controllers/dnspolicy/dns_helper.go
makslion: removed

pkg/utils/gateway_wrapper.go
makslion: wondering why it was `a` initially :thinking:  you are correct - now it will be `g`

pkg/utils/gateway_wrapper.go
makslion: done

pkg/utils/gateway_wrapper.go
makslion: got rid of them

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
makslion: moved. Thanks for spotting!

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
makslion: That is true. The `AddressTypeToSingleCluster` is already doing that check

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -45,42 +40,48 @@ func (r *DNSPolicyReconciler) reconcileDNSRecords(ctx context.Context, dnsPolicy
...
-func (r *DNSPolicyReconciler) reconcileGatewayDNSRecords(ctx context.Context, gateway *gatewayv1beta1.Gateway, dnsPolicy *v1alpha1.DNSPolicy) error {
+func (r *DNSPolicyReconciler) reconcileGatewayDNSRecords(ctx context.Context, gw *gatewayv1beta1.Gateway, dnsPolicy *v1alpha1.DNSPolicy) error {
 	log := crlog.FromContext(ctx)
 
-	if err := r.dnsHelper.removeDNSForDeletedListeners(ctx, gateway); err != nil {
+	gatewayWrapper := utils.NewGatewayWrapper(gw)
+	err := gatewayWrapper.Validate()
+	if err != nil {
+		return err
+	}
```
mikenairn, line 50: For consistency with the rest of this file, this check might be better written as:

```
gatewayWrapper := utils.NewGatewayWrapper(gw)
if err := gatewayWrapper.Validate(); err != nil {
	return err
}
```

pkg/utils/gateway_wrapper.go
```diff
@@ -0,0 +1,129 @@
...
+
+// AddressTypeToMultiCluster returns a multi cluster version of the address type
+// and a bool to indicate that provided address has supported type
+func AddressTypeToMultiCluster(address gatewayv1beta1.GatewayAddress) (gatewayv1beta1.AddressType, bool) {
+	if *address.Type == gatewayv1beta1.IPAddressType {
+		return MultiClusterIPAddressType, true
+	} else if *address.Type == gatewayv1beta1.HostnameAddressType {
+		return MultiClusterHostnameAddressType, true
+	}
+	return "", false
```
mikenairn, line 118: Not sure where this would be used other than where it is already, but you could have this return the original address type (*address.Type) which might be more useful in other circumstances.

```.go
return *address.Type, false
```

test/policy_integration/dnspolicy_controller_test.go
```diff
@@ -137,6 +137,55 @@ var _ = Describe("DNSPolicy", func() {
...
+			}, TestTimeoutMedium, TestRetryIntervalMedium).ShouldNot(HaveOccurred())
+
+			// expect no dns records
+			Consistently(func() []v1alpha1.DNSRecord {
+				dnsRecords := v1alpha1.DNSRecordList{}
+				err := k8sClient.List(ctx, &dnsRecords, client.InNamespace(dnsPolicy.GetNamespace()))
+				Expect(err).ToNot(HaveOccurred())
+				return dnsRecords.Items
+			}, time.Second*15, time.Second).Should(BeEmpty())
+		})
```
mikenairn, line 187: A better, or at least additional, check here would probably be to check the dnspolicy status is as expected see https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/test/policy_integration/dnspolicy_controller_test.go#L100-L110. It should be false with an error in the message.

The `Consistently` check for no records is probably fine, but we need to be careful adding many of these as they do slow down the whole test suite.

test/policy_integration/dnspolicy_controller_test.go
```diff
@@ -137,6 +137,55 @@ var _ = Describe("DNSPolicy", func() {
...
 
+		It("should not process gateway with inconsistent addresses", func() {
+			// build invalid gateway
+			gateway = testutil.NewGatewayBuilder("test-gateway", gatewayClass.Name, testNamespace).
+				WithHTTPListener(TestListenerNameOne, TestHostOne).Gateway
+			Expect(k8sClient.Create(ctx, gateway)).To(BeNil())
+
+			// ensure gateway exists
+			Eventually(func() error {
+				return k8sClient.Get(ctx, client.ObjectKey{Name: gateway.Name, Namespace: gateway.Namespace}, gateway)
```
mikenairn, line 148: You can probably just do this get in the same `Eventually` closure as the status update, it's not really adding anything other than extra lines to the test file.

pkg/utils/gateway_wrapper.go
mikenairn: Thinking about it, if you wanted to keep the signatures of `AddressTypeToMultiCluster` and `AddressTypeToSingleCluster` the same  (I kind of like being able to do that where possible) we could add back the bool to `AddressTypeToSingleCluster` and have both return `*address.Type, false` in the case where we are not converting it. The bool value tells you if it was converted or not, rather being "supported"

mikenairn:
Few small comments/suggestions, but i think this is looking good. I'm not sure if `utils` will be the eventual home of the gateway wrapper but its fine for now, we can decide where it should live as it evolves (Closer alignment/merge with MultiClusterGatewayTarget).

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
makslion: changed

pkg/utils/gateway_wrapper.go
makslion: hmm, i can see an argument for: if we converted an address to the multicluster type we always must be able to convert it back :thinking:  and I can't see how information of was it converted or not will be useful :thinking: 
But in the same time: consistency and for the future changes - changed.

pkg/utils/gateway_wrapper.go
```diff
@@ -0,0 +1,129 @@
...
+
+// AddressTypeToSingleCluster converts provided multicluster address to single cluster version if applicable
+func AddressTypeToSingleCluster(address gatewayv1beta1.GatewayAddress) gatewayv1beta1.AddressType {
+	if *address.Type == MultiClusterIPAddressType {
+		return gatewayv1beta1.IPAddressType
+	} else if *address.Type == MultiClusterHostnameAddressType {
+		return gatewayv1beta1.HostnameAddressType
+	}
+	return *address.Type
+}
```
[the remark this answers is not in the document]
makslion: Also added a few extra (tinny) unit cases. Should not be a problem to maintain them

test/policy_integration/dnspolicy_controller_test.go
makslion: Added an extra check, thanks! 
It feels strange to check for the absence of something with `Eventually` :thinking:  But yes, we should not use it often.

test/policy_integration/dnspolicy_controller_test.go
makslion: That is a good advice, thanks. Joined them together

test/policy_integration/dnspolicy_controller_test.go
```diff
@@ -137,6 +137,55 @@ var _ = Describe("DNSPolicy", func() {
 			}, TestTimeoutMedium, time.Second).Should(Succeed())
 		})
 
+		It("should not process gateway with inconsistent addresses", func() {
+			// build invalid gateway
+			gateway = testutil.NewGatewayBuilder("test-gateway", gatewayClass.Name, testNamespace).
+				WithHTTPListener(TestListenerNameOne, TestHostOne).Gateway
+			Expect(k8sClient.Create(ctx, gateway)).To(BeNil())
```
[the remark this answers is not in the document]
makslion: that is nice!

makslion:
> Few small comments/suggestions, but i think this is looking good. I'm not sure if `utils` will be the eventual home of the gateway wrapper but its fine for now, we can decide where it should live as it evolves (Closer alignment/merge with MultiClusterGatewayTarget).

Would you have a better place/name in mind for the GatewayWrapper package? :thinking:  don't like either of those but couldn't come up with a better place and name

pkg/utils/gateway_wrapper.go
```diff
@@ -0,0 +1,130 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+const (
+	SingleClusterAddressValue = "kudarant.io/single"
```
mikenairn, line 11: This isn't an address value, `SingleClusterNameValue` might be more appropriate.

test/policy_integration/dnspolicy_controller_test.go
```diff
@@ -137,6 +137,65 @@ var _ = Describe("DNSPolicy", func() {
...
+			}, time.Second*15, time.Second).Should(BeEmpty())
+			Eventually(func(g Gomega) {
+				err := k8sClient.Get(ctx, client.ObjectKeyFromObject(dnsPolicy), dnsPolicy)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(dnsPolicy.Status.Conditions).To(
+					ContainElement(MatchFields(IgnoreExtras, Fields{
+						"Type":    Equal(string(conditions.ConditionTypeReady)),
+						"Status":  Equal(metav1.ConditionFalse),
+						"Reason":  Equal("ReconciliationError"),
+						"Message": ContainSubstring("ateway is invalid: inconsistent status addresses"),
```
mikenairn, line 193:
```suggestion
						"Message": ContainSubstring("gateway is invalid: inconsistent status addresses"),
```

makslion: fixed

Issue #622, opened 2023-10-11
Remove OCM module dependencies from DNSPolicy


mikenairn:
Follow on from https://github.com/Kuadrant/multicluster-gateway-controller/issues/311

**What**

Remove the ocm module dependencies from the dnspolicy reconcilers: [dnspolicy_controller.go](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/controllers/dnspolicy/dnspolicy_controller.go#L26)  and [dnspolicy_dnsrecords.go](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go#L8).

**Why**

* Allow the policy to work independently of OCM or any other control plane. 
* Allow the policy to be moved to another repo without OCM being required.
* Ensure all OCM integration is performed by the gateway controller

**How**

Update the Gateway Controller to watch mangedclusterresources and add annotations to placed gateways on those clusters with any customisation such as geo or weighted values. Gateway controller should add the following annotation
```kuadrant.io/{managedcluster}/geo: "US"``` 
The policy controller need to be updated to look for these annotations on the gateway resources and use them for the associated DNS records. 
It should also be able to handle an annotation that has no {managedcluster} part as a single cluster value

makslion:
for the last item - there is a follow up after #311 - the #496

mikenairn:
> for the last item - there is a follow up after #311 - the #496 it could have a bit of an overlap

Remove last item in favour of https://github.com/Kuadrant/multicluster-gateway-controller/issues/496. This issue is now only concerned about making the "cluster" meta resource configurable for use when OCM is not the control plane of choice, or no control plane exists. Low priority at this moment in time.

PR #644, opened 2023-10-26
Separate policy controller

### LINKED ISSUE #639: separate binaries and installs for policy vs gateway controller
**What**

Create a new binary for the policy controller within the MGC repo. This is the first part of separating out the policy controller.

- [x] Add a new main for the policy controller.  Start only the DNS and TLS policy controllers in this main.
- [x] Change the existing main for the Gateway Controller to only start the gateway and OCM reconcilers
- [x] Add flag to turn off DNSPolicy watching managedclusters
- [x] create a new image build in the GH action
- [x] create a new set of deployment resources
- [x] update the integration tests to have 2 separate suites (Policy and Gateway)
- [x] Ensure the quick start and installation guides are updated as needed 
- [ ][ Add integration test for single cluster (ocm not enabled)](https://github.com/Kuadrant/multicluster-gateway-controller/issues/496)
- [x] setup separate deployment
- [x] ensure metrics are exposed and new service monitor present

maleck13:
partially fixes 
#639


cmd/policy_controller/main.go
```diff
@@ -63,28 +47,25 @@ var (
...
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
-	flag.StringVar(&certProvider, "cert-provider", "glbc-ca", "The name of the certificate provider to use")
+
+	flag.BoolVar(&ocmHub, "ocm-hub", true,
+		"tells the controller if it is in an ocm hub"+
+			"setting this to false will cause the controller to stop watching for managedcluster resources")
```
mikenairn, line 68: This seems fine as a temp workaround but once https://github.com/Kuadrant/multicluster-gateway-controller/issues/622 is done I'd expect this to be a more generic option that would allow any gvk to be specified as the "Cluster" resource to watch.

cmd/gateway_controller/main.go
```diff
@@ -0,0 +1,148 @@
...
+
+	//+kubebuilder:scaffold:scheme
+}
+
+func main() {
+	var metricsAddr string
+	var enableLeaderElection bool
+	var probeAddr string
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
```
mikenairn, line 71: This is probably the correct defaults, but just noting that when running this locally we will need to override the values for one of the controllers.

config/policy-controller/kustomization.yaml
```diff
@@ -0,0 +1,9 @@
+resources:
+- namespace.yaml
+- manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: quay.io/kuadrant/multicluster-gateway-controller
```
mikenairn, line 8: Should this be something else?

maleck13: just fixing that now :)

hack/make/policy_controller.make
```diff
@@ -0,0 +1,51 @@
...
+
+.PHONY: build-policy-controller
+build-policy-controller: manifests generate fmt vet ## Build controller binary.
+	go build -o bin/policy_controller ./cmd/policy_controller/main.go
+
+.PHONY: run-policy-controller
+run-policy-controller: manifests generate fmt vet  install
+	go run ./cmd/gateway_controller/main.go \
+	    --metrics-bind-address=:8080 \
+	    --health-probe-bind-address=:8081 \
```
mikenairn, line 14: Should probably change the ports here so both controller can run locally

pkg/controllers/dnspolicy/dnspolicy_dnsrecords.go
```diff
@@ -174,7 +186,7 @@ func getClusterGatewayAddresses(gw *gatewayv1beta1.Gateway) map[string][]gateway
 		var gatewayAddresses []gatewayv1beta1.GatewayAddress
 
 		//Default to Single Cluster (Normal Gateway Status)
-		cluster := "none"
+		cluster := singleCluster
 		addressValue := address.Value
```
mikenairn, line 190: I think the changes here are fine as a temp solution until we get https://github.com/Kuadrant/multicluster-gateway-controller/issues/496, which will hopefully remove the dependency on the name.

Although it is still possible for someone to mess this up, i guess it is less likely they will call there cluster "kudarant.io/single"

.vscode/settings.json
```diff
@@ -1,3 +1,3 @@
 {
-  "go.buildTags": "e2e unit integration"
+  "go.buildTags": "e2e unit integration gateway_integration policy_integration"
```
mikenairn, line 2: I;m not sure you need these extra build tags do you? Looks like you are still using "integration" with tests moved into separate directories.

config/policy-controller/kustomization.yaml
```diff
@@ -0,0 +1,9 @@
+resources:
+- namespace.yaml
+- manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
```
mikenairn, line 7: Think this will need to be policy-controller

.vscode/settings.json
maleck13: yeah cleaned that up

maleck13:
e2e pass here https://github.com/Kuadrant/multicluster-gateway-controller/actions/runs/6669439243

.github/workflows/ci-e2e.yaml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - main
       - "release-*"
+      - separate-policy-controller
```
mikenairn, line 8: @mikenairn ToDo Remove this before merge

.github/workflows/ci-e2e.yaml
```diff
@@ -67,13 +68,18 @@ jobs:
...
           kubectl --context kind-mgc-control-plane logs --all-containers --ignore-errors deployment/mgc-controller-manager -n multicluster-gateway-controller-system
           kubectl get managedzones -n multi-cluster-gateways mgc-dev-mz-aws -o yaml
           kubectl --context kind-mgc-control-plane -n multi-cluster-gateways wait --timeout=60s --for=condition=Ready managedzone/mgc-dev-mz-aws
           kubectl get managedzones -n multi-cluster-gateways mgc-dev-mz-gcp -o yaml
           kubectl --context kind-mgc-control-plane -n multi-cluster-gateways wait --timeout=60s --for=condition=Ready managedzone/mgc-dev-mz-gcp
+      - name: Deploy Policy Controller
+        run: |
+          make docker-build-policy-controller kind-load-policy-controller deploy-policy-controller
+          kubectl --context kind-mgc-control-plane -n multicluster-gateway-controller-system wait --timeout=300s --for=condition=Available deployment/mgc-policy-controller-manager
+          kubectl --context kind-mgc-control-plane logs --all-containers --ignore-errors deployment/mgc-policy-controller-manager -n multicluster-gateway-controller-system
```
mikenairn, line 82: I don't think you need the "Deploy Policy Controller" stage since the deployment of MGC includes the gateway controller.

hack/make/policy_controller.make
```diff
@@ -0,0 +1,51 @@
...
+POLICY_CONTROLLER_IMG ?= policy-controller:$(TAG)
+LOG_LEVEL ?= 3
+
+.PHONY: build-policy-controller
+build-policy-controller: manifests generate fmt vet ## Build controller binary.
+	go build -o bin/policy_controller ./cmd/policy_controller/main.go
+
+.PHONY: run-policy-controller
+run-policy-controller: manifests generate fmt vet  install
+	go run ./cmd/gateway_controller/main.go \
```
mikenairn, line 12:
```suggestion
	go run ./cmd/policy_controller/main.go \
```

.github/workflows/ci-e2e.yaml
maleck13: not sure I follow. What would reconcile the policies if I don't deploy it? Or do you  mean it is already deployed somehow?

.github/workflows/ci-e2e.yaml
mikenairn, line 8: @maleck13 Can be removed. The e2e works, but the PR one will always fail until this is merged since its using the workflow from main.

.github/workflows/ci-e2e.yaml
mikenairn: You are including it in the deployment  of MGC so it's getting deployed when you run `deploy-gateway-controller`. Making me think you need to add `docker-build-policy-controller kind-load-policy-controller` to that stage, although it did seem to work locally :thinking:

mikenairn: OK, it will work because we have the default image set to `quay.io/kuadrant/policy-controller:main`, so we definitely do need to add `docker-build-policy-controller kind-load-policy-controller` to the MGC deploy stage or we will be testing the wrong thing.

maleck13: yep
```
      - name: Deploy
        run: |
          make docker-build-policy-controller kind-load-policy-controller deploy-policy-controller
          make docker-build-gateway-controller kind-load-gateway-controller deploy-gateway-controller
          kubectl --context kind-mgc-control-plane -n multicluster-gateway-controller-system wait --timeout=300s --for=condition=Available deployment/mgc-controller-manager
          kubectl --context kind-mgc-control-plane -n multicluster-gateway-controller-system wait --timeout=300s --for=condition=Available deployment/mgc-policy-controller-manager
          kubectl --context kind-mgc-control-plane logs --all-containers --ignore-errors deployment/mgc-controller-manager -n multicluster-gateway-controller-system
          kubectl --context kind-mgc-control-plane logs --all-containers --ignore-errors deployment/mgc-policy-controller-manager -n multicluster-gateway-controller-system
          kubectl get managedzones -n multi-cluster-gateways mgc-dev-mz-aws -o yaml
          kubectl --context kind-mgc-control-plane -n multi-cluster-gateways wait --timeout=60s --for=condition=Ready managedzone/mgc-dev-mz-aws
          kubectl get managedzones -n multi-cluster-gateways mgc-dev-mz-gcp -o yaml
          kubectl --context kind-mgc-control-plane -n multi-cluster-gateways wait --timeout=60s --for=condition=Ready managedzone/mgc-dev-mz-gcp
```

.github/workflows/ci-e2e.yaml
```diff
@@ -65,11 +65,14 @@ jobs:
         run: |
           export OCM_SINGLE=1
           make local-setup
-      - name: Deploy MGC
+      - name: Deploy
         run: |
-          make docker-build-controller kind-load-controller deploy-controller
+          make docker-build-policy-controller kind-load-policy-controller deploy-policy-controller
```
mikenairn, line 70: I don't think you need `deploy-policy-controller` here since `deploy-gateway-controller` will deploy both anyway

hack/make/policy_controller.make
```diff
@@ -0,0 +1,51 @@
...
+	kind load docker-image ${POLICY_CONTROLLER_IMG} --name mgc-control-plane  --nodes mgc-control-plane-control-plane
+
+.PHONY: docker-push-policy-controller
+docker-push-policy-controller: ## Push docker image with the controller.
+	docker push ${POLICY_CONTROLLER_IMG}
+
+.PHONY: deploy-policy-controller
+deploy-policy-controller: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
+	cd config/policy-controller && $(KUSTOMIZE) edit set image policy-controller=${POLICY_CONTROLLER_IMG}
+	$(KUSTOMIZE) --load-restrictor LoadRestrictionsNone build config/deploy/local | kubectl apply -f -
```
mikenairn, line 33: This is probably not really what we want here. Currently is will do exactly the same as `deploy-gateway-controller`, but only patch the image for policy.

.github/workflows/ci-e2e.yaml
mikenairn: OK, i think for this to work with the current makefiles you need to do all the build and load tasks before you run any deploy one or you will end up with part of the deployment using the wrong thing.

```
```

mikenairn: @maleck13 Can you add back the `separate-policy-controller` branch in this workflow so we can check it is actually working as expected.

hack/make/policy_controller.make
maleck13: What I was thinking with this was we could lift it wholesale over to a new repo. Do you think it is ok to keep as is with that in mind?

.github/workflows/ci-e2e.yaml
maleck13: done

Issue #630, opened 2023-10-19
MGC support for GWAPI v1


maleck13:
**What**

V1 of GWAPI is approaching we need to start testing with the new resources. https://github.com/kubernetes-sigs/gateway-api/releases/tag/v1.0.0-rc2

**How**

- [x] Update our GWAPI resources, install and create TLS and DNS policies 
- [x] update our GWAPI dependencies in the Gateway controller
- [x] #665 
- [x] update kuadrant-operator version which supports GWAPI v1 GA.
- [x] Update Gateway API Resources in the docs files to be V1

laurafitzgerald:
WIP Branch https://github.com/laurafitzgerald/multicluster-gateway-controller/tree/gh-630

laurafitzgerald:
There's some discussion at https://kubernetes.slack.com/archives/C05J0D0V525/p1698851598969379 about an issue I faced due to rules being too costly for httproutes when it comes to parentRef.

kuadrant-operator have logged it here [GEP: Add Name to HTTPRouteRule and HTTPRouteMatch · Issue #995 · kubernetes-sigs/gateway-api](https://github.com/kubernetes-sigs/gateway-api/issues/995#issuecomment-1787005682)

And have a solution on their side https://github.com/Kuadrant/kuadrant-operator/pull/286/commits/56a9a346b38907f93cff17b4935467fd0e5c17a6

It's blocking us from running our integration tests because the environment fails on CRD creation producing the following error 

```
[...]

PR #645, opened 2023-10-27
Gh 630 gateway-api v1.0.0

laurafitzgerald:
Addresses #630 

This is WIP and includes a commit required which also bumps controller-runtime. 

Running this requires the use of the kuadrant-operator branch -> https://github.com/Kuadrant/kuadrant-operator/compare/main...laurafitzgerald:kuadrant-operator:gateway-api-v1 
There has been some discussion around removing the kuadrant-operator dependency but this will work for now with a temporary commit which points the dependency. 

So far e2e tests are passing locally. 

Following the multicluster gateway walkthrough

Setting up for the walkthrough can be done via the [quickstart](https://docs.kuadrant.io/getting-started/) and then updates for gateway-api and crds configs using the following commands.

Scale down the gateway-controller pod which comes with the quickstart `kc edit deployment mgc-controller-manager -n multicluster-gateway-controller-system` and change replicas to 0.
Scale down the policy-controller pod which comes with the quickstart `kc edit deployment mgc-policy-controller -n multicluster-gateway-controller-system` and change replicas to 0.

`kustomize build config/istio --enable-helm --helm-command helm | kubectl --context kind-mgc-control-plane apply -f -`
`kustomize build config/istio --enable-helm --helm-command helm | kubectl --context kind-mgc-workload-1 apply -f -`
`kustomize build config/gateway-api --enable-helm --helm-command helm | kubectl --context kind-mgc-workload-1 apply -f -`
`kustomize build config/gateway-api --enable-helm --helm-command helm | kubectl --context kind-mgc-control-plane apply -f -`

Run controllers locally from this copy

Follow the steps in the walkthrough.
The walkthrough itself was successful


pkg/apis/v1alpha1/dnspolicy_types.go
```diff
@@ -21,16 +21,15 @@ import (
...
+	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
 )
 
 // DNSPolicySpec defines the desired state of DNSPolicy
 type DNSPolicySpec struct {
 
 	// +kubebuilder:validation:Required
 	// +required
-	TargetRef gatewayapiv1alpha2.PolicyTargetReference `json:"targetRef"`
+	ParentRef gatewayapiv1.ParentReference `json:"parentRef"`
```
guicassolato, line 32: Why moving to [`ParentReference`](https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1/shared_types.go#L35)? Is that because of `sectionName`?

My understanding is that this type is for specifying a parent of the resource, not a target of a policy. If it's for the `sectionName`, now there's [`PolicyTargetReferenceWithSectionName`](https://github.com/kubernetes-sigs/gateway-api/blob/68194976485b269e68214b5a40e5951815bbb80b/apis/v1alpha2/policy_types.go#L67) that can be used.

laurafitzgerald: Hi @guicassolato my reasoning for moving is that it appeared to me that parentRef was replacing targetRef as targetRef wasn't available in v1. But I think that I might have misunderstood that usage. I'm digging into it a little bit deeper now.

pkg/policysync/policy_test.go
```diff
@@ -40,11 +40,11 @@ func TestReflectPolicy(t *testing.T) {
 		t.Fatalf("expected targetRef.Kind to be test, got %s", targetRef.Name)
 	}
 
-	reflectPolicy.UpdateTargetRef(func(targetRef *gatewayapiv1alpha2.PolicyTargetReference) {
```
sergioifg94, line 43: I'd keep the `targetRef` parameter in here. It still compiles and works because the targetRef variable is in the scope from the previous assertion, but the call to `UpdateTargetRef` should be independent to that

sergioifg94 (commented):
Code changes look good, just a couple of small details

go.mod
```diff
@@ -5,32 +5,32 @@ go 1.20
...
 	github.com/aws/aws-sdk-go v1.44.175
 	github.com/go-logr/logr v1.2.4
-	github.com/google/uuid v1.3.0
+	github.com/google/uuid v1.3.1
 	github.com/goombaio/namegenerator v0.0.0-20181006234301-989e774b106e
 	github.com/jetstack/cert-manager v1.7.1
-	github.com/kuadrant/authorino v0.10.0
-	github.com/kuadrant/kuadrant-operator v0.1.1-0.20230323151616-58593d01833a
+	github.com/kuadrant/authorino v0.15.0
+	github.com/laurafitzgerald/kuadrant-operator v0.0.0-20231103170836-625c57c4769f
```
mikenairn, line 12: @laurafitzgerald fyi, you'd be better adding a replace line in here that way you don't have to update every import to test changes from your branch:

```
replace github.com/kuadrant/kuadrant-operator => github.com/laurafitzgerald/kuadrant-operator branhc-with-my-needed-chnages
```

laurafitzgerald: ah sweet. didn't know about this. so it wont error on the module name in this case?

go.mod
```diff
@@ -5,32 +5,32 @@ go 1.20
 require (
 	github.com/aws/aws-sdk-go v1.44.175
 	github.com/go-logr/logr v1.2.4
-	github.com/google/uuid v1.3.0
+	github.com/google/uuid v1.3.1
 	github.com/goombaio/namegenerator v0.0.0-20181006234301-989e774b106e
 	github.com/jetstack/cert-manager v1.7.1
-	github.com/kuadrant/authorino v0.10.0
-	github.com/kuadrant/kuadrant-operator v0.1.1-0.20230323151616-58593d01833a
+	github.com/kuadrant/authorino v0.15.0
```
mikenairn, line 11: No idea why authorino is in here, separate issue but this controller should not depend directly on it. Seemec to get added here for some reason https://github.com/Kuadrant/multicluster-gateway-controller/commit/a75da134f7595a6600dbe3c48f97bb172a295988#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R11

mikenairn: Will be removed here https://github.com/Kuadrant/multicluster-gateway-controller/pull/629/commits/62b95d76a9f19a11c420b81926cfe3e09ff8ace4

laurafitzgerald: that one is added automatically by go. when I used the approach above authorino was removed.

hack/make/dependencies.make
```diff
@@ -29,7 +29,7 @@ CONTROLLER_TOOLS_VERSION ?= v0.10.0
 KIND_VERSION ?= v0.17.0
 HELM_VERSION ?= v3.10.0
 YQ_VERSION ?= v4.30.8
-ISTIOVERSION ?= 1.17.0
+ISTIOVERSION ?= 1.20.0-rc.0
```
maleck13, line 32: so question for me is how do we ensure we bump this once released? I don't want to hold open the PR if we don't need to  but also don't want to forget @laurafitzgerald perhaps we could add an issue and bump this and move to top of backlog so we always see it?

maleck13 (commented):
This change looks good. Two minor things

1) issue to track moving to the rlease of 1.20 of istio
2) issue to investigate re-enabling the CRD validation

sergioifg94:
@maleck13 I repurposed the issue that I created to fix the integration tests to re-enable the CRD validation: #658

mikenairn:
/hold

This all looks good to me, i just want to make sure we don't accidentally merge with this still here https://github.com/Kuadrant/multicluster-gateway-controller/pull/645/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R32

hack/make/dependencies.make
laurafitzgerald: Tracking that here https://github.com/Kuadrant/multicluster-gateway-controller/issues/665

laurafitzgerald:
> /hold
> 
> This all looks good to me, i just want to make sure we don't accidentally merge with this still here https://github.com/Kuadrant/multicluster-gateway-controller/pull/645/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R32

We require [this pr](https://github.com/Kuadrant/kuadrant-operator/pull/286) to merge to avoid merging this

mikenairn:
As discussed we should follow suit and bump the go version to 1.21 https://github.com/Kuadrant/kuadrant-operator/pull/286/commits/05fe71afffdbc03d6929a513acce4972286a12e0.

Other than that it all looks good to go :+1:

mikenairn:
Trying this locally, but changes all look good to me. Will also just verify the e2e test still passes with this change https://github.com/Kuadrant/multicluster-gateway-controller/pull/674 so we can get this merged ASAP.

mikenairn:
/hold cancel

Verified locally, all working as expected.

PR #648, opened 2023-11-03
Separate controller deploy

maleck13:
draft at separating out the policy controller deployment to allow it to be deployed independently


config/default/kustomization.yaml
```diff
@@ -10,30 +6,7 @@ namespace: multicluster-gateway-controller-system
...
-
 resources:
-- ../crd
-- ../rbac
-- namespace.yaml
 - ../manager
 - ../add-on-manager
 - ../policy-controller
-patches:
-- path: manager_metrics_patch.yaml
```
mikenairn, line 39: I would avoid changing this, and there really shouldn't be a need to. The point of default is it's an overlay that is specifying the namespace, prefix etc.. to use and calls on the individual parts (which should not specify a namespace)

This his how all kubebuilder controllers are set up see https://github.com/Kuadrant/kuadrant-operator/blob/main/config/default/kustomization.yaml

config/manager/kustomization.yaml
```diff
@@ -1,8 +1,15 @@
...
+# Adds namespace to all resources.
+namespace: multicluster-gateway-controller-system
+
 images:
 - name: controller
   newName: quay.io/kuadrant/multicluster-gateway-controller
   newTag: main
+
+patches:
+- path: manager_metrics_patch.yaml
```
mikenairn, line 15: This should probably be reverted, it shouldn't include rbac here, or set a namespace.

manager_metrics_patch.yaml is a patch that should be called from outside of this kustomization  doing it here is making it redundant and you might as well just update manager.yaml

config/manager/namespace.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: system
```
mikenairn, line 4: I'm not really sure how we ended up with a namesapce resource, but if we are updating all this I'd probably prefer we just put it back where it's usually found in the manager yaml see https://github.com/Kuadrant/kuadrant-operator/blob/main/config/manager/manager.yaml

config/policy-controller/kustomization.yaml
```diff
@@ -1,8 +1,21 @@
...
   newTag: main
+
+
+resources:
+- ./namespace.yaml
+- ./manager.yaml
+- ../crd
+- ./rbac
+
+namespace: kuadrant-policy-controller
```
mikenairn, line 21: Personally i think you would be better setting this up like it would be if it was it's own controller, and have the expected config structure for policy-controller setup in here (config/policy-controller). When you come to move it to it's own repo its just a matter of moving directory as is.

```
config/policy-controller
├── crd
│   ├── bases
│   │   ├── kuadrant.io_dnshealthcheckprobes.yaml
│   │   ├── kuadrant.io_dnspolicies.yaml
│   │   ├── kuadrant.io_dnsrecords.yaml
│   │   ├── kuadrant.io_managedzones.yaml
│   │   └── kuadrant.io_tlspolicies.yaml
│   ├── kustomization.yaml
│   ├── kustomizeconfig.yaml
│   └── patches
│       └── policy-patch.yaml
├── default
│   ├── issuer.yaml
│   ├── kustomization.yaml
│   └── namespace.yaml
├── deploy
│   └── local
│       ├── kustomization.yaml
│       └── manager_config_patch.yaml
├── kustomization.yaml
├── manager
│   ├── kustomization.yaml
│   └── manager.yaml
├── manager.yaml
└── rbac
    ├── dnsrecord_editor_role.yaml
    ├── dnsrecord_viewer_role.yaml
    ├── kustomization.yaml
    ├── leader_election_role_binding.yaml
    ├── leader_election_role.yaml
    ├── managedzone_editor_role.yaml
    ├── managedzone_viewer_role.yaml
    ├── role_binding.yaml
    ├── role.yaml
    ├── service_account.yaml
    ├── tlspolicy_editor_role.yaml
    └── tlspolicy_viewer_role.yaml
```

Does it not complain about `../crd` being outside the directory?

docs/demos/dns-policy/dnspolicy-demo.md
```diff
@@ -14,7 +14,13 @@
 
 ```bash
 ./install.sh
-(export $(cat ./controller-config.env | xargs) && export $(cat ./aws-credentials.env | xargs) && make build-controller install run-controller)
+(export $(cat ./controller-config.env | xargs) && export $(cat ./aws-credentials.env | xargs) && make build-gateway-controller install run-gateway-controller)
+
+## separate window
+
+(export $(cat ./controller-config.env | xargs) && export $(cat ./aws-credentials.env | xargs) && make build-policy-controller install run-policy-controller)
```
mikenairn, line 21: You can remove `(export $(cat ./controller-config.env | xargs) && export $(cat ./aws-credentials.env | xargs)` from these commands, it's not been needed for a while since the controller no longer uses the env vars

config/add-on-manager/manager.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: kuadrant-add-on-manager
-  namespace: system
+  namespace: multicluster-gateway-controller-system
```
mikenairn, line 5: This should always be getting called from an overlay with `namespace: foo` set so you don't need to hard code the namespace here

config/policy-controller/manager.yaml
```diff
@@ -59,5 +59,5 @@ spec:
           requests:
             cpu: 10m
             memory: 64Mi
-      serviceAccountName: controller-manager
+      serviceAccountName: policy-controller
```
mikenairn, line 62: You likely will end up with a conflict here when pulling these into the mgc manifests, so probably do need to rename these, but maybe for consistency we could just prepend "policy" to the usual names? "policy-controller-manager", "policy-manager-role" etc... ?

config/policy-controller/default/kustomization.yaml
```diff
@@ -0,0 +1,21 @@
...
+# names of all resources, e.g. a deployment named
+# "wordpress" becomes "alices-wordpress".
+# Note that it should also match with the prefix (text before '-') of the namespace
+# field above.
+namePrefix: kuadrant-
+
+images:
+- name: policy-controller
+  newName: quay.io/kuadrant/policy-controller
+  newTag: main
```
mikenairn, line 12: images section should be in `config/policy-controller/manager/kustomization.yaml`

config/policy-controller/default/kustomization.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+
+resources:
+- ../manager
+- ./issuer.yaml
```
mikenairn, line 17: Where did we create this issuer before? Not sure it was included in the default resources.

maleck13: I am not sure TBH.

maleck13:
passing e2e https://github.com/Kuadrant/multicluster-gateway-controller/actions/runs/6769481175

.github/workflows/policy-controller-image.yaml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - main
       - "release-*"
+      - separate-controller-deploy
```
mikenairn, line 8: remove before merge

Makefile
```diff
@@ -117,13 +125,6 @@ ifndef ignore-not-found
   ignore-not-found = false
 endif
 
-.PHONY: install
-install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
-	$(KUSTOMIZE) build config/crd | kubectl apply -f -
-
-.PHONY: uninstall
-uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
-	$(KUSTOMIZE) build config/crd | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
```
mikenairn, line 126: Any reason to remove these?

config/policy-controller/rbac/leader_election_role.yaml
```diff
@@ -0,0 +1,45 @@
...
+  labels:
+    app.kubernetes.io/name: role
+    app.kubernetes.io/instance: policy-controller-leader-election-role
+    app.kubernetes.io/component: rbac
+    app.kubernetes.io/created-by: policy-controller
+    app.kubernetes.io/part-of: kuadrant
+    app.kubernetes.io/managed-by: kustomize
+  name: policy-controller-leader-election-role
+  namespace: system
+rules:
```
mikenairn, line 14: We should probably get rid of all these labels.

mikenairn: Reason being, they look odd here https://github.com/Kuadrant/kuadrant-operator/pull/293/files#diff-d40f82c17f85444a248492d3d46940851c2bf53af0e6c00dfd3e8e2782e0f936R8. You'd need to patch then there, but not sure they offer much anyway. Think they were added by kubebuilder during generation.

config/policy-controller/default/kustomization.yaml
```diff
@@ -0,0 +1,16 @@
+
+# Value of this field is prepended to the
+# names of all resources, e.g. a deployment named
+# "wordpress" becomes "alices-wordpress".
+# Note that it should also match with the prefix (text before '-') of the namespace
+# field above.
+namePrefix: kuadrant-
```
mikenairn, line 7: Does "kuadrant" make sense a prefix? This gets added to the names of all resources, so when you are including it from other kustomizations which also use a namePrefix, it prepends them so you end up with the likes of this in kuadrant-operator `kuadrant-operator-kuadrant-policy-controller-5bbb7545bf-vd5jw`.

Maybe we don't need a prefix here since it's always included by something else that is adding one?

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -486,18 +484,6 @@ func buildAcceptedCondition(generation int64, acceptedStatus metav1.ConditionSta
...
-		&v1alpha1.DNSPolicy{},
-		policy.POLICY_TARGET_REF_KEY,
-		func(obj client.Object) []string {
-			return []string{policy.GetTargetRefValueFromPolicy(obj.(*v1alpha1.DNSPolicy))}
-		},
-	)
-	if err != nil {
-		return err
-	}
-
```
mikenairn, line 500: :+1: No longer needed.

pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -69,6 +69,9 @@ type DNSPolicyReconciler struct {
 //+kubebuilder:rbac:groups=kuadrant.io,resources=dnspolicies/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=kuadrant.io,resources=dnspolicies/finalizers,verbs=update
 //+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclusters,verbs=get;list;watch
+// +kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gateways,verbs=get;list;watch;create;update;patch;delete
```
mikenairn, line 72: Probably doesn't need some of these permissions, can't think it will ever be using create and delete anyway. I guess it might update status.

Makefile
mikenairn: nvm, gateway-controller has no crds.

pkg/controllers/tlspolicy/tlspolicy_controller.go
```diff
@@ -70,6 +70,8 @@ type TLSPolicyReconciler struct {
 //+kubebuilder:rbac:groups=kuadrant.io,resources=tlspolicies/finalizers,verbs=update
 //+kubebuilder:rbac:groups="cert-manager.io",resources=issuers,verbs=get;list;
 //+kubebuilder:rbac:groups="cert-manager.io",resources=clusterissuers,verbs=get;list;
+// +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;delete
```
mikenairn, line 73: Do we need delete?

pkg/controllers/tlspolicy/tlspolicy_controller.go
```diff
@@ -70,6 +70,8 @@ type TLSPolicyReconciler struct {
...
+// +kubebuilder:rbac:groups="cert-manager.io",resources=certificates,verbs=get;list;watch;create;update;patch;delete
```
mikenairn, line 74: Formatting looks a bit off, extra space.

mikenairn:
I think this looks fine, just some questions about the naming of things, and i think we might want to consider removing all the generated labels from the resources.

.github/workflows/ci-e2e.yaml
```diff
@@ -69,9 +70,9 @@ jobs:
...
           kubectl --context kind-mgc-control-plane -n multicluster-gateway-controller-system wait --timeout=300s --for=condition=Available deployment/mgc-controller-manager
-          kubectl --context kind-mgc-control-plane -n multicluster-gateway-controller-system wait --timeout=300s --for=condition=Available deployment/mgc-policy-controller-manager
+          kubectl --context kind-mgc-control-plane -n multicluster-gateway-controller-system wait --timeout=300s --for=condition=Available deployment/mgc-kuadrant-policy-controller
           kubectl --context kind-mgc-control-plane logs --all-containers --ignore-errors deployment/mgc-controller-manager -n multicluster-gateway-controller-system
-          kubectl --context kind-mgc-control-plane logs --all-containers --ignore-errors deployment/mgc-policy-controller-manager -n multicluster-gateway-controller-system
+          kubectl --context kind-mgc-control-plane logs --all-containers --ignore-errors deployment/mgc-kuadrant-policy-controller -n multicluster-gateway-controller-system
```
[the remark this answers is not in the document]
maleck13: yeah ok will change

pkg/controllers/dnspolicy/dnspolicy_controller.go
maleck13: updated

pkg/controllers/tlspolicy/tlspolicy_controller.go
maleck13: updated

config/policy-controller/rbac/leader_election_role.yaml
maleck13: got rid of them in the policy controller

maleck13:
https://github.com/Kuadrant/multicluster-gateway-controller/actions/runs/6809444407/job/18515739375

mikenairn:
This looks fine to me. 

It was annoying me that we had to deviate from normal naming in the config/policy-controller files since i would have thought specifying a different namePrefix would have been sufficient, but looks as though we would have to do something more like [this](https://github.com/Kuadrant/multicluster-gateway-controller/commit/81d8c21376015d02e248a24c55589fc96a16bb53) for just using namePrefix to work.

We might consider changing to this later, but not essential now.

mikenairn (approved):

PR #649, opened 2023-11-03
remove http check for wildcard certificate in favour of checking certificates

laurafitzgerald:
$subject
As discussed here -> https://kubernetes.slack.com/archives/C05J0D0V525/p1699520931987239

### DIFF
--- a/test/e2e/gateway_single_spoke_test.go
+++ b/test/e2e/gateway_single_spoke_test.go
@@ -11,13 +11,14 @@ import (
 	"strings"
 	"time"
 
-	cmmetav1 "github.com/jetstack/cert-manager/pkg/apis/meta/v1"
+	certmanv1 "github.com/jetstack/cert-manager/pkg/apis/certmanager/v1"
+	certmanmetav1 "github.com/jetstack/cert-manager/pkg/apis/meta/v1"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gstruct"
 	ocm_cluster_v1beta1 "open-cluster-management.io/api/cluster/v1beta1"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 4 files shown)


test/util/helper.go
```diff
@@ -148,5 +148,5 @@ func TestCertificate(dnsName string, resp *http.Response) error {
 			}
 		}
 	}
-	return fmt.Errorf("wildcard hostname not found in the certificate via get request")
+	return fmt.Errorf("%s not found in the certificate via get request", dnsName)
```
mikenairn, line 151: Should we just remove all usages of this TestCertificate function i.e. https://github.com/Kuadrant/multicluster-gateway-controller/blob/3dfbf32c018cbfcddd8d036763c8dbf71ef01a08/test/e2e/gateway_single_spoke_test.go#L354

test/e2e/gateway_single_spoke_test.go
```diff
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	v1 "github.com/jetstack/cert-manager/pkg/apis/certmanager/v1"
```
mikenairn, line 14: nit: Rename this to `certmanv1`

test/e2e/gateway_single_spoke_test.go
```diff
@@ -398,31 +417,28 @@ var _ = Describe("Gateway single target cluster", func() {
...
+									for _, dnsName := range cert.Spec.DNSNames {
+										if dnsName == string(wildcardHostname) {
+											return nil
+										}
+									}
+								}
+							}
+							return fmt.Errorf("dns names for certificate not as expected")
+						}).WithContext(ctx).WithTimeout(180 * time.Second).WithPolling(2 * time.Second).ShouldNot(HaveOccurred())
 					}
```
mikenairn, line 442: Lately I've been trying to better use the gomega matchers in these tests, what about trying somethign like this:

```.go
By("checking tls certificate")
{
	Eventually(func(g Gomega, ctx context.Context, expected ...string) {
		certList := &certmanv1.CertificateList{}
		err := tconfig.HubClient().List(ctx, certList)
		g.Expect(err).NotTo(HaveOccurred())
		g.Expect(certList.Items).To(Not(BeEmpty()))
		g.Expect(certList.Items).Should(ContainElement(
			MatchFields(IgnoreExtras, Fields{
				"ObjectMeta": MatchFields(IgnoreExtras, Fields{
					"Labels": HaveKeyWithValue("gateway", testID),
				}),
				"Spec": MatchFields(IgnoreExtras, Fields{
					"DNSNames": ConsistOf(string(hostname), string(wildcardHostname)),
				}),
			})))
	}).WithContext(ctx).Should(Succeed())
}
```

mikenairn:
I think we need to look at these tests a bit and decide what the responsibility of each test suite (integration/e2e) is as some of the checks we are doing in the e2e look like the might be better elsewhere. Also probably want to look at creating separate more focused e2e test files for the TLS specific tests.

/lgtm
/approve

Issue #411, opened 2023-08-08
remove kube-rbac-proxy sidecar


mikenairn:
Remove the kube-rbac-proxy container see https://github.com/Kuadrant/kuadrant-operator/pull/33

philbrookes:
This issue is stale because it has been open for 60 days with no activity.

philbrookes:
This issue was closed because it has been inactive for 30 days since being marked as stale.

philbrookes:
This issue is stale because it has been open for 60 days with no activity.

philbrookes:
This issue was closed because it has been inactive for 30 days since being marked as stale.

PR #389, opened 2023-07-31
Add Istio Workload Dashboard

david-martin:
Closes #307 

Changes:

* Move the grafana kustomize configuration alongside the kube-prometheus configuration to allow patching the grafana deployment with new dashboard configmaps
* Add the [Isito Workload Dashboard](https://grafana.com/grafana/dashboards/7630-istio-workload-dashboard/) json 
* Enable scraping of MGC metrics (ServiceMonitor added via kustomize)
* Add dev docs for how to add/update dashboards

TODO:

* [x] Validate the panels & expressions on the Istio Workload Dashboard behave as expected.
* [x] Document (in config/prometheus-for-federation ?) the steps for importing, updating and reloading a dashboard

Follow Up:
* Consider a second (or more) dashboard that are tailored more at the Multi Cluster Gateway administrator persona - e.g. aggregation across gateway instances that isn't shown on the Istio workload dashboard, anything linked from https://github.com/Kuadrant/multicluster-gateway-controller/issues/307#issuecomment-1628613470 that would be good to show, any DNS or TLS metrics exposed from mgc that can be visualised (or if none available, suggest some that would be good for that persona), cert-manager metrics (could be an existing dashboard)

*Use cases for an Multi Cluster Gateway administrator persona*

Please add or suggest more in the comments

*  As a Multi Cluster Gateway administrator, I want to see how traffic going to my application across multiple clusters is being balanced (e.g. is weighting behaving as expected)
*  As a Multi Cluster Gateway administrator, I want to see where, geographically, traffic is coming from to my app deployed in multiple regions (e.g. is traffic using geo dns as expected)

*Notes on importing an existing dashboard*

Grafana > Import > Import via grafana.com > input the dashboard ID *OR* paste json > Load > change any dropdowns until data is shown

*Notes on exporting a dashboard as json*

* Share symbol at top > Export > Export as json > save to a .json file in config/prometheus-for-federation
* Update ./config/prometheus-for-federation/kustomization.yaml to add a new configMapGenerator for the new dashboard
* Update ./config/prometheus-for-federation/grafana_deployment_patch.yaml to add the new configmap as a volume mount

*Notes on reloading a dashboard after changing the json content*

```
# re-build with kustomize, filter out the dashboard configmap & grafana deployment and apply changes
[...]

PR #654, opened 2023-11-06
Fix metrics service port

averevki:
Fix mapping for metrics service port:
https://github.com/Kuadrant/multicluster-gateway-controller/blob/aa223edcfff267b738e63f0ac4798725a1bfeeff/config/default/manager_metrics_patch.yaml#L19-L21


bundle/manifests/mgc-controller-manager-metrics-service_v1_service.yaml
```diff
@@ -16,7 +16,7 @@ spec:
   - name: https
     port: 8443
     protocol: TCP
-    targetPort: https
+    targetPort: metrics
```
maleck13, line 19: If I am not mistaken this is copied to the bundle from https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/config/prometheus/service.yaml

averevki: can you elaborate a little more please?

maleck13: The definition for this lives in the config directory. This definition is used when we do a `make bundle` to create the bundle

david-martin:
@averevki This change might conflict with, or already be solved by the Service/ServiceMonitor in https://github.com/Kuadrant/multicluster-gateway-controller/tree/main/config/prometheus

Those resources are used by https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/hack/quickstart-metrics.sh in this [metrics walkthrough](https://docs.kuadrant.io/multicluster-gateway-controller/docs/how-to/metrics-walkthrough/)

What problem are you seeing that led to you needing the change in this PR?

averevki:
> It looks like it should have been removed. See comment thread [#389 (comment)](https://github.com/Kuadrant/multicluster-gateway-controller/pull/389#discussion_r1284163440) and the (probably incorrectly closed automatically, but I now reopened) issue #411

I see... Then this PR is irrelevant and feel free to close it. Thank you for the explanation :+1:

PR #667, opened 2023-11-10
fix perms around issuers

maleck13:
fix `pkg/mod/k8s.io/client-go@v0.28.3/tools/cache/reflector.go:229: Failed to watch *v1.ClusterIssuer: unknown (get clusterissuers.cert-manager.io)`


maleck13:
@mikenairn tested this locally and not seeing the error any more so looks good as the fix

maleck13:
added a fix for `make deploy-policy-controller`

mikenairn:
I'm not totally sure why we haven't seen this error before (e2e tests), but it does appear that when using a cached client you also need the watch perm to get a resource, so ....

/lgtm
/approve

PR #674, opened 2023-11-15
Fix gateway controller deploy target

mikenairn:
`make deploy-gateway-controller` deploys both the gateway and policy controller but was only updating the image for the gateway controller to the locally built version. This change updates the make targets to ensure that both the gateway and policy controller images are updated when testing the combined gateway/policy controller installation (e2e tests).

When deploying the controller using a local image loaded into kind, we also need to change the imagePullPolicy to Never or it will never start.

**Verification**

```
make local-setup OCM_SINGLE=true
```
Check the gateway and policy deployments are both using the local image
```
kubectl get deployments/mgc-policy-controller -n multicluster-gateway-controller-system -o json| jq .spec.template.spec.containers[0].image
"policy-controller:latest"
kubectl get deployments/mgc-controller-manager -n multicluster-gateway-controller-system -o json| jq .spec.template.spec.containers[0].image
"controller:latest"
```


.gitignore
```diff
@@ -15,6 +15,8 @@ vendor/
 
 #Ignore local directory (for stuff to keep locally)
 local/
+!config/deploy/local
```
laurafitzgerald, line 18: Is this definitely needed? Is it to negate the above `local/`? Seems that git will pick up those files even without these entries.

config/policy-controller/deploy/local/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
+# Local deployment overlay.
```
laurafitzgerald, line 1: This could be a nit: 
And not essential for this change. 

For a number of things I'd suggest some naming updates.
For example the 
- config/deploy is referring to gateway-controller
- config/policy-controller is referring to policy-controller
Also
- mgc-controller-manager is used for deployment/pod naming for gateway-controller
- mgc-policy-controller is used for deployment/pod naming for policy-controller

technically mgc-policy-controller could be expanded to multicluster-gateway-controller-policy-controller which could cause some confusion. maybe the prefix isn't appropriate anymore.

laurafitzgerald (commented):
@mikenairn left a couple of questions inline. none are blockers for merging this.

I've completed the verification steps and the expected images are running in cluster.

.gitignore
mikenairn: If you try and add a directory anywhere called "local" the `local/` in the gitignore prevents you from ever being able to add it. Seems like it doesn't care if it's already committed, so `config/deploy/local` was fine, i just added it for consistency.

laurafitzgerald: 👍 i only tried with editing.

config/policy-controller/deploy/local/kustomization.yaml
mikenairn: > This could be a nit: And not essential for this change.
> 
> For a number of things I'd suggest some naming updates. For example the
> 
>     * config/deploy is referring to gateway-controller
> 
>     * config/policy-controller is referring to policy-controller

If it was intending to stay this way I'd agree, something like `config/policy-controller ` and `config/gateway-controller ` would make more sense, but moving the policy config was a first step and will hopefully soon be moved out of this repo entirely.

The policy-controller will eventually be it's own repo, and the current contents of `config/policy-controller` should end up being the contents of `config` in that repo and removed entirely from here. 

>       Also
> 
>     * mgc-controller-manager is used for deployment/pod naming for gateway-controller
> 
>     * mgc-policy-controller is used for deployment/pod naming for policy-controller
> 

I agree there were some naming inconsistencies brought in during the last couple of PRs (https://github.com/Kuadrant/multicluster-gateway-controller/pull/644, https://github.com/Kuadrant/multicluster-gateway-controller/pull/648, https://github.com/Kuadrant/multicluster-gateway-controller/pull/667) that need resolved. Not really my intention to try and fix all of that here though.

> 
> technically mgc-policy-controller could be expanded to multicluster-gateway-controller-policy-controller which could cause some confusion. maybe the prefix isn't appropriate anymore.

I'm not sure what this should end up being, but it should have a prefix i think that is added to all resources created as part of this deployment. Currently that is mgc.

laurafitzgerald: > > This could be a nit: And not essential for this change.
> > For a number of things I'd suggest some naming updates. For example the
> > ```
> > * config/deploy is referring to gateway-controller
> > 
> > * config/policy-controller is referring to policy-controller
> > ```
> 
> If it was intending to stay this way I'd agree, something like `config/policy-controller ` and `config/gateway-controller ` would make more sense, but moving the policy config was a first step and will hopefully soon be moved out of this repo entirely.
> 
> The policy-controller will eventually be it's own repo, and the current contents of `config/policy-controller` should end up being the contents of `config` in that repo and removed entirely from here.
> 

cool. no need for this change then. 

> > ```
> >   Also
> > 
> > * mgc-controller-manager is used for deployment/pod naming for gateway-controller
> > 
> > * mgc-policy-controller is used for deployment/pod naming for policy-controller
> > ```
> 
> I agree there were some naming inconsistencies brought in during the last couple of PRs (#644, #648, #667) that need resolved. Not really my intention to try and fix all of that here though.

yeah that's fair. I think it would be good to address them.  And it's not essential to fix them here. We could create a follow on issue about it? 

> 
> > technically mgc-policy-controller could be expanded to multicluster-gateway-controller-policy-controller which could cause some confusion. maybe the prefix isn't appropriate anymore.
> 
> I'm not sure what this should end up being, but it should have a prefix i think that is added to all resources created as part of this deployment. Currently that is mgc.

maleck13 (approved):

PR #636, opened 2023-10-23
WIP Use DNSProvider secret ref in DNSPolicy

mikenairn:
closes #715 

Removes the requirement for a ManagedZone to be configured for a host in order for the DNSPolicy to create dns records for it.

Implements the `providerRef` option for DNSPolicy, ManagedZone and DNSRecord as described here https://github.com/Kuadrant/architecture/blob/main/rfcs/0005-single-cluster-dnspolicy.md

Todo:
[] Update docs

PR #675, opened 2023-11-15
Update e2e GH Workflow

mikenairn:
Add TEST_MANAGED_ZONE_NAME env var to e2e test suite steps. This will be required for
https://github.com/Kuadrant/multicluster-gateway-controller/pull/636, but adding these here now to make life easier testing that PR.


.github/workflows/ci-e2e.yaml
```diff
@@ -68,6 +68,8 @@ jobs:
       - name: Deploy
         run: |
           make docker-build-gateway-controller kind-load-gateway-controller docker-build-policy-controller kind-load-policy-controller deploy-gateway-controller
+          kubectl --context kind-mgc-control-plane -n multicluster-gateway-controller-system get deployments/mgc-controller-manager -o yaml | yq .spec.template.spec.containers[0].image
```
laurafitzgerald, line 71: I didn't see this log in the e2e test run. I would have expected to see it after line 254 and before 256 below. Is that an indication something isn't running as expected?
Screenshot 2023-11-15 at 16 36 38

laurafitzgerald: ah i think this would be expected. 
the config changes need to merge before being applied right?

PR #680, opened 2023-11-20
make routingStrategy optional

maleck13:


mikenairn:
/close

Issue was caused by an older version of kubectl being used.

PR #686, opened 2023-11-21
update role with managedcluster perms

maleck13:


pkg/controllers/gateway/gateway_controller.go
```diff
@@ -84,6 +84,7 @@ type GatewayPlacer interface {
 // +kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=placementdecisions,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;delete
 // +kubebuilder:rbac:groups="cert-manager.io",resources=certificates,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclusters,verbs=get;list;watch
```
mikenairn, line 87: Used here https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/placement/ocm.go#L183, so fair enough.

PR #691, opened 2023-11-22
Install guide updates

maleck13:
Fix issue in docs with no add-ons enabled via ocm
Fix issue with glbc ClusterIssuer being installed
Fix issue with install check not looking at policy controller
Fix Gateway API version

Add Context to hub commands - after ocm quickstart the context is set to the last spoke cluster.
Remove one of the options for adding the `RawFeedbackJsonString` - option 1 was problematic on mac. also if the user follows https://open-cluster-management.io/getting-started/quick-start/ some of the steps that were in option 1 were done by the installer. 
Move the installation of the cert-manager step above (optional with the removal of the issuer during installation)
Improve the flow of creating the ManagedZone section/commands

### DIFF
--- a/docs/installation/control-plane-installation.md
+++ b/docs/installation/control-plane-installation.md
@@ -38,7 +38,7 @@ All OCM spoke clusters must be configured with the `RawFeedbackJsonString` featu
 First, run the following command in the context of your hub cluster to install the Gateway API CRDs:
 
 ```bash
-kubectl apply -k "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.6.2"
+kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.0.0/standard-install.yaml"
 ```
 
 We can then add a `wait` to verify the CRDs have been established:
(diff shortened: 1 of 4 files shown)


docs/installation/control-plane-installation.md
```diff
@@ -54,19 +55,20 @@ customresourcedefinition.apiextensions.k8s.io/httproutes.gateway.networking.k8s.
 Then run the following command to install the MGC:
 
 ```bash
-kubectl apply -k "github.com/kuadrant/multicluster-gateway-controller.git/config/mgc-install-guide?ref=main"
+kubectl apply -k "github.com/kuadrant/multicluster-gateway-controller.git/config/mgc-install-guide?ref=release-0.2"
```
maleck13, line 58: def need a better way of handling these updates. Will create an issue

docs/installation/control-plane-installation.md
```diff
@@ -5,46 +5,51 @@ This guide will show you how to install and configure the Multi-Cluster Gateway
...
+By configure HUB_CLUSTER which will be used in the commands:
 
-    ```bash
-    kubectl patch klusterlet klusterlet --type merge --patch '{"spec": {"workConfiguration": {"featureGates": [{"feature": "RawFeedbackJsonString", "mode": "Enable"}]}}}' --context <EACH_SPOKE_CLUSTER>
-    ```
+```bash
+export HUB_CLUSTER=<hub-cluster-name>
+```
+
+## Install Cert-Manager
```
laurafitzgerald, line 32: Moving this up here might no longer be necessary given the removal of the certissuer from the kustomize.

maleck13: It is prob still needed as if they want to use TLSPolicy they still need to install it

PR #692, opened 2023-11-22
[WIP] GH-641: Add Enforced condition to DNSPolicy status

### LINKED ISSUE #641: Add `Enforced` status condition to DNSPolicy
**What**

Based on https://github.com/Kuadrant/architecture/blob/main/rfcs/0004-policy-status.md we want to have some consistent status in place. 

One of those status conditions is the Ready condition which is already present in DNSPolicy. 

But we want to add a new condition `Enforced` this reflects the state of enforcement of the policy

- If there is no error and everything is in place (Ie all sub objects such as DNSPolicy and they are ready) then we move enforced to true
- If there is some form of error we mark enforced as unknown

**Done**

- [ ] Add logic to set the Enforced condition. DNSRecord should be in a ready state
- [ ] Add new tests where appropriate 
- [ ] Update the DNSPolicy reference doc to call out the new status conditions https://docs.kuadrant.io/multicluster-gateway-controller/docs/dnspolicy/dns-policy/
- [ ] add a message to state that the policy is in an `unknown` state with a message of [resource x is not ready]

sergioifg94:
## Description

> Closes #641 

TODO:
- [x] Add tests
- [ ] Decide approach to handle target ref not found
- [ ] Decide approach to handle unexpected errors when calculating the status


pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -274,6 +294,47 @@ func (r *DNSPolicyReconciler) readyCondition(targetNetworkObjectectKind string,
 	return cond
 }
 
+func (r *DNSPolicyReconciler) enforcedCondition(ctx context.Context, targetNetworkObjectKind string, dnsPolicy *v1alpha1.DNSPolicy) (*metav1.Condition, error) {
```
laurafitzgerald, line 297:
```suggestion
func (r *DNSPolicyReconciler) enforcedCondition(ctx context.Context, dnsPolicy *v1alpha1.DNSPolicy) (*metav1.Condition, error) {
```

pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -274,6 +294,47 @@ func (r *DNSPolicyReconciler) readyCondition(targetNetworkObjectectKind string,
...
+
+	notReadyRecords := slice.Filter(dnsRecords, slice.Not((v1alpha1.DNSRecord).IsReady))
+	allRecordsReady := len(notReadyRecords) == 0
+
+	if allRecordsReady {
+		return &metav1.Condition{
+			Type:    string(conditions.ConditionTypeEnforced),
+			Status:  metav1.ConditionTrue,
+			Message: "All DNSRecords ready",
+			Reason:  "DNSPolicyAllRecordsReady",
```
laurafitzgerald, line 322: I think this should be "Enforced" based on https://github.com/Kuadrant/architecture/blob/main/rfcs/0004-policy-status.md#conditions

pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -274,6 +294,47 @@ func (r *DNSPolicyReconciler) readyCondition(targetNetworkObjectectKind string,
...
+
+	notReadyRecordNames := slice.Map(notReadyRecords, func(dnsRecord v1alpha1.DNSRecord) string {
+		return fmt.Sprintf("%s/%s", dnsRecord.Namespace, dnsRecord.Namespace)
+	})
+
+	return &metav1.Condition{
+		Type:    string(conditions.ConditionTypeEnforced),
+		Status:  metav1.ConditionUnknown,
+		Message: fmt.Sprintf("DNSRecords %v not Ready", notReadyRecordNames),
+		Reason:  "DNSPolicySomeRecordsNotReady",
```
laurafitzgerald, line 334: The docs https://github.com/Kuadrant/architecture/blob/main/rfcs/0004-policy-status.md#conditions would suggest this is unknown but given the status is Uknown already I'm not sure what the reason should be here. 🤔

sergioifg94: Looking at the document again I'm guessing the Status should be `False` and the Reason should be `Unknown`, which kinda makes sense too. I'll change it to that to be more consistent as I guess other Kuadrant policies will be following that

pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -85,61 +86,74 @@ func (r *DNSPolicyReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	dnsPolicy := previous.DeepCopy()
 	log.V(3).Info("DNSPolicyReconciler Reconcile", "dnsPolicy", dnsPolicy)
 
+	// add finalizer to the dnsPolicy
```
laurafitzgerald, line 89: This is the beginning of a refactor that I did in the main reconcile which I'd like to folks' thoughts on.

laurafitzgerald:
tests are broken with the latest commits but pushed them early for feedback. Fixing the tests now.

philbrookes:
This PR is stale because it has been open for 60 days with no activity.

mikenairn:
Needs to be redone against the kuadrant operator repo

PR #694, opened 2023-11-23
docs: Add DNSPolicy overview doc

### LINKED ISSUE #670: Update docs for DNSPolicy providerRef
**What**

Update the quick start guides and reference docs based on the work done for DNSPolicy in a single cluster context. (.e. providerRef)

https://github.com/Kuadrant/architecture/pull/30/files

mikenairn:
* Created a dnspolicy overview doc docs/dnspolicy/dnspolicy.md based on the structure of other kuadrant overview docs. 
*  Moved current dnspolicy doc to docs/how-to/multicluster-loadbalanced-dnspolicy.md (This needs a follow on to convert it correctly to a how to guide). What is in this doc currently is kind of a mix of a guide and reference.

closes #670

### DIFF
--- a/docs/dnspolicy/dnspolicy.md
+++ b/docs/dnspolicy/dnspolicy.md
@@ -1,0 +1,282 @@
+# Kuadrant DNSPolicy
+
+The DNSPolicy is a [GatewayAPI](https://gateway-api.sigs.k8s.io/) policy that uses `Direct Policy Attachment` as defined in the [policy attachment mechanism](https://gateway-api.sigs.k8s.io/v1alpha2/references/policy-attachment/) standard.
+This policy is used to provide dns management for gateway listeners by managing the lifecycle of dns records in external dns providers such as AWS Route53 and Google DNS.
+
+## How it works
+
+A DNSPolicy and its targeted Gateway API networking resource contain all the statements to configure both the ingress gateway and the external DNS service. 
+The needed dns names are gathered from the listener definitions and the IPAdresses | CNAME hosts are gathered from the status block of the gateway resource.
+
+### The DNSPolicy custom resource
+
+#### Overview
+
+The `DNSPolicy` spec includes the following parts:
+
+* A reference to an existing Gateway API resource (`spec.targetRef`)
+* DNS Routing Strategy (`spec.routingStrategy`)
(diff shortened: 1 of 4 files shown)


docs/dnspolicy/dnspolicy.md
```diff
@@ -0,0 +1,281 @@
+# Kuadrant DNSPolicy
+
+The DNSPolicy is a [GatewayAPI](https://gateway-api.sigs.k8s.io/) policy that uses `Direct Policy Attachment` as defined in the [policy attachment mechanism](https://gateway-api.sigs.k8s.io/v1alpha2/references/policy-attachment/) standard.
+This policy is used to provide dns management for gateway listeners by managing the lifecycle of dns records in external dns providers such as AWS Route53 and Google DNS.
+
+## How it works
+
+A DNSPolicy and its targeted Gateway API networking resource contain all the statements to configure both the ingress gateway and the external DNS service.
```
maleck13, line 8:
```suggestion
A DNSPolicy and its targeted Gateway API networking resource contain all the statements to configure both the ingress gateway and the external DNS service. The dnsNames are gathered from the listener definitions and the IPAdresses | CNAME hosts are gathered from the status block of the gateway resource. 
```

docs/dnspolicy/dnspolicy.md
```diff
@@ -0,0 +1,281 @@
...
+  targetRef:
+    group: gateway.networking.k8s.io
+    kind: Gateway
+    name: mygateway
+   
+  # (optional) routing strategy to use when creating DNS records, defaults to `loadbalanced`
+  # determines what DNS records are created in the DNS provider
+  # check out Kuadrant RFC 0005 https://github.com/Kuadrant/architecture/blob/main/rfcs/0005-single-cluster-dnspolicy.md to learn more about the Routing Strategy field
+  # One-of: simple, loadbalanced.
+  routingStrategy: loadbalanced
```
maleck13, line 41: Not to do with this doc, but I wonder if part of v1alpha2 if we should move this default to be simple rather than load balanced?

mikenairn: I think it would make sense that the simple routing strategy, even more so if the policies end up in a different repo from the OCM multi cluster stuff like we intend to do.

PR #705, opened 2023-12-01
Fixing prompt text when asking for AWS credentials

pmccarthy:
**Summary**
Fixing the wording used in the quickstart prompts for AWS credentials. Current wording causes confusion as the term `secret` is used for both the access key and secret key

### DIFF
--- a/hack/.quickstartEnv
+++ b/hack/.quickstartEnv
@@ -23,13 +23,13 @@ requiredENV() {
 
   if [[ "$PROVIDER" == "aws" ]]; then
     if [[ -z "${MGC_AWS_ACCESS_KEY_ID}" ]]; then
-    echo "Enter an AWS secret access key id for an Account where you have access to Route53:"
+    echo "Enter an AWS access key ID for an account where you have access to Route53:"
     read MGC_AWS_ACCESS_KEY_ID </dev/tty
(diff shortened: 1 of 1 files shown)


hack/.quickstartEnv
```diff
@@ -23,13 +23,13 @@ requiredENV() {
...
     if [[ -z "${MGC_AWS_ACCESS_KEY_ID}" ]]; then
-    echo "Enter an AWS secret access key id for an Account where you have access to Route53:"
+    echo "Enter an AWS access key id for an Account where you have access to Route53:"
     read MGC_AWS_ACCESS_KEY_ID </dev/tty
     echo "export MGC_AWS_ACCESS_KEY_ID for future executions of the script to skip this step"
     fi
 
     if [[ -z "${MGC_AWS_SECRET_ACCESS_KEY}" ]]; then
-    echo "Enter your AWS secret access key for an Account where you have access to Route53:"
+    echo "Enter an AWS secret access key for an Account where you have access to Route53:"
```
guicassolato, line 32: Is it clear that _the_ secret access key has to be the one that pairs with the access key id entered before?

pmccarthy: @guicassolato wording updated now - mind taking another look please?

guicassolato (approved):

PR #706, opened 2023-12-04
#703: Re-add health check components

### LINKED ISSUE #703: DNSHealthCheckProbe doesn't contain status field
### Summary
`DNSHealthCheckProbe` object doesn't contain a status field with the status of the endpoint

### Issue
After creating `DNSPolicy` with health checks enabled, the corresponding automatically created `DNSHealthCheckProbe` object doesn't contain the status field anymore (it was present on the mgc-controller `release-0.1` image)

### Expected Behavior
`DNSHealthCheckProbe` object after the creation contains a status field that shows the status(health/unhealthy) of the endpoint

### Environment
mgc-controller image: `release-0.2`

sergioifg94:
> Closes #703 

Re-add health check monitor and queue, as well as reconciler to policy controller. These components were removed accidentally as part of the separation of controllers

## Verification steps

Run the local setup and include a health check section in the DNSPolicy
Verify that a DNSHealthCheckProbe is created and the status is updated

### DIFF
--- a/cmd/policy_controller/main.go
+++ b/cmd/policy_controller/main.go
@@ -19,6 +19,7 @@ package main
 import (
 	"flag"
 	"os"
+	"time"
 
 	certmanv1 "github.com/jetstack/cert-manager/pkg/apis/certmanager/v1"
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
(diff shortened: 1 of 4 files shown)


mikenairn:
@sergioifg94 Do we not have any tests that check this? Maybe an e2e test that uses health checks in some way would be good to add?

sergioifg94:
@mikenairn We only have integration tests and unfortunately they didn't catch this because the health check components didn't get deleted from the test suite. I agree we could use some e2e tests for it. I'll create an issue

PR #711, opened 2023-12-07
feat: Remove ocm dependency from policy controller

### LINKED ISSUE #622: Remove OCM module dependencies from DNSPolicy

mikenairn:
Remove all dependencies related to OCM from the policy controller and associated code.

Instead of labelling the OCM ManagedCluster resources with weight and geo metadata, the labels are now added to the gateway itself with the cluster name embedded.

```
"kuadrant.io/kind-mgc-control-plane_lb-attribute-weight":   "TSTATTR",
"kuadrant.io/kind-mgc-control-plane_lb-attribute-geo-code": "EU"
```

closes #622 

ToDo

- [x] Add Watch for ManagedClusters to Gateway controller and transfer kuadrant.io labels


pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -67,9 +65,9 @@ type DNSPolicyReconciler struct {
 //+kubebuilder:rbac:groups=kuadrant.io,resources=dnspolicies/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=kuadrant.io,resources=dnspolicies/finalizers,verbs=update
 //+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclusters,verbs=get;list;watch
```
maleck13, line 67: should this be removed now?

pkg/controllers/events/cluster_eventmapper.go
```diff
@@ -65,14 +56,14 @@ func (m *ClusterEventMapper) mapToPolicyRequest(ctx context.Context, obj client.
 
 	requests := make([]reconcile.Request, 0)
 	for _, gw := range allGwList.Items {
-		val := metadata.GetAnnotation(&gw, gateway.GatewayClustersAnnotation)
+		val := metadata.GetAnnotation(&gw, "kuadrant.io/gateway-clusters")
```
maleck13, line 59: wondering why we moved away from a constant?

pkg/controllers/dnspolicy/dnspolicy_controller.go
makslion: nice catch. Removed

pkg/controllers/events/cluster_eventmapper.go
makslion: I think the idea Mike had while doing it - remove as many ties to the gateway package as possible to ease the transition to a separate repo :thinking:  could undo it, if you prefer.

pkg/dns/target.go
```diff
@@ -82,12 +83,6 @@ func (t *MultiClusterGatewayTarget) setClusterGatewayTargets(clusterGateways []C
 	return nil
 }
 
-// ClusterGateway contains the addresses of a Gateway on a single cluster and the attributes of the target cluster.
```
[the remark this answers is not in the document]
maleck13: like perhaps `cluster_types.go`

pkg/utils/gateway_wrapper.go
```diff
@@ -4,12 +4,11 @@ import (
 	"fmt"
 	"strings"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
maleck13, line 7: wondering if this really belongs in utils or perhaps it could be also added to the pkg/apis ?

maleck13 (commented):
Question about location of the wrapper types. Also would be good to have some verification steps

pkg/dns/target.go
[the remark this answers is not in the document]
mikenairn: Yeah gateway wrapper belongs elsewhere. `pkg/apis` certainly makes more sense. Moving that though would probably be better done in a different PR either before or after this one. Will look into it, if it doesn't add a lot of additional changes I'll add it here.

pkg/controllers/events/cluster_eventmapper.go
mikenairn: This was more because the constant is in the gateway package and that now imports this. I think I'll event handler back into the gateway package since its only relevant there.

philbrookes (approved):
/lgtm

pkg/dns/target.go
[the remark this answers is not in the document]
mikenairn: I'm going to do a follow up for this. There are some other changes/refactors/cleanup i want to make to the code that could all be part of a separate PR that has no functional changes.

PR #728, opened 2023-12-13
add status to DNS policy for ignored wilcard hosts

### LINKED ISSUE #726: dns health checks improve status reporting
**What**

We don't create health checks for wildcard listeners. However this is not expressed in the status of the DNSPolicy in anyway leaving the end user guessing why there is no health check setup. 

We should express in the status if no health check has been setup

philbrookes:
When a gateway has wildcard listeners and does not create health probes for them, it should be more obvious that it has skipped them.

## Verification

- Create a gateway with no wildcard listeners
- Target with DNS Policy with health check spec
- see status on DNS policy is:
```
  healthCheck:
    conditions:
    - lastTransitionTime: "....."
      message: 'all gateways routes are creating health checks'
      observedGeneration: 1
      reason: NoGatewaysWithWildcardHosts
      status: "True"
      type: NoIgnoredHosts
```
- Add a wildcard listener to the gateway
- See status on DNS Policy is:
```
  healthCheck:
    conditions:
    - lastTransitionTime: "...."
      message: 'gateways with ignored wildcard hosts: <NAMESPACE>/<GATEWAY NAME>'
      observedGeneration: 2
      reason: NoGatewaysWithWildcardHosts
      status: "False"
      type: NoIgnoredHosts
```
## Closes
#726

### DIFF
--- a/pkg/controllers/dnspolicy/dnspolicy_controller.go
+++ b/pkg/controllers/dnspolicy/dnspolicy_controller.go
@@ -44,10 +44,12 @@ import (
 )
 
 const (
-	DNSPolicyFinalizer                                    = "kuadrant.io/dns-policy"
-	DNSPoliciesBackRefAnnotation                          = "kuadrant.io/dnspolicies"
-	DNSPolicyBackRefAnnotation                            = "kuadrant.io/dnspolicy"
-	DNSPolicyAffected            conditions.ConditionType = "kuadrant.io/DNSPolicyAffected"
+	DNSPolicyFinalizer                                           = "kuadrant.io/dns-policy"
+	DNSPoliciesBackRefAnnotation                                 = "kuadrant.io/dnspolicies"
+	DNSPolicyBackRefAnnotation                                   = "kuadrant.io/dnspolicy"
+	DNSPolicyAffected                   conditions.ConditionType = "kuadrant.io/DNSPolicyAffected"
+	DNSPolicyHealthChecksNoIgnoredHosts                          = "NoIgnoredHosts"
+	DNSPolicyHealthChecksIgnoredReason                           = "NoGatewaysWithWildcardHosts"
 )
 
 type DNSPolicyRefsConfig struct{}
--- a/pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
+++ b/pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	k8serror "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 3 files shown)


pkg/controllers/dnspolicy/dnspolicy_controller.go
```diff
@@ -44,10 +44,12 @@ import (
...
-	DNSPolicyFinalizer                                    = "kuadrant.io/dns-policy"
-	DNSPoliciesBackRefAnnotation                          = "kuadrant.io/dnspolicies"
-	DNSPolicyBackRefAnnotation                            = "kuadrant.io/dnspolicy"
-	DNSPolicyAffected            conditions.ConditionType = "kuadrant.io/DNSPolicyAffected"
+	DNSPolicyFinalizer                                           = "kuadrant.io/dns-policy"
+	DNSPoliciesBackRefAnnotation                                 = "kuadrant.io/dnspolicies"
+	DNSPolicyBackRefAnnotation                                   = "kuadrant.io/dnspolicy"
+	DNSPolicyAffected                   conditions.ConditionType = "kuadrant.io/DNSPolicyAffected"
+	DNSPolicyHealthChecksNoIgnoredHosts                          = "NoIgnoredHosts"
+	DNSPolicyHealthChecksIgnoredReason                           = "NoGatewaysWithWildcardHosts"
```
maleck13, line 52: I am finding the reason here hard to reason about. Should it not be `DNSPolicyHealthChecksIgnoredReason = GatewayHasWildcardHosts`?

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
```diff
@@ -32,21 +33,51 @@ func (r *DNSPolicyReconciler) reconcileHealthCheckProbes(ctx context.Context, dn
...
 		if err := r.deleteUnexpectedGatewayHealthCheckProbes(ctx, expectedProbes, gw.Gateway, dnsPolicy); err != nil {
 			return fmt.Errorf("error removing unexpected probes for gateway %v: %w", gw.Gateway.Name, err)
 		}
-
+		if ignoredHosts != nil {
+			gwsWithIgnoredHosts = append(gwsWithIgnoredHosts, gw.Namespace+"/"+gw.Name)
+		}
 	}
+
+	addHealthCheckCondition(dnsPolicy, gwsWithIgnoredHosts)
```
maleck13, line 52: so I can see this is iterating through gateways, but i think this is to allow for a policy changing targetRef. A policy can only ever target one gateway. So do we want to add a condition if the ref is no longer valid?

pkg/controllers/dnspolicy/dnspolicy_healthchecks.go
```diff
@@ -32,21 +33,51 @@ func (r *DNSPolicyReconciler) reconcileHealthCheckProbes(ctx context.Context, dn
...
+		ObservedGeneration: dnsPolicy.Generation,
+		LastTransitionTime: metav1.Now(),
+		Reason:             DNSPolicyHealthChecksIgnoredReason,
+		Message:            "all gateways routes are creating health checks",
+	}
+
+	if len(gwsWithIgnoredHosts) > 0 {
+		noIgnoredHostsCondition.Status = "False"
+		noIgnoredHostsCondition.Reason = DNSPolicyHealthChecksIgnoredReason
+		noIgnoredHostsCondition.Message = "gateways with ignored wildcard hosts: " + strings.Join(gwsWithIgnoredHosts, ", ")
```
maleck13, line 74: This might be a bit noisy. Perhaps be better to just say, there are wildcard hosts/liseteners that have been ignored. Then leave it as an action of the user to look at the gateway definiton?

maleck13 (commented):
code looks fine. Just a bit confused about the reason description

maleck13:
@philbrookes wondering what you want to do with this PR? Coming back to it?

pkg/controllers/dnspolicy/dnspolicy_controller.go
philbrookes: yeah, this is a bit confusing, I can't recall why I phrased it like this... It looks mental. I'll update it to be more readable.

mikenairn:
If this is still required it will need to be moved to the kuadrant operator repo

PR #755, opened 2024-01-09
make target

R-Lawton:


Makefile
```diff
@@ -95,6 +95,21 @@ test: test-unit test-integration ## Run tests.
 test-e2e: ginkgo
 	$(GINKGO) -tags=e2e -v ./test/e2e
 
+
+
+	.PHONY: local-setup-api
```
david-martin, line 100: The whitespace doesn't seem to affect running the make target (I thought makefiles would be strict on that).
For consistency, let's remove it?
```suggestion
.PHONY: local-setup-api
```

PR #756, opened 2024-01-10
Add stitch dashboards

david-martin:
Closes #744 
Closes #745 

The first commit is a bugfix for something I noticed when making changes. A patch to the number of alertmanager replicas wasn't being applied.

The 2nd commit installs the prometheus, thanos, grafana, kube-state-metrics stack, in the same manner as how it is installed for the mgc metrics quickstart. This sets up metric collection from the components we want (istio, istiod, envoy), and federates the metrics back to thanos in the hub. It also installs some grafana dashboards for mgc use (which are not relevant to stitch api) and some stitch api specific dashboards (imported from the stitch poc private repo).

No data will show up on the new dashboard unless an app is configured in the istio gateway. 
Therefore it's sufficient to just see the 3 stitch dashboards show up in grafana

image

PR #757, opened 2024-01-11
Fix Alertmanager name so patch gets applied

david-martin:
If following the metrics quickstart, alertmanager runs with 3 replicas.
This change fixes a patch so that only 1 replica runs (which was the expected behaviour when originally added)

@R-Lawton This makes sense to cherry-pick from #756 back to main as well.


david-martin:
> ill cherry pick this commit to the api-upstream branch

It's already on that branch.
This PR is enough to make sure it's everywhere it needs to be

maleck13:
I think if you rebase this it should fix the failing tests. Although with this one I am ok to merge without that

/lgtm

PR #758, opened 2024-01-11
Copy the general flow and example commands to the API walkthrough, to…

david-martin:
… be used with a new quickstart script

A few changes here:

- Copy the mgc quickstart script to a new quickstart for trying out the API walkthrough
  - The main thing this gives us is a local environment with mgc running and configured with route53 integration, all from a 1 liner
  - This script can exist along with the new local-setup-api make target as both have different audiences (trying it out vs. local development)
- Copy the general flow of the stitch poc demo script into the api walkthrough doc. 
  - Includes numerous TODOs as placeholders
  - Includes example commands copied form the demo script. Some may work, some may need to be tweaked
  - Will have assumptions about certain things the stitch poc setup would have done in the clusters. Hopefully all captured in the TDOOs
- Modify the stitch api dashboards install function to work with either local files or remote files (so the script can be run locally while working on it)

To try this out locally, you can follow the first part of 'Platform Engineer Steps (Part 1)' up until the 'Create a Gateway' section

### DIFF
--- a/docs/how-to/api-walkthrough.md
+++ b/docs/how-to/api-walkthrough.md
@@ -2,7 +2,155 @@
 
 ## Introduction
 
-This document will detail the setup of a reference architecture to support a number of API management use-cases connecting Kuadrant with other projects the wider API management on Kubernetes ecosystem.
+This document will detail the setup of a reference architecture to support a number of API management use-cases connecting Kuadrant with other projects in the wider API management on Kubernetes ecosystem.
 
-## Petstore App Deployment
+## Platform Engineer Steps (Part 1)
 
+
+<!-- TODO: Copy formatting & env var info from the MGC Getting Started guide -->
+
+Export the following env vars:
+
+    ```
+    export KUADRANT_AWS_ACCESS_KEY_ID=<key_id>
+    export KUADRANT_AWS_SECRET_ACCESS_KEY=<secret>
+    export KUADRANT_AWS_REGION=<region>
+    export KUADRANT_AWS_DNS_PUBLIC_ZONE_ID=<zone>
+    export KUADRANT_ZONE_ROOT_DOMAIN=<domain>
+    ```
+
+Run the following command, choosing `aws` as the dns provider:
+
+<!-- TODO: Change to a curl command that fetches everything remotely -->
+
+    ```bash
+    MGC_LOCAL_QUICKSTART_SCRIPTS_MODE=true MGC_BRANCH=api-upstream ./hack/quickstart-setup-api.sh
+    ```
+
+### Create a gateway
+
+<!-- TODO: Create Gateway & TLSPolicy as part of quickstart, if possible -->
+
+
+View the ManagedZone, Gateway and TLSPolicy:
+
+    ```bash
+    kubectl --context kind-mgc-control-plane describe managedzone mgc-dev-mz -n multi-cluster-gateways
+    kubectl --context kind-mgc-control-plane describe gateway -n multi-cluster-gateways
+    kubectl --context kind-mgc-control-plane describe tlspolicy -n multi-cluster-gateways
+    ```
+
+### Guard Rails: Show Constraint warnings about missing policies ( DNS, AuthPolicy, RLP)
+
+<!-- TODO: Instructions how to get to the dashboard -->
+
+### Create missing Policies
+
+<!-- TODO: Guard Rails: Show Constraint warnings about missing policies ( DNS, AuthPolicy, RLP) -->
+
+Create a DNSPolicy:
+
+<!-- TODO: Import dnspolicy from platform-engineer repo into this repo -->
+
+    ```bash
+    envsubst < ./resources/dnspolicy.yaml | kubectl --context kind-mgc-control-plane apply -f -
+    kubectl --context kind-mgc-control-plane describe dnspolicy prod-web -n multi-cluster-gateways
+    ```
+
+View ns entries in Route 53 DNS Zone
+
+<!-- TODO: Instructions how to find ns entries in route53 zone -->
+
+Create and configure a Gateway-wide RateLimitPolicy
+
+<!-- TODO: Import ratelimitpolicy from platform-engineer repo into this repo -->
+
+    ```bash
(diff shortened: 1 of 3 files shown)


docs/how-to/api-walkthrough.md
```diff
@@ -2,7 +2,155 @@
...
 
-## Petstore App Deployment
+## Platform Engineer Steps (Part 1)
 
+
+<!-- TODO: Copy formatting & env var info from the MGC Getting Started guide -->
+
+Export the following env vars:
+
+    ```
```
R-Lawton, line 14: Your quotes are showing in the rendered  markdown

docs/how-to/api-walkthrough.md
```diff
@@ -2,7 +2,155 @@
...
+<!-- TODO: Instructions how to get to the dashboard -->
+
+* Gateways & Policies - Gateway Policies created
+* No Route Policies yet (as no APIs/Apps deployed yet) - Can see a TLSPolicy, DNSPolicy, AuthPolicy and RateLimitPolicy
+* Constraints & Violations - Highlight no more violations
+* APIs Summary - Highlight there are no APIs yet
+
+## App Developer Steps
+
+### API Setup
```
R-Lawton, line 95: This is showing as is rather then a heading

docs/how-to/api-walkthrough.md
```diff
@@ -2,7 +2,155 @@
...
+
+TODO
+
+* deploy the petstore app to 1 cluster
```
R-Lawton, line 99: Can this be made capitalised like the rest of the list

hack/.deployUtils
```diff
@@ -321,13 +321,28 @@ deployPrometheusForFederation() {
 }
 
 installAPIDashboards() {
+  local PROMETHEUS_FOR_FEDERATION_API_DASHBOARDS_KUSTOMIZATION_DIR="$2"
+  local PROMETHEUS_FOR_FEDERATION_API_DASHBOARDS_GRAFANA_PATCH="$3"
+  if [ -z "$2" ]; then
+    PROMETHEUS_FOR_FEDERATION_API_DASHBOARDS_KUSTOMIZATION_DIR=${LOCAL_SETUP_DIR}/../config/prometheus-for-federation/api-dashboards
+  fi
+  if [ -z "$3" ]; then
```
R-Lawton, line 329: In a use case where you dont want to patch grafana this will still be populated I'm thinking could we add a extra check

hack/.deployUtils
```diff
@@ -321,13 +321,28 @@ deployPrometheusForFederation() {
...
   clusterName=${1}
   if [[ -n "${METRICS_FEDERATION}" ]]; then
     echo "Deploying API Dashboards in ${clusterName}"
     kubectl config use-context kind-${clusterName}
-    ${KUSTOMIZE_BIN} build ${PROMETHEUS_FOR_FEDERATION_KUSTOMIZATION_DIR}/api-dashboards | kubectl apply -f -
-
-    kubectl patch deployment grafana -n monitoring --type=json -p "$(cat ${PROMETHEUS_FOR_FEDERATION_KUSTOMIZATION_DIR}/api-dashboards/grafana_deployment_patch.yaml)"
+    ${KUSTOMIZE_BIN} build ${PROMETHEUS_FOR_FEDERATION_API_DASHBOARDS_KUSTOMIZATION_DIR} | kubectl apply -f -
+
+    if [[ ${PROMETHEUS_FOR_FEDERATION_API_DASHBOARDS_GRAFANA_PATCH} == https* ]]; then
```
R-Lawton, line 338: if the env var is empty the value would be grafana_deployment_patch.yaml never being https ?

hack/.deployUtils
```diff
@@ -321,13 +321,28 @@ deployPrometheusForFederation() {
 }
 
 installAPIDashboards() {
+  local PROMETHEUS_FOR_FEDERATION_API_DASHBOARDS_KUSTOMIZATION_DIR="$2"
+  local PROMETHEUS_FOR_FEDERATION_API_DASHBOARDS_GRAFANA_PATCH="$3"
```
R-Lawton, line 325: Im thinking a format check on the env var to ensure its in a way we expect

hack/quickstart-setup-api.sh
```diff
@@ -0,0 +1,132 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+export KFILT="docker run --rm -i ryane/kfilt"
+
+METRICS_FEDERATION=true
```
R-Lawton, line 21: do we ever expect this to be false?

hack/.deployUtils
david-martin: The patching is to mount the new dashboards.
I don't see a reason not to always do both.

hack/.deployUtils
david-martin: There's a default set above to `${PROMETHEUS_FOR_FEDERATION_API_DASHBOARDS_KUSTOMIZATION_DIR}/grafana_deployment_patch.yaml`
In that case it would be a relative file.

This if/else check is to allow for the script being run locally (which we do for now), or being run from a 1 liner curl (where it has to fetch everything remotely)

hack/.deployUtils
david-martin: I didn't add any validation as it's not done anywhere else in this file where env vars are passed in.
If the value is invalid in some way, the script will likely fail with an error.
At this time, the values being passed in to this function are not influenced by the person running the script, so they are controlled.

hack/.deployUtils
R-Lawton: Never mind had a brain fart moment 🙈

hack/.deployUtils
R-Lawton: no your right i forgot how were were adding the dashboard using patch for the POC i was thinking adding new dashboards to existing one ignore me 😂

PR #760, opened 2024-01-11
Adding gatekeeper, gatekeeper scorecard and constraints/muatations

R-Lawton:
Added gatekeeper to the local setup as well as gatekeeper scorecard which gives extra metrics and the constraint templates constraints and mutation. 

### verification 
`make local-setup-api OCM_SINGLE=true MGC_WORKLOAD_CLUSTERS_COUNT=1`
Ensure the script runs with no errors and the components get deployed
There wont be any constraint violation unless you make a gateway


david-martin:
👀

hack/.deployUtils
```diff
@@ -479,4 +479,57 @@ deployPrometheus() {
...
   wait_for "${KUSTOMIZE_BIN} --load-restrictor LoadRestrictionsNone build ${PROMETHEUS_DIR} --enable-helm --helm-command ${HELM_BIN} | kubectl apply -f -" "${PROMETHEUS_DIR} cluster config apply" "1m" "5"
+}
+
+deployGatekeeper(){
+  clusterName=${1}
+  kubectl config use-context kind-${clusterName}
+  echo "Deploying Gatekeeper in kind-${clusterName}"
+  ${HELM_BIN} repo add gatekeeper https://open-policy-agent.github.io/gatekeeper/charts
+  ${HELM_BIN} install gatekeeper/gatekeeper --name-template=gatekeeper --namespace gatekeeper-system --create-namespace
+  echo "test"
```
david-martin, line 490: stray debug echo?

hack/upstream-api/local-setup-mgc-api.sh
```diff
@@ -104,6 +113,9 @@ if [[ -n "${MGC_WORKLOAD_CLUSTERS_COUNT}" ]]; then
     deployOCMSpoke ${KIND_CLUSTER_WORKLOAD}-${i}
     deployPrometheusForFederation ${KIND_CLUSTER_WORKLOAD}-${i}
     configureManagedAddon ${KIND_CLUSTER_CONTROL_PLANE} ${KIND_CLUSTER_WORKLOAD}-${i}
+    deployGatekeeper ${KIND_CLUSTER_WORKLOAD}-${i}
+    configureGatekeeper ${KIND_CLUSTER_WORKLOAD}-${i}
+    installGatekeeperScoreCard ${KIND_CLUSTER_WORKLOAD}-${i}
```
david-martin, line 118: This is fine to get things installed with examples in all clusters.
We'll definitely end up changing things for the quickstart as there's 1 hub cluster and 2 spokes.
Some constraints belong only on the hub, while others are for the app developer and belong in the workload clusters.

hack/.deployUtils
R-Lawton: Ooops yep

hack/.deployUtils
```diff
@@ -479,4 +479,57 @@ deployPrometheus() {
...
+      gatekeeper.sh/system: 'yes'
+EOF
+echo "Gatekeeper Ready"
+}
+
+configureGatekeeper() {
+  kubectl --context kind-${clusterName} apply -f ./config/api-upstream/gatekeeper/constraintTemplates.yaml
+  kubectl --context kind-${clusterName} wait  --for=condition=established --timeout=300s crd/requirepolicytargetinggateway.constraints.gatekeeper.sh
+  kubectl --context kind-${clusterName} apply -f ./config/api-upstream/gatekeeper/constraints.yaml
+  kubectl --context kind-${clusterName} apply -f ./config/api-upstream/gatekeeper/mutations.yaml
```
david-martin, line 513: Would you mind using kustomize here, similar to the other functions, so that relative files are not used.
That will make it easier for calling the function from a quickstart script that's been pulled down in isolation (without the rest of the repo).
Putting the 3 files into a kustomization file should do the trick normally, but I see there's a 'wait'.
Perhaps splitting into 2 kustomization files or using kfilt (there's an example elsewhere for filtering on CRDs first, then omitting them)

hack/upstream-api/local-setup-mgc-api.sh
R-Lawton: i keep thinking of all hubs being both hub and spoke rather then a hub just being a hub. ill update the script to be able to specify where the constraints should go i don't think it makes sense for the make target to be different to the quickstart.

R-Lawton: ok grand

jasonmadigan: @david-martin I don't think we showed any for an app dev, only PE ones (at the "hub")

maleck13:
I am wondering what the goal is in adding this to the deployment scripts? I already think we should prob slim things down quite a bit. IE remove things like argocd and anything else we don't currenty integrate directly with. We are increasing the maintenance overhead where I think ideally we should be looking for ways to reduce it

david-martin (approved):

PR #761, opened 2024-01-12
Use full sha for policy-controller builds

### LINKED ISSUE #759: Use full sha for policy-controller builds
The kuadrant nightly builds depend on policy-controller nightlies. We extract the commit sha from the tag in quay, but when pulling the dependencies to create the bundle, the pipeline fails. This is due to kustomize not supporting pulling the dependencies using the short sha and needs the full sha. If we tag the quay images with the full commit sha the nightlies should start to work again.

adam-cattermole:
Updating the policy-controller image build to use the full commit sha

Resolves #759

### DIFF
--- a/.github/workflows/policy-controller-image.yaml
+++ b/.github/workflows/policy-controller-image.yaml
@@ -20,20 +20,18 @@ jobs:
     name: Build controller image
     runs-on: ubuntu-22.04
     outputs:
-      sha_short: ${{ steps.vars.outputs.sha_short }}
-      controller_image: ${{ steps.vars.outputs.base_image }}:${{ steps.vars.outputs.sha_short }}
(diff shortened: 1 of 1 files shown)


maleck13 (approved):

maleck13:
@adam-cattermole I assume there is no issue with the tag length here?

adam-cattermole:
> @adam-cattermole I assume there is no issue with the tag length here?

Not that I know of, we're using the full sha for our tags for kuadrant -> https://github.com/Kuadrant/kuadrant-operator/blob/main/.github/workflows/build-images-scheduled.yaml#L13-L14

maleck13:
These two failing items are known issues so this can be merged
/lgtm

PR #775, opened 2024-01-22
Add kaudrant resources to ksm permissions

david-martin:
This has been already applied to the api-upstream branch [here](https://github.com/Kuadrant/multicluster-gateway-controller/commit/b9f53406c94827fecaeac263796afac857660509), which is needed & referenced by the api-quickstart repo.
However, it should land back on main here as well.

### DIFF
--- a/config/prometheus-for-federation/ksm_clusterrole_patch.yaml
+++ b/config/prometheus-for-federation/ksm_clusterrole_patch.yaml
@@ -24,3 +24,16 @@
     verbs:
     - list
     - watch
+- op: add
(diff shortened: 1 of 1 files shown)


mikenairn:
@david-martin Is this still needed?

david-martin:
@mikenairn Yes. 
This will become important if we include metrics and dashboards content like what's in the api-quickstart. Right now that is in the experimental section of the docs site and uses the `api-upstream` branch, where this change has been made.

david-martin:
Thanks @mikenairn

PR #777, opened 2024-01-30
Remove old code and docs

mikenairn:
[Remove redundant docs:](https://github.com/Kuadrant/multicluster-gateway-controller/commit/3b91db7bcdd01175d3ecd899b5002168f227b531)

- docs/demos/dns-policy - Out of date and hard to maintain with constant changes
- docs/proposals - Should all be in the architecture repo now.

ToDo Need to check docs site as it might be referencing some of these.

[Remove provider health checks](https://github.com/Kuadrant/multicluster-gateway-controller/commit/89e1893b8316ad265be036050cb923eaa83ad117)

Remove the provider health checks from the dns provider implementations.
Replaced by provider agnostic health checks and there is no mention of these in the docs any longer.

[Refactor dns provider code](https://github.com/Kuadrant/multicluster-gateway-controller/commit/47f781a13422c6768d1b8914df355ed1032087b3)

* Move dns provider related code into provider package (dns/provider). Matches external-dns providers and also reduces the chance of import cycle issues.
* Rename dns/provider/aws/dns/go -> dns/provider/aws/aws/go (Matches google provider and the naming we will use for future providers).

Note: Not all that has been agreed yet, the PR will remain a draft until then.


mikenairn:
@maleck13 @philbrookes Could you take a look at this and see if we are happy enough with removing the docs and provider health checks.

The health checks i don't really see the need to keep the code and would rather just remove it now if we can, provider agnostic health checks are what we support AFAIK, and we have no intention of implementing provider health checks for other providers. The docs removal, i want to get rid of `docs/demos/dns-policy`  as it's just always out of date, but wasn't so sure about proposals, can undo that if we think there is value in leaving it.

PR #780, opened 2024-02-01
Dns refactor

mikenairn:
Number of changes required(mostly) to isolate the dns code from the policy and gateway logic in preparation for moving it to it's own repo.

Provider health checks were removed from the aws dns provider implementation. These have been replaced by provider agnostic health checks and are no longer a supported solution.

DNS provider implementations are moved into a provider package(dns/provider) along with all other dns provider related interfaces and types. Matches external-dns layout and naming and how we want future providers to be added.

Updates the dns provider factory to have each provider register its constructor with the factory on init.  Adds fake provider and factory for testing and updates field names in reconcilers to better represent what the Factory resource is (DNSProvider -> ProviderFactory). Factory method (ProviderFor) now expects a resource implementing the ProviderAccessor interface.
Note: These changes were based on how certmanager implements registration and a factory for Issuers and seemed like a nice solution. Also gets round some import cycle issues i was having with the refactor.

Move common dns consts and types into api (DefaultWeight, DefaultGeo, WildcardGeo, GeoCode ), remove redundant code from DNSRecord and make receiver name consistent. Added common pkg for consistency with kuadrant-operator and moved target and gateway wrapper resources into here.

Move health package into dns/health. The contents of the dns package will all be moved to the new dns controller component as top level packages:                                   
```                                 
.                                   
├── controller                                         
└── pkg                                           
    ├── health                                                      
    └── provider                                        
        ├── aws                                
        ├── fake                                           
        └── google
```


maleck13 (approved):
Changes all look reasonable and make sense

PR #781, opened 2024-02-01
GH-661 join addon-manager and gw-controller binaries

### LINKED ISSUE #661: Combine OCM add-on manager and gateway controller
**What**

There is little to no value in having a separate binary for the add-on manager. This should be combined now with the gateway controller binary

makslion:
## What
Joining addon-manager binary into the gateway controller one. 
## Verification
Go through the use cases for the addon-manager to ensure it is functioning as before

closes #661


cmd/gateway_controller/main.go
```diff
@@ -143,8 +216,35 @@ func main() {
...
+
+	ctx := ctrl.SetupSignalHandler()
+
+	wg := sync.WaitGroup{}
+	wg.Add(2)
+	go startAddonManager(ctx)
+	go startGatewayController(ctx)
+	wg.Wait()
+
+	<-ctx.Done()
```
mikenairn, line 249: Would it be possible to add an addon manager runnable to the controller manager which is already managing multiple go routines for us:

```.go
if err := mgr.Add(ocm.AddonRunnable{}); err != nil {
	setupLog.Error(err, "unable to set up ocm addon manager")
	os.Exit(1)
}

type AddonRunnable struct{}

func (r AddonRunnable) Start(ctx context.Context) error {
	fmt.Println("starting add-on manager")
    ...

	if err := addonMgr.Start(ctx); err != nil {
		return err
	}

	<-ctx.Done()
	return nil
}
```

AFAIK, this should work, and will take care of what to do when the likes of leadership election is being used.
You can keep the ocm addon stuff in it's own package somewhere (pkg/ocm maybe), and import it into main to keep it a bit more organised, will also mean main.go is only changing slightly.

wdyt?

makslion: Reworked. Now should be fine

cmd/gateway_controller/addon-manager/addon-manager.go
```diff
@@ -1,9 +1,9 @@
-package main
+package addon_manager
```
mikenairn, line 1: I'd avoid using underscores in teh package names, see https://go.dev/blog/package-names. It shoudl also match the file name. Personally I'd just call it ocm.

bundle/manifests/mgc-kuadrant-addon_rbac.authorization.k8s.io_v1_clusterrole.yaml
```diff
@@ -0,0 +1,137 @@
...
+- apiGroups:
+  - kuadrant.io/v1beta1
+  resources:
+  - kuadrant
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
```
mikenairn, line 137: What is generating this file? I wouldn't expect to be seeing anything specifically related to the addon manager now. All rbac required for the gateway controller and addon manager should just be all in one, and one service account

cmd/gateway_controller/addon-manager/addon-manager.go
makslion: renamed to `ocm`

bundle/manifests/mgc-kuadrant-addon_rbac.authorization.k8s.io_v1_clusterrole.yaml
makslion: updated. Now all permissions should reside in the controller manager

pkg/controllers/gateway/gateway_controller.go
```diff
@@ -78,14 +78,27 @@ type GatewayPlacer interface {
...
+// +kubebuilder:rbac:groups=certificates.k8s.io,resources=signers,verbs=approve
+// +kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclusters,verbs=get;list;watch;update
+// +kubebuilder:rbac:groups=work.open-cluster-management.io,resources=manifestworks,verbs=get;list;watch;create;update;delete;deletecollection;patch
+// +kubebuilder:rbac:groups=addon.open-cluster-management.io,resources=managedclusteraddons/finalizers,verbs=update
+// +kubebuilder:rbac:groups=addon.open-cluster-management.io,resources=clustermanagementaddons/finalizers,verbs=update
+// +kubebuilder:rbac:groups=addon.open-cluster-management.io,resources=clustermanagementaddons,verbs=get;list;watch
+// +kubebuilder:rbac:groups=addon.open-cluster-management.io,resources=managedclusteraddons,verbs=get;list;watch;create;update;delete
+// +kubebuilder:rbac:groups=addon.open-cluster-management.io,resources=managedclusteraddons/status,verbs=update;patch
+// +kubebuilder:rbac:groups=kuadrant.io,resources=kuadrant,verbs=get;list;watch;create;update
+
```
mikenairn, line 95: As discussed, it seems unlikely to me that the addon manager really needs all these permissions, but since this is just trying to get combined having an entry of each of [these](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/config/rbac/add-on-clusterrole.yaml) makes sense.

We might also want to put these elsewhere at some point, addonmanager related code, but just adding them to the gateway controller seems fine for now.

mikenairn (approved):
Looks good to me :+1:

PR #782, opened 2024-02-01
Walkthrough text updates and quickstart cleanup fix

trepel:
Small updates to Multicluster Gateways Walkthroughs
Quickstart cleanup fix - the `cleanup` function no longer exists


maleck13 (approved):
Thanks @trepel

PR #785, opened 2024-02-08
Unified kuadrant - Remove TLSPolicy and DNSPolicy

mikenairn:
close #767 #766 #784 

Removes all dns and tls related resources.

* Remove dnspolicy, tlspolicy, dnsrecord, dnshealth and managedzone api and reconcilers
* Remove all policy-controller related config and make targets
* Adds the kuadrant-operator as a dependency of the MGC operator installation and is installed on the OCM hub cluster during local-setup.

**Verification**

Tested with local-setup and e2e tests.

**Requires**

* https://github.com/Kuadrant/kuadrant-operator/pull/416

PR #787, opened 2024-02-21
fix: ci bundle image build

mikenairn:
Install correct golang version (1.21.x)

Related to https://github.com/Kuadrant/multicluster-gateway-controller/pull/785

Test run here https://github.com/Kuadrant/multicluster-gateway-controller/actions/runs/7990993383


philbrookes (approved):
/lgtm

PR #789, opened 2024-02-22
Remove unnecessary dependencies and align local setup and quickstart scripts

mikenairn:
[refactor: remove unnecessary dependencies](https://github.com/Kuadrant/multicluster-gateway-controller/commit/9b77983908fa56e376421e4990f13441798d1155)

* argocd
* external-dns
* redis
* webhook
* skupper and submariner
* deploy-sample-applicationset

[refactor: Align local-setup and quick-start deployment paths](https://github.com/Kuadrant/multicluster-gateway-controller/commit/972f32bdb085f324ecdc0c086c8bece47650945d)

Updates local-setup to use the same kustomizations as the quick start scripts. Installation and configuration of hub and spoke dependencies is now consistent across both deployment paths.

The only difference between quick start and local setup now is that metrics is enabled by default with local-setup, managedzones and dns providers are created using local environment files instead of environment variables and the MGC gateway controller is not running on the hub cluster.

Note: The skupper POC docs are listed in our[ docs site](https://github.com/Kuadrant/docs.kuadrant.io/blob/main/mkdocs.yml#L157-L158) and would need to be removed before the next release if we merge this as is. Personally i see no advantage to leaving these here.


maleck13:
I think it is reasonable to remove the skupper and submariner docs at this point. We don't ever test them and really if you use gateway API and use an e/w technology that either support multi-cluster service or a regular service we will be able to work with it. Changes look fine.

Will run a local-setup to verify

maleck13:
local setup completed so happy to merge
/lgtm

PR #793, opened 2024-03-13
fix for failing quick start

maleck13:
https://github.com/Kuadrant/kuadrant-operator/issues/459

This  will get the quick start back to a working state. However we need a better fix than sourcing these scripts from main of the gateway controller repo


mikenairn (approved):

jasonmadigan:
looks like the  E2E Test Suite started failing after this landed

PR #797, opened 2024-05-01
Change multi cluster docs link to new location

david-martin:
The new url is https://docs.kuadrant.io/getting-started-multi-cluster-ocm, which will be live after https://github.com/Kuadrant/docs.kuadrant.io/pull/97 lands.
It's just the url that is changing. The page and content is the same as before.
This is to accommodate a move of all mgc & ocm related content into a separate sub-section in the docs.
The previous url https://docs.kuadrant.io/getting-started-multi-cluster will still be live, but point to a new page that doesn't use ocm or mgc.


jasonmadigan:
👀

